PR #168, opened 2026-04-23
fix(workflow): reuse resolved llm client for auto-pr

fkukuck:
Thread the workflow's resolved LLM client into native pull request generation so PR bodies use the same vault-backed provider resolution as normal runs. This fixes auto-PR failures when OpenAI is configured via credentials like openai_codex instead of process environment variables, and keeps the legacy fabro pr create call site compatible with the new signature.

Fixes fabro error log
```json
{"event":"run.notice","id":"019db948-cbf9-7690-a516-9abc253cac4a","properties":{"code":"pull_request_failed","level":"warn","message":"PR creation failed: LLM generation failed: Configuration error: Provider 'openai' not registered"},"run_id":"01KPWMCTFY4MSRTQB102839XH1","ts":"2026-04-23T07:40:49.017159628Z"}
```


brynary:
Hi @fkukuck -- Thank you for this bug fix. The bug looks legit, and it caused me to step back to how it came to be in the first place. That led me to this root cause analysis and resolution plan: docs/plans/2026-04-23-003-refactor-llm-client-resolution-and-run-services-plan.md

I'm wrapping up that implementation on this branch. The goal is that this bug will be fixed and also a bug like this will be much less likely to occur in the future.

fkukuck:
Hey @brynary , thank you for your time and effort to fix this properly. I could now find time to confirm that your fix works for me. Thanks a lot

PR #252, opened 2026-05-13
fix(slack): make per-button action_id unique (re #251 )

### LINKED ISSUE #251: Slack interview gates with >1 button silently fail with invalid_blocks (duplicate action_id)
## Summary

Slack interviews with multiple options don't work... And there's no signal in the logs to diagnose the issue. 

When a workflow hits an interview gate with more than one button (yes/no, confirmation, or multiple_choice with N>1 options), Fabro builds a chat.postMessage block where every button shares the same `action_id = "interview.answer"`. Slack rejects the entire payload with `invalid_blocks`, so no message lands in the configured channel. The error is invisible at default log levels because `SlackService::handle_event` swallows the result via `if let Ok(posted) = self.client.post_message(...)`.

Single-button cases (e.g. pure freeform, or a workflow with only one outgoing edge from the gate) post fine, which masks the bug for simple workflows.

## Repro

1. Configure server-side Slack:
   ```toml
   # ~/.fabro/settings.toml
   [server.integrations.slack]
   enabled = true
   default_channel = "#fabro-reviews"
   ```
   Plus `FABRO_SLACK_BOT_TOKEN` / `FABRO_SLACK_APP_TOKEN` in `server.env` with `chat:write` and `chat:write.public` scopes; bot invited to the channel.
2. Run a workflow whose hexagon gate has 2 or more outgoing edges, e.g. an `Approve Plan` node with `[A] Approve` and `[R] Revise` edges.
3. Wait for the gate to fire. Socket Mode connects, logs show `INFO fabro_slack::connection: Socket Mode handshake complete`, then nothing.
4. No message appears in Slack. Run blocks on `human_input_required` indefinitely; user has to fall back to the web UI to answer.

## Root cause

`lib/crates/fabro-slack/src/blocks.rs` reuses one constant:

```rust
const ANSWER_ACTION_ID: &str = "interview.answer";
// ...
QuestionType::MultipleChoice => {
    let elements = question.options.iter().map(|opt| {
        button(&opt.label, &encoded_value, ANSWER_ACTION_ID)  // same id for every button
    }).collect();
}
```

Per Slack's [block-element docs](https://docs.slack.dev/reference/block-kit/block-elements/button-element/), `action_id` must be unique within the containing message (max 255 chars). MITM'ing the chat.postMessage call returns:

```json
{
  "ok": false,
  "error": "invalid_blocks",
  "errors": [
    "`action_id` \"interview.answer\" already exists [json-pointer:/blocks/1/elements/1/action_id]"
  ]
}
```

`lib/crates/fabro-server/src/server.rs` `SlackService::handle_event` then drops the error silently:

```rust
if let Ok(posted) = self
    .client
    .post_message(&self.default_channel, &blocks, None)
    .await
{ /* ... */ }
```

which is also why no log line at INFO surfaces the failure.

## Proposed fix

I have a working patch in a fork.

Branch: https://github.com/dajulia3/fabro/tree/fix/slack-action-id-uniqueness
Diff: https://github.com/fabro-sh/fabro/compare/main...dajulia3:fabro:fix/slack-action-id-uniqueness

Shape:

- Outbound (`blocks.rs`): each button gets a unique suffix.
  - `YesNo` / `Confirmation`: `interview.answer.yes` / `interview.answer.no`
  - `MultipleChoice`: `interview.answer.<index>`
  - Using the option *index* (not the raw key) keeps suffixes short, ASCII, and well under Slack's 255-char cap regardless of author-supplied option keys. The selected key is still carried in the button `value` payload, so the action_id only needs to be unique for Slack, not semantic.
- Inbound (`interaction.rs`): `parse_interaction` now matches the legacy exact-prefix shape (so in-flight buttons posted by older builds remain clickable across upgrades) and the new suffixed shape via a pre-computed `ANSWER_ACTION_ID_PREFIX_DOT` constant (avoids per-parse `format!`).
- Tests (+5): suffixed yes/no parses, suffixed multi-choice parses, legacy exact prefix still parses, lookalike `interview.answers.yes` is correctly rejected, and a sync assertion that the dotted-prefix constant matches the canonical prefix.
- Updated one fixture each in `dispatch.rs` and `connection.rs` to the suffixed shape; left one legacy fixture in each to document the backwards-compatibility branch.

74/74 `fabro-slack` tests pass (was 69), `cargo +nightly-2026-04-14 fmt --check --all` and `cargo +nightly-2026-04-14 clippy -p fabro-slack --all-targets -- -D warnings` clean.

I also opened a courtesy PR (#252) from the fork branch in case you'd like a ready-made diff to feed your workflow, fully understanding that CONTRIBUTING.md says outside PRs aren't accepted. Close it as you wish; the issue is the canonical place to discuss.

## Environment

- Fabro `0.231.0-nightly.3` (installed via `brew install fabro-sh/tap/fabro-nightly`)
- macOS 14 / aarch64
- Rust 1.95.0 stable (build), nightly-2026-04-14 (fmt + clippy per CLAUDE.md)

dajulia3:
Heads up: [CONTRIBUTING.md](https://github.com/fabro-sh/fabro/blob/main/CONTRIBUTING.md) says you don't accept outside PRs, *"Instead of accepting outside pull requests, we accept bug reports and feature requests as GitHub Issues."* I filed the canonical bug report as **#251**, and that's where any actual discussion belongs.

This PR is a courtesy ready-made diff in case it's useful to whoever supervises the AI workflow that lands this fix. Feel free to close it without comment; nothing is being asked of you here. I just thought it'd be useful to have a reference for what I did locally to fix it.

## What

Two-file behaviour fix in `lib/crates/fabro-slack/`: every interview button on a multi-button gate now gets a Slack-unique `action_id`. Today they all share `"interview.answer"`, so Slack rejects the `chat.postMessage` with `invalid_blocks` and `SlackService::handle_event` silently drops the error.

## Why

Multi-button Slack interview gates never reach Slack. Full reproducer, MITM-captured `invalid_blocks` response, and root-cause walkthrough are in **#251**.

## Diff shape

- `blocks.rs`: each button gets a unique suffix.
  - `YesNo` / `Confirmation`: `interview.answer.yes` / `interview.answer.no`
  - `MultipleChoice`: `interview.answer.<index>` (index, not raw key, to dodge Slack's 255-char `action_id` cap and any author-supplied charset surprises; selected key still rides in the button `value`)
- `interaction.rs`: `parse_interaction` accepts both the legacy exact-prefix shape (in-flight buttons keep working across upgrade) and the new suffixed shape, via a pre-computed `ANSWER_ACTION_ID_PREFIX_DOT` constant so the parse hot path doesn't `format!` on every event.
- Tests: +5 in `interaction.rs` (suffixed yes/no, suffixed multi-choice, legacy exact prefix, lookalike `interview.answers.yes` rejected, prefix-sync assertion). Updated the existing block-builder tests to assert uniqueness instead of the old single constant. One fixture each in `dispatch.rs` and `connection.rs` updated to the suffixed shape; one legacy fixture left in each to document backwards compatibility.

74/74 `fabro-slack` tests pass (was 69/69). `cargo +nightly-2026-04-14 fmt --check --all` and `cargo +nightly-2026-04-14 clippy -p fabro-slack --all-targets -- -D warnings` clean.

## Verified end-to-end

Built a patched `fabro` binary, swapped it for the brew install, triggered a fresh multi-choice `Approve Plan` gate against a real Slack workspace, message rendered correctly in the configured channel with two clickable `[A] Approve` and `[R] Revise` buttons. Before the patch, the exact same gate produced zero Slack output and only the swallowed `invalid_blocks` was visible via MITM.

## Suggested follow-up (separate concern, not in this diff)

The silent error swallow in `SlackService::handle_event` (`if let Ok(posted) = self.client.post_message(...)`) is what hid this bug. Worth logging at `WARN`. Mentioned in #251 as a separate item.

## Closes

Closes #251 if you choose to land this directly. Otherwise this PR is just background material for the issue.

### DIFF
--- a/lib/crates/fabro-slack/src/blocks.rs
+++ b/lib/crates/fabro-slack/src/blocks.rs
@@ -4,10 +4,22 @@ use serde_json::{Value, json};
 
 use crate::payload::{SlackActionPayload, encode_action_value};
 
-const ANSWER_ACTION_ID: &str = "interview.answer";
+pub(crate) const ANSWER_ACTION_ID_PREFIX: &str = "interview.answer";
 const MULTI_SELECT_BLOCK_ID: &str = "interview.checkboxes";
 const MULTI_SELECT_ACTION_ID: &str = "interview.select";
 const MULTI_SELECT_SUBMIT_ACTION_ID: &str = "interview.submit";
+
+/// Build a Slack-unique `action_id` for an interview button.
+///
+/// Slack requires `action_id`s to be unique within a single message and caps
+/// them at 255 characters. The selected option is carried in the button
+/// `value` payload, so the `action_id` only needs to be unique — it doesn't
+/// have to encode the selection. Suffixes are short, fixed-shape tokens
+/// (`yes`, `no`, or the option index) to avoid any character-set or length
+/// concerns when option keys are author-supplied.
+fn answer_action_id(suffix: &str) -> String {
+    format!("{ANSWER_ACTION_ID_PREFIX}.{suffix}")
+}
 
 fn text_block(text: &str) -> Value {
     json!({
@@ -48,11 +60,11 @@ pub fn question_to_blocks(run_id: &str, question_id: &str, question: &Question) 
                     button("Yes", &encode_action_value(&SlackActionPayload::Yes {
                         run_id: run_id.to_string(),
                         qid: question_id.to_string(),
-                    }), ANSWER_ACTION_ID),
+                    }), &answer_action_id("yes")),
                     button("No", &encode_action_value(&SlackActionPayload::No {
                         run_id: run_id.to_string(),
                         qid: question_id.to_string(),
-                    }), ANSWER_ACTION_ID),
+                    }), &answer_action_id("no")),
                 ]
             });
             vec![section, actions]
@@ -61,7 +73,8 @@ pub fn question_to_blocks(run_id: &str, question_id: &str, question: &Question) 
             let elements: Vec<Value> = question
                 .options
                 .iter()
-                .map(|opt| {
+                .enumerate()
+                .map(|(idx, opt)| {
                     button(
                         &opt.label,
                         &encode_action_value(&SlackActionPayload::Selected {
@@ -69,7 +82,7 @@ pub fn question_to_blocks(run_id: &str, question_id: &str, question: &Question) 
                             qid:    question_id.to_string(),
                             key:    opt.key.clone(),
                         }),
-                        ANSWER_ACTION_ID,
+                        &answer_action_id(&idx.to_string()),
                     )
                 })
                 .collect();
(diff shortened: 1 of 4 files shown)


brynary:
@dajulia3 -- This is great, thank you for the bug fix! I am merging this now.

The contribution guidelines are outdated -- they are from a time when the architecture was under substantial revision and so any PRs would have had trouble landing. That work is now done, and we are excited to accept contributions directly now! I am updating the CONTRIBUTING.md file this morning.

PR #259, opened 2026-05-13
feat(auth): opt-in FABRO_CLI_TRUST_HOST_AUTH so claude CLI can use a host-side Max subscription

PeterBell:
## Problem

For operators on a Claude Max subscription who want to use the `cli` backend with `provider="anthropic"`, the current behavior silently bills the registered Anthropic API key per token rather than consuming Max quota. The flow:

1. Operator runs `fabro provider login --provider anthropic` and registers an API key.
2. Operator authors a workflow with `backend="cli"` (the documented path to use the inner `claude` CLI's own auth).
3. On run, Fabro's credential resolver returns `CliCredential { env_vars: { "ANTHROPIC_API_KEY": ... } }`. `launch_env.rs` injects that into the spawned subprocess. claude CLI sees `ANTHROPIC_API_KEY` in its env and prefers it over any host-side OAuth credentials it already has.
4. The run consumes API tokens, billed against the registered key, even though the operator has Max and wants to use it.

Verified on `0.232.0-nightly.0` against the latest claude CLI (2.1.77):

- claude CLI's init payload in `agent.cli.completed` reports `apiKeySource: "ANTHROPIC_API_KEY"`.
- Anthropic Console balance shows real per-token spend (confirmed at `total_cost_usd: 0.063` on a trivial single-step lookup).

This is a blocker for the subscription-only deployment shape: any operator who wants Fabro orchestration with Max-covered execution cannot get there with current Fabro.

## Short-term fix in this PR

Add an opt-in env var, `FABRO_CLI_TRUST_HOST_AUTH`. Three files, +41 lines:

- `lib/crates/fabro-static/src/env_vars.rs`: declare the constant.
- `lib/crates/fabro-server/src/spawn_env.rs`: add it to `WORKER_ENV_ALLOWLIST` so it reaches the worker that runs the resolver.
- `lib/crates/fabro-auth/src/resolve.rs`: when the flag is truthy AND the credential is `(Provider::Anthropic, AuthDetails::ApiKey, CliAgentKind::Claude)`, `to_cli_credential` returns `CliCredential { env_vars: empty, login_command: None }`.

Behavior unchanged for: any unset / falsy flag value, any provider other than Anthropic, any credential variant other than `ApiKey`, any CLI agent other than Claude. Existing API-key users see no change.

When the flag is set, claude CLI starts with no `ANTHROPIC_API_KEY` in its environment and falls back to whatever credentials it finds on the host (typically a Max subscription OAuth session at `~/.claude/`). End-to-end verification on a Max subscription:

- `apiKeySource` in the agent.cli init payload flips from `"ANTHROPIC_API_KEY"` to `"none"`.
- Anthropic Console-side token counts do not increase across patched runs.
- A registered Anthropic API key is still required to satisfy the resolver gate; the value is never propagated.

The reasoning behind the env-var-and-allowlist approach (rather than a settings.toml field): smallest possible diff, easy to reason about, easy to revert. Operator's settings.toml stays untouched. Happy to refactor to a `[providers.anthropic] cli_skip_env_credentials = true` field in `settings.toml` if you'd prefer that surface; the resolver-side logic is the same either way.

## A more robust solution we'd love to work on

This PR is the smallest correct change to unblock Max-subscription operators today. We'd be happy to follow up with a proper Anthropic OAuth credential path that mirrors the existing `CodexOAuth` shape for OpenAI:

- New `AuthDetails::AnthropicOAuth { tokens, account_id, ... }` variant alongside `AuthDetails::ApiKey`.
- `fabro provider login --provider anthropic` defaults to a browser OAuth PKCE flow (with `--api-key` fallback for headless setups), tied to the operator's Max account.
- Tokens stored in Fabro's vault; refresh handled by the existing `refresh_oauth_credential` infrastructure that already covers Codex.
- `to_cli_credential` for `(Anthropic, AnthropicOAuth, Claude)` either (a) returns an empty env_vars with a `login_command` that runs `claude /login` (or equivalent) on host-side, or (b) injects the OAuth access_token directly. Whichever matches Anthropic's actual CLI semantics; we'd dig into claude CLI's auth contract in the implementation.

The two approaches coexist cleanly. The env-var flag from this PR becomes "I'll manage claude's auth on the host myself"; the OAuth path becomes "Fabro manages it for me." Different deployment shapes call for different modes. No migration cost for operators on either path.

**Would you be open to a follow-on PR for the OAuth variant?** We have bandwidth to start on it today if there's interest. Happy to align on the credential-variant shape and the `provider login` UX before we write code, or to send a draft for review. If you'd prefer to leave the OAuth path on the roadmap and merge just this minimal env-var fix for now, that's a clean stopping point too.

## Notes

- This PR is from `gathercommunity/fabro`, a fork in the gather.community org. Filing on behalf of a small team that's building an agentic orchestration layer on Fabro for our own workflows.
- We've been running the patched binary for a few hours of testing against a real Max-subscription workflow; no surprises beyond what's documented here.
- The 3 test failures on `cargo test --release -p fabro-server --lib` in our environment (`install::tests::write_artifact_store_metadata_creates_marker_in_overridden_storage_root`, `serve::tests::build_object_store_from_settings_rejects_partial_static_credentials`, `serve::tests::build_slatedb_store_uses_configured_local_root`) reproduce on a clean `main` checkout in the same environment, so they are pre-existing environment-dependent issues unrelated to this change. Both `spawn_env` tests pass.

Thanks for Fabro. The DOT-graph model + the CLI-backend escape hatch is exactly the shape we needed.

### DIFF
--- a/lib/crates/fabro-auth/src/resolve.rs
+++ b/lib/crates/fabro-auth/src/resolve.rs
@@ -373,18 +373,31 @@ impl CredentialResolver {
         let provider = Provider::from_id(&credential.provider);
         let login_command = match (provider, &credential.details, kind) {
             // Opt-in: when FABRO_CLI_TRUST_HOST_AUTH is set, do not inject
-            // ANTHROPIC_API_KEY into the claude subprocess environment. claude
-            // CLI then falls back to its own credentials on the host (e.g. a
-            // Max subscription's OAuth session), so the registered API key is
-            // not billed for CLI-backend runs. A registered credential is
-            // still required to satisfy the resolver gate; the key value
-            // itself is never propagated to the subprocess.
+            // provider credentials into the CLI agent subprocess for the
+            // (provider, agent) pairs that support a host-side credential
+            // fallback (currently Anthropic+claude and OpenAI+codex). The CLI
+            // then resolves auth using whatever it finds on the host (e.g.
+            // ~/.claude/ for claude Max OAuth, ~/.codex/auth.json for a Codex
+            // ChatGPT Pro session). A registered credential is still required
+            // to satisfy the resolver gate; the secret material is never
+            // propagated to the subprocess and the codex_login_command
+            // (which would clobber ~/.codex/auth.json) is skipped.
             //
             // The flag must reach the worker subprocess that runs the
             // credential resolver; see WORKER_ENV_ALLOWLIST in
             // `lib/crates/fabro-server/src/spawn_env.rs`.
             (Some(Provider::Anthropic), AuthDetails::ApiKey { .. }, CliAgentKind::Claude)
-                if claude_cli_trust_host_auth() =>
+                if cli_trust_host_auth() =>
+            {
+                None
+            }
+            (Some(Provider::OpenAi), AuthDetails::ApiKey { .. }, CliAgentKind::Codex)
+                if cli_trust_host_auth() =>
+            {
+                None
+            }
+            (Some(Provider::OpenAi), AuthDetails::CodexOAuth { .. }, CliAgentKind::Codex)
+                if cli_trust_host_auth() =>
             {
                 None
             }
@@ -475,18 +488,26 @@ fn primary_api_key_env_var(provider: &ProviderId) -> Option<&'static str> {
 
 /// Read the opt-in `FABRO_CLI_TRUST_HOST_AUTH` flag.
 ///
-/// When set to a truthy value, Fabro skips injecting provider API-key env vars
-/// into the CLI agent subprocess for providers that support a host-side
-/// credential fallback (currently: Anthropic + claude). The CLI then resolves
-/// auth using whatever it finds on the host (for example, a Max subscription
-/// OAuth session under `~/.claude/`).
+/// When set to a truthy value, Fabro skips injecting provider credentials into
+/// the CLI agent subprocess for (provider, agent) pairs that support a
+/// host-side credential fallback. Currently covered:
+///
+/// - Anthropic + claude: skip `ANTHROPIC_API_KEY` injection. claude CLI then
+///   resolves auth from `~/.claude/` (e.g. a Max subscription OAuth session).
+/// - OpenAI + codex: skip `OPENAI_API_KEY` injection AND skip the
+///   `codex login --with-api-key` step that would clobber `~/.codex/auth.json`.
+///   codex CLI then resolves auth from `~/.codex/auth.json` (e.g. a ChatGPT
+///   Pro OAuth session).
+///
+/// A registered credential is still required in both cases to satisfy the
+/// resolver gate; the secret material is never propagated to the subprocess.
 ///
 /// Truthy: any non-empty value except `0` / `false` (case-insensitive).
 #[expect(
     clippy::disallowed_methods,
     reason = "Opt-in env flag for trusting host-side CLI credentials instead of injecting API keys into subprocesses."
 )]
-fn claude_cli_trust_host_auth() -> bool {
+fn cli_trust_host_auth() -> bool {
     std::env::var(EnvVars::FABRO_CLI_TRUST_HOST_AUTH)
         .ok()
         .is_some_and(|value| {
--- a/lib/crates/fabro-auth/src/resolve.rs
+++ b/lib/crates/fabro-auth/src/resolve.rs
@@ -372,6 +372,22 @@ impl CredentialResolver {
         let mut env_vars = HashMap::new();
         let provider = Provider::from_id(&credential.provider);
         let login_command = match (provider, &credential.details, kind) {
+            // Opt-in: when FABRO_CLI_TRUST_HOST_AUTH is set, do not inject
+            // ANTHROPIC_API_KEY into the claude subprocess environment. claude
+            // CLI then falls back to its own credentials on the host (e.g. a
+            // Max subscription's OAuth session), so the registered API key is
+            // not billed for CLI-backend runs. A registered credential is
+            // still required to satisfy the resolver gate; the key value
+            // itself is never propagated to the subprocess.
+            //
+            // The flag must reach the worker subprocess that runs the
+            // credential resolver; see WORKER_ENV_ALLOWLIST in
+            // `lib/crates/fabro-server/src/spawn_env.rs`.
+            (Some(Provider::Anthropic), AuthDetails::ApiKey { .. }, CliAgentKind::Claude)
+                if claude_cli_trust_host_auth() =>
+            {
+                None
+            }
             (Some(Provider::OpenAi), AuthDetails::ApiKey { key }, CliAgentKind::Codex) => {
                 env_vars.insert(EnvVars::OPENAI_API_KEY.to_string(), key.clone());
                 Some(codex_login_command(key))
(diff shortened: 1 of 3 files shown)


brynary:
Thanks @PeterBell! I'm going to dig into this while on the flight I'm on.

Longer term, I'm _hoping_ to deprecate and eventually replace `cli` mode with `acp`. Could you give `acp` a try because if we can make ACP mode use your subscription properly that's a better long term path?

PeterBell:
Perfect - once I've spiked Codex in CLI mode, I'll run a new spike for both Claude and Codex using ACP

brynary:
Great. Also if you run into issues, I would use `fabro secret rm` to delete the API key secrets so they are inaccessible to Fabro and it doesn't inadvertently use them.

Future setup improvement will be making them optional.

PeterBell:
Thanks - will do. Although honestly I just rotate them to be safe!

brynary:
@PeterBell to confirm, it sounds like with ACP working this can be closed out, is that right?

PeterBell:
Hey @brynary I think this is still required - give me a minute

PeterBell:
From Sloane: Hey Bryan, still needed. Short version:

The patch lives in fabro-auth::resolve::to_cli_credential. Both backends call into that same resolver via resolve_agent_launch_env(...), including the ACP path at lib/crates/fabro-workflow/src/handler/llm/acp.rs:90-101. Without the patch, an ACP-launched claude subprocess still gets ANTHROPIC_API_KEY in its environment and bills the API key instead of falling through to the host-side Max OAuth session.

Some ACP shims (e.g. claude-code-acp) strip provider env vars before exec’ing the underlying CLI, in which case this becomes defense-in-depth rather than load-bearing. We’d rather not rely on shim behavior, so we’d like to keep the patch.

Likely follow-up: we’ll need to extend the same gate to (Provider::OpenAi, *, CliAgentKind::Codex) so codex login --with-api-key doesn’t clobber the host’s ChatGPT auth (bit us in our Phase 1 spike). Happy to revise the env var name to something provider-agnostic at that point if you’d prefer (FABRO_CLI_TRUST_HOST_AUTH already covers it semantically, but claude_cli_trust_host_auth() the helper does not).

brynary:
OK, thanks Peter. I will take a closer look.

To ensure the system stays well architected, I think I need to answer the question of "How should this work?" which will take some thinking, and then from there it should illuminate the right code change.

PeterBell:
Yeah - this is a patch not the right longer term solution - I think it should work same way as it does for OpenAI - I'll share one possible approach below from Sloane - happy to submit a patch for the bigger piece, but wanted to stay surgical for now...

brynary:
@PeterBell I think you may have explained this to me already, however I am have a difficult time remembering... What happens if you remove `ANTHROPIC_API_KEY` from Fabro?

My understanding is that most of the agent CLIs have a place they store credentials (e.g. JSON on disk, system keychain, etc.). Also, it is often conventional for a CLI program to allow environment based overrides, e.g. I would expect that running `export ANTHROPIC_API_KEY=... claude` to prefer `$ANTHROPIC_API_KEY` over the credentials it has stored.

So the expected behavior (Fabro aside) is generally: Agents prefer tokens in the environment is present, otherwise they Agents use their credentials on disk. It feels like Fabro should follow the same pattern.

In other words, providing a `$ANTHROPIC_API_KEY` to Fabro should mean "I opt-in to using this Anthropic API key for Fabro's usage of Claude". And omitting the `$ANTHROPIC_API_KEY` would be all that is necessary to ensure Fabro uses Anthropic subscription usage.

Thoughts?

PeterBell:
Hey Bryan, good question...

You’re right about the general CLI convention: ANTHROPIC_API_KEY in env wins, otherwise fall through to whatever the agent has stored on disk (~/.claude/ OAuth session for claude code). That’s exactly how claude code behaves when invoked directly.

The wrinkle is in how Fabro currently wires things together:

To dispatch a Claude worker, Fabro requires a registered provider credential. You run fabro provider login --provider anthropic ... once. The credential lands in Fabro’s own store. With no credential registered, the resolver refuses to dispatch.

At spawn time, to_cli_credential in lib/crates/fabro-auth/src/resolve.rs returns a CliCredential whose env_vars field carries ANTHROPIC_API_KEY=<the registered key>.

Workers run with cmd.env_clear() plus an explicit WORKER_ENV_ALLOWLIST (fail-closed), and the resolver’s env_vars are injected into that scrubbed env. So inside the worker, ANTHROPIC_API_KEY is always set, regardless of what was in the host shell when Fabro started.

So from claude code’s perspective inside the worker, the env var is always present, so it always wins, and the ~/.claude/ Max OAuth session never gets called. There’s no way today to say “ignore the registered credential for this run.”

So the patch creates a minimal opt-in escape hatch. When FABRO_CLI_TRUST_HOST_AUTH=1 is set, the resolver still validates that a provider credential exists (gate stays in place), but to_cli_credential for the (Anthropic, ApiKey, Claude) arm returns empty env_vars and no login_command. The worker spawns with no ANTHROPIC_API_KEY in its env, claude code falls through to its host-side OAuth, and the registered key is never billed.

The behavior you’re describing, “no ANTHROPIC_API_KEY registered means defer to host auth,” is the cleaner end state and I agree it’s where Fabro should land. It’s a bigger change though: today the resolver treats credential absence as a hard error, not as “pass through.” The 41-line patch is a deliberately small, opt-in step we could ship behind a flag while the larger refactor gets designed. If you’d rather we make registration optional and rewire the resolver to fall through on absence, happy to take a swing at that path instead. The flag-gated version was our way of not changing default behavior for anyone else on Fabro.

One adjacent note: the patch may want a sibling arm for (OpenAi, *, Codex) for the same reason. Fabro’s CodexOAuth path actually overwrites ~/.codex/auth.json with a scoped-down access token during login, which broke direct codex CLI usage on my laptop. Same env-injection problem, different blast radius. Happy to extend the PR or split it.

From Sloane

brynary:
Thanks, Peter. The problem around the hard error if anthropic is "not configured" seems like the root here. I'm going to see if we can simplify that in such a way that it enables this.

The "not configured" error applies to _API_-based Claude usage but is not relevant for `claude` ACP and CLI usage. If there is no env var, and Claude is not authenticated in the sandbox, then `claude` will just error when it starts up which is fine.

natea:
a simple workaround that i found was to add in the `ANTHROPIC_API_KEY` in `fabro.toml` an environment variable:

```
[run.sandbox.env]
ANTHROPIC_API_KEY = ""      
```

This satisfies Fabro's requirement that there's an environment variable defined, but then when `claude -p` sees that the value is null, it falls back to use the Claude Max subscription (via auth token) rather than the API key approach.

Note, this requires using `backend: cli;` in the stylesheet definition in `fabro.workflow` file:

```
        model_stylesheet="
            *         { model: claude-sonnet-4-6; backend: cli; }
            .analysis { model: claude-sonnet-4-6; reasoning_effort: high; backend: cli; }
        "
```

brynary:
OK, I went with a massive simplification to the backend which rooted out the original sin which caused this bug. There is a test proving the fix here:

https://github.com/fabro-sh/fabro/pull/307/changes/2c1410e1893af3188252f7d8aaaa349f91e21011#diff-48a94ad42d92ee47a8bc8c34482604a668042dc3d77413bb81d7292b59be3445R88

That PR supersedes this one, so moving over there.

PeterBell:
Awesome - thanks @brynary - I'll pull and test [feat/api-acp-backend-contract](https://github.com/fabro-sh/fabro/tree/feat/api-acp-backend-contract) later today!

PR #301, opened 2026-05-17
fix: MiniJinja can't find partials

salieri:
Fixes an issue where use of MiniJinja [`include`](https://jinja.palletsprojects.com/en/stable/templates/#include) control structure (`{% include "filename.ext" %}`) causes a render error `template not found: tried to include non-existing template "filename.ext"`

### Example broken diagram
``` dot
digraph ValidatePlan {
    start [shape=Mdiamond, label="Start"]
    exit  [shape=Msquare, label="Exit"]

    test_inline_prompt [label="moo" prompt="{% include 'test.tpl.md' %}"]
                                          // ^^^^^^^^^^^^^^^^^^^^^^^^^
    start -> test_inline_prompt -> exit
}
```

### Fix
The core issue was that template rendering knew the source name for diagnostics, but did not have a loader rooted at the prompt/goal file location. Includes therefore failed even when the included file existed next to the rendered file. The fix adds optional loader support to `fabro-template`, then wires workflow rendering to the existing `FileResolver` so includes resolve relative to the file currently being rendered.

For `fabro validate`, there was a second manifest-specific problem: validation runs through a bundled manifest, and the manifest builder only bundled explicit `prompt.md` / `goal.md` files, not static MiniJinja `include` dependencies inside those files. The manifest builder now scans prompt/goal template text for literal `{% include "file" %}` / `{% include 'file' %}` references and bundles those files too. Missing or unsafe include names are left for MiniJinja/runtime validation rather than expanding scope.

(For clarity: The fix does not support variables or arrays in `include`.)


brynary:
Hi @salieri -- Thanks for putting this fix together. The core issue you identified was right: MiniJinja templates that use `{% include %}` need a loader rooted at the
  workflow/template location, not just string rendering in isolation.

  I built on your fix in a few ways so the behavior is covered across the full manifest/runtime path:

  - Added rooted template stores for filesystem and bundled manifest assets.
  - Added MiniJinja AST-based dependency discovery for `include`, `extends`, `import`, and `from import`.
  - Made manifest building bundle all statically discoverable template dependencies, including dependencies inside branches and macro bodies.
  - Rejected dynamic template dependencies during manifest build, because we cannot know which asset to bundle safely unless it is explicit.
  - Moved `ManifestPath` into `fabro-types` so template, manifest, workflow, and server code share one logical path model.

  The main reason for this extra structure is that the server/runtime only has access to files included in the run manifest. A template can render locally
  with a loader and still fail later if an included file was never bundled. The new discovery step catches that at manifest/validate time and makes the
  runtime path deterministic.

  I also reorganized the CLI fixtures under `test/templates/` so the command-level tests stay focused on user-visible behavior, while the lower-level
  MiniJinja mechanics are covered in `fabro-template` and `fabro-manifest`.

  Thanks again for the initial fix! It gave us the right direction and exposed the broader architecture gap around rooted template dependencies.

salieri:
Cool! Happy it helped. >)

PR #302, opened 2026-05-18
fix(server): persist manifest metadata names

bokmann:
When Fabro creates a detached run through the server, the resulting run metadata should still read like something a human can trust at a glance. Before this change, those runs could persist with `settings.project.name` and `settings.workflow.name` left `null` even though Fabro already had enough local context to infer them. That made `inspect` output look half-populated and made it harder to tell whether the saved run state was complete.

This fixes that trust gap in the server-backed manifest flow.

## Summary
- backfill missing manifest-backed project and workflow names during server run preparation
- prefer explicit `[workflow].name` from bundled `workflow.toml`, then fall back to graph name or workflow slug
- cover both manifest preparation and persisted run-state behavior with server tests

## Testing
- cargo test -p fabro-server prepare_manifest_backfills_missing_project_and_workflow_names -- --nocapture
- cargo test -p fabro-server prepare_manifest_preserves_explicit_project_and_workflow_names -- --nocapture
- cargo test -p fabro-server create_run_persists_backfilled_project_and_workflow_names -- --nocapture


brynary:
Thanks again for the thoughtful work here. Your change surfaced the real issue: server manifest runs were losing explicit [workflow] metadata from bundled `workflow.toml`, and that needed to be preserved.

I built on top of that by keeping the metadata preservation, but removing the synthetic-name backfill behavior. In particular, we now persist explicit workflow/project names when they are provided, but we no longer infer project names from paths, workflow names from graph names, or names from slugs. Graph names and slugs are still useful facts, so the API now carries workflow.graph_name separately and allows workflow.name to be null.

In addition to capturing the workflow slug, name, and graph name, we also have capture the `source_directory` and repository `origin_url` when present.

The concept of a "Project" was added prematurely by me and probably should be removed. A Project was contemplated to be a Repository plus a Branch and a root path what would be registered with the server. I thought we were going to need that, however it has not turned out to the be the case so far. So I'm going to clean that up and clean up all the project API surface soon unless that changes.

The reason for that split is to keep stored/API data truthful. UI, CLI, and MCP output can still display a helpful fallback like workflow name, graph name, slug, or -, but those fallbacks should stay presentation-only rather than being persisted as real names. This keeps the data model cleaner while preserving the user-facing behavior.

PR #315, opened 2026-05-19
fix(cli): print full run id on rm

bokmann:
When `fabro rm` succeeds, the confirmation output should identify exactly which run was removed. Today the human-readable path prints a shortened run ID, which is less precise than the JSON output and less useful for copy/paste confirmation.

## Summary
- print the full run ID after successful `fabro rm` removal
- keep `--json` behavior unchanged
- update CLI snapshots to expect full IDs on success paths

## Testing
- cargo test -p fabro-cli rm_ -- --nocapture


brynary:
Thanks!

PR #370, opened 2026-05-23
MCP subprocesses do not receive interpolated env from `project.toml` or sandbox env

swerner:
## Summary

While trying to give a custom Gemini MCP a real `GEMINI_API_KEY`, I ran into
the same wall on every documented `project.toml` path — templates leaked
through as literal strings, `[run.sandbox.env]` never reached the MCP, and
`fabro secret set --type environment` populated the vault but nothing read it
on MCP launch. Wrapper-script-with-sibling-`.env` is the only thing that
I could get working today.

This PR is partly a fix and partly a discussion-starter: it adds a single failing test
per entry point (so we can talk about each one concretely), and folds in the
one cheap fix that doesn't require any product decisions. The remaining red
tests are placeholders for that discussion — see **Conversation** below.

## Changes

- New: `lib/crates/fabro-mcp/tests/env_propagation.rs` — 6 black-box tests that
  exercise different ways to feed an env var to an MCP. Each test parses
  a project.toml snippet through the public `WorkflowSettingsBuilder::from_toml`
  API, spawns the existing `test_mcp_server.py` via `McpClient`, and asks the
  subprocess via the `__env:KEY__` echo sentinel what env it actually saw.
- Fix: `lib/crates/fabro-config/src/resolve/run.rs` — `resolve_mcp_entry` was
  collapsing every interpolated MCP field via `InterpString::as_source()`, so
  `{{ env.X }}` templates leaked through as literal strings. New
  `resolve_or_source` helper mirrors `fabro_workflow::operations::start::resolve_interp`:
  try `InterpString::resolve(process_env_var)`, fall back to `as_source()` when
  the env var is unset (so existing tests that don't seed the env still pass).
  Applies to `Stdio.env`, `Sandbox.env`, `Http.url`, `Http.headers`, and the
  script/command args of both stdio and sandbox transports.
- Dev-deps: `fabro-vault` and `tempfile` added to `fabro-mcp/Cargo.toml` for
  the Finding 3 test.

## Verification

```
$ cargo test -p fabro-mcp --test env_propagation

mcp_env_table_with_literal_value_propagates_to_subprocess              PASS  baseline
mcp_env_table_with_env_template_resolves_in_subprocess                 PASS  F1 stdio (fixed)
sandbox_mcp_env_table_with_env_template_resolves_at_config_layer       PASS  F1 sandbox (fixed)
sandbox_env_with_literal_value_propagates_to_mcp_subprocess            FAIL  see Conversation
sandbox_env_with_env_template_propagates_to_mcp_subprocess             FAIL  see Conversation
fabro_secret_environment_type_propagates_to_mcp_subprocess             FAIL  see Conversation
```

Also confirmed `cargo test -p fabro-config resolve_preserves_source_templates_for_mcp_and_hook_strings`
still passes — the fallback semantics preserve the existing behavior whenever
the referenced env var isn't set.

## Conversation

The three failing tests below are deliberate placeholders. I'd love feedback on
whether each is a bug we should fix here, or whether it warrants a product
decision first.

### `sandbox_env_with_literal_value_propagates_to_mcp_subprocess` (Finding 2, literal)

```
left:  ""
right: "finding-2-literal"
```

`docs/public/execution/run-configuration.mdx` documents `[run.sandbox.env]`
as the canonical run-wide env knob, with no caveat about MCPs. Today the
resolved sandbox env is fed into `WorkflowToolEnvProvider` for agent tools
and never threaded into `runtime_mcp_server` at
`lib/crates/fabro-workflow/src/operations/start.rs:650-673`. Is this intentional? Solution would be to merge `SandboxEnvSpec.toml_env` into the MCP transport's
env before handing to `McpClient`, with the MCP-local table winning on
conflict. **Fix here, or split into a follow-up?**

### `sandbox_env_with_env_template_propagates_to_mcp_subprocess` (Finding 2, interpolated)

Same gap, with the additional layer that the user wrote `{{ env.X }}`. Fixed
by the same wiring change as above — the template resolution already happens
upstream in `build_sandbox_env`.

### `fabro_secret_environment_type_propagates_to_mcp_subprocess` (Finding 3)

```
left:  ""
right: "finding-3-secret"
note:  vault contains TARGET=Some("finding-3-secret"), but nothing in the
       MCP launch path consults it.
```

This is the one that probably wants a product decision. Today `Vault::snapshot()`
(the function that filters `SecretType::Environment`) has zero production
callers — secrets are read by name via `Vault::get(name)` for LLM provider
auth only (`fabro-auth/src/resolve.rs:299`), so is `--type environment`
really intended as a storage-shape tag rather than a behavioral switch?

Reasonable options:
1. Auto-inject all environment-type secrets into MCP env. Easy, but every MCP would see every secret.
2. Opt-in scoping, e.g. `[run.agent.mcps.X.secrets] = ["GEMINI_API_KEY"]`.
3. A new interpolation token, e.g. `{{ secret.X }}`, that resolves from the vault — keeps explicit naming, parallels `{{ env.X }}`.

I'm partial to (2) or (3) but happy to defer to whatever you think fits
Fabro's secrets model.

---

Happy to iterate, split into smaller PRs, or rebase any of this. Mostly
hoping the failing tests give us a concrete shared artifact to talk against.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2098,10 +2098,12 @@ dependencies = [
  "fabro-config",
  "fabro-http",
  "fabro-types",
+ "fabro-vault",
  "futures",
  "rmcp",
  "serde",
  "serde_json",
+ "tempfile",
  "tokio",
  "tracing",
 ]
--- a/lib/crates/fabro-config/src/resolve/run.rs
+++ b/lib/crates/fabro-config/src/resolve/run.rs
@@ -370,14 +370,14 @@ pub(crate) fn resolve_mcp_entry(name: &str, entry: &McpEntryLayer) -> McpServerS
             command: resolve_mcp_command(script.as_ref(), command.as_ref()),
             env:     env
                 .iter()
-                .map(|(key, value)| (key.clone(), value.as_source()))
+                .map(|(key, value)| (key.clone(), resolve_or_source(value)))
                 .collect(),
         },
         McpEntryLayer::Http { url, headers, .. } => McpTransport::Http {
-            url:     url.as_source(),
+            url:     resolve_or_source(url),
             headers: headers
                 .iter()
-                .map(|(key, value)| (key.clone(), value.as_source()))
+                .map(|(key, value)| (key.clone(), resolve_or_source(value)))
                 .collect(),
         },
         McpEntryLayer::Sandbox {
@@ -391,7 +391,7 @@ pub(crate) fn resolve_mcp_entry(name: &str, entry: &McpEntryLayer) -> McpServerS
             port:    *port,
             env:     env
                 .iter()
-                .map(|(key, value)| (key.clone(), value.as_source()))
+                .map(|(key, value)| (key.clone(), resolve_or_source(value)))
                 .collect(),
         },
     };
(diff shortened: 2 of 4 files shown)


brynary:
Hi @swerner -- Thanks for this. The key here to get right is going to be how the security model should work. There's enough inherent complexity that it's easier to get that wrong than write, and a risk of yielding painful vulnerability risk and/or backwards compatibility breakages down the road.

Let's plan for this to be a whiteboard topic this week!

swerner:
@brynary sounds good! Figured it would be a longer discussion, but wanted to get it pushed up at least. It's been a helpful task for getting a sense of how everything works.

brynary:
@swerner so for this one, as discussed today IRL, I think the next step is to zoom out and make a behavior table for which settings strings:

1. Support interpolation at all
2. Support file-inlining `@` refs
3. For strings supporting interpolation, which vars do they support

Once we decide on that, we can do a larger change to resolve ambiguities, converge behavior, and document

swerner:
Closing this as we've discussed - working to untangle all the config variables and unify the behavior. First two PRs for laying the foundation are up at #474 and #472 with more to come.

PR #383, opened 2026-05-24
fix(skills): Ensure the local server can access project skills

cpennington:
The local server wasn't setting git_root, and wasn't defaulting to the workspace working directory, so it didn't find project-specific skills.


cpennington:
N.B. I haven't been able to test this locally (the project is failing to build in the amount of memory I have locally)

brynary:
@cpennington -- Thanks for this. The fix looks correct for me, and I've added an integration test. Will finish cleanup shortly then merge.

PR #386, opened 2026-05-24
Add legacy SSE MCP transport support

fkukuck:
## Summary

Adds support for MCP servers that use the SSE-based HTTP transport, including Playwright MCP.

Fabro already supports stdio and Streamable HTTP MCP servers. Some MCP servers still expose the SSE transport shape where the client opens an SSE stream, receives an `endpoint` event, and sends JSON-RPC requests back to that endpoint. This PR adds an explicit `protocol = "sse"` option while keeping Streamable HTTP as the default.

## What Changed

- Added `McpHttpProtocol` with `streamable_http` as the default and `sse` as an opt-in protocol.
- Added an SSE MCP client transport implementation.
- Wired HTTP MCP setup to choose Streamable HTTP or SSE based on config.
- Added `protocol = "sse"` support for both `http` and `sandbox` MCP entries.
- Updated sandbox MCP resolution so SSE sandbox servers connect through the preview `/sse` path.
- Documented `protocol = "sse"` for Playwright MCP.
- Added an integration test covering SSE initialize, tool listing, and tool calls.

## Example

```toml
[run.agent.mcps.playwright]
type = "sandbox"
protocol = "sse"
command = ["npx", "@playwright/mcp@latest", "--port", "3100", "--headless", "--browser", "chromium"]
port = 3100
startup_timeout = "60s"
tool_timeout = "2m"
```

## Compatibility

Existing MCP configs are unchanged because `protocol` defaults to `streamable_http`.

## Validation

- `cargo nextest run -p fabro-mcp`
- `cargo check -p fabro-agent -p fabro-workflow -p fabro-config`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2099,6 +2099,7 @@ name = "fabro-mcp"
 version = "0.243.0-nightly.1"
 dependencies = [
  "anyhow",
+ "axum",
  "fabro-config",
  "fabro-http",
  "fabro-types",
@@ -2107,6 +2108,7 @@ dependencies = [
  "serde",
  "serde_json",
  "tokio",
+ "tokio-stream",
  "tracing",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2107,6 +2107,8 @@ dependencies = [
  "rmcp",
  "serde",
  "serde_json",
+ "sse-stream",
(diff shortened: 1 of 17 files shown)


brynary:
Thanks @fkukuck! This looks good to me, just doing a few more checks and then will merge.

PR #414, opened 2026-05-26
fix(server): use checkpoint patch for active run files fallback

fkukuck:
## Summary

- Fixes the degraded Files Changed response for active runs.
- Uses the latest checkpoint diff patch when a terminal conclusion patch does not exist yet.
- Preserves the existing terminal conclusion patch preference for completed or failed runs.

## When This Happens

This fixes active runs where the run summary already has checkpoint diff stats, but the Files Changed endpoint cannot use the live sandbox diff path and must fall back to stored patch data.

Examples include cases where sandbox reconnect is temporarily unavailable, the sandbox has already disappeared, or git diff range materialization cannot read the required objects from the sandbox. Before this change, the fallback only looked at `conclusion.diff.patch`, so in-progress runs returned an empty file list until the run completed.

### DIFF
--- a/lib/crates/fabro-server/src/run_files.rs
+++ b/lib/crates/fabro-server/src/run_files.rs
@@ -777,39 +777,55 @@ async fn sandbox_git_stdout(
     Ok(res.stdout)
 }
 
-/// Build the degraded response from the stored terminal diff patch.
-/// When `conclusion.diff.patch` is `None`, returns the empty envelope (UI maps
-/// this to R4(c)). Keeps the same `FileDiff[]` shape as live responses, but
-/// leaves contents unavailable because the server only has a unified patch.
+/// Build the degraded response from a stored diff patch.
+/// Prefers the terminal conclusion patch when present, and falls back to the
+/// latest checkpoint patch for in-progress runs. When no patch is available,
+/// returns the empty envelope (UI maps this to R4(c)). Keeps the same
+/// `FileDiff[]` shape as live responses, but leaves contents unavailable
+/// because the server only has a unified patch.
 fn build_fallback_response(
     projection: &fabro_store::RunProjection,
     reason: RunFilesMetaDegradedReason,
     run_id: &RunId,
     start: Instant,
 ) -> PaginatedRunFileList {
-    let Some(patch) = projection
+    let stored_patch = projection
         .conclusion
         .as_ref()
-        .and_then(|conclusion| conclusion.diff.patch.as_deref())
-    else {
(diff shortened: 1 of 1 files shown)


brynary:
Thanks, @fkukuck. Codex provided this analysis:

> The fallback now uses only the latest `checkpoint.diff.patch,` but checkpoint patches are produced as _deltas from the previous checkpoint_, not from the run base.
>
> `GitLifecycle` computes checkpoint patches from `last_git_sha.or(base_sha)` to `HEAD`, then updates `last_git_sha`. So after a run has multiple checkpoints, this fallback will omit files changed in earlier checkpoints and meta.total_changed / stats will describe only the last checkpoint delta.

Can you shed some more light on the real world conditions you are running into when the fallback is needed?

It is true it would be needed if the sandbox is stopped, but for my normal day to day the sandboxes restart fast enough it is not an issue.

fkukuck:
This is about transient read-side sandbox failures, not necessarily a lost sandbox. The workflow can keep running on its existing sandbox handle while the Files Changed tab, which reconnects separately and runs git inspection, temporarily fails and falls back.

Your point about checkpoint patches being deltas is correct. The current PR should not show the latest checkpoint delta as the full run diff.

Options I see going forward:
1. Drop this PR if this is a non-issue in practice. The scenario in which this occurs in my fork succeeds now (with a fix in my fork) as long as live sandbox materialization succeeds; the remaining gap is only degraded active-run reads before the terminal patch exists.
1. Store/use cumulative checkpoint patches from `base_sha..checkpoint_sha`. Most complete, but broader change.
1. Use checkpoint fallback only when there is exactly one checkpoint patch; otherwise keep the existing empty degraded response until conclusion. Smaller and safe, but partial.

Would you like me to come up with a follow up commit, or do we deem it unworthy to pursue for now? I really don't want to block with small priority stuff.

brynary:
OK, thanks for the extra color. I am in meetings most of today so I’ll
circle back to this tonight or tomorrow AM. Cheers, -Bryan

brynary:
Hi @fkukuck -- OK, after considering this, I recommend we go with option one for now:

> Drop this PR if this is a non-issue in practice.

I think we should re-evaluate if/when we have more data collected from future fallback cases to help inform us and decide which direction to choose at that time. Until then, I think we should limit the inherent complexity and accept that in a sandbox-temporarily-unavailable case then the diff also will not be available via the Files Changed tab.

fkukuck:
Agreed

PR #438, opened 2026-05-27
Add OpenRouter as an opt-in built-in provider

swerner:
## Summary

- Ships OpenRouter as a built-in catalog provider, disabled by default, with a curated 17-model catalog (Anthropic, OpenAI, Gemini, plus 10 popular open-weights models).
- Enable with one section in `~/.fabro/settings.toml` (`[llm.providers.openrouter] enabled = true`) plus configuring an `OPENROUTER_API_KEY`. No further configuration needed — models, base URL, and adapter are pre-wired.
- Reuses the existing `OpenAiCompatibleAdapter` via `adapter = "openai_compatible"` plus per-provider naming, so no new adapter file. Provider-specific request fields are accessible through the existing `provider_options.openrouter.*` convention.
- Extends `ApiUsage` parsing in `openai_compatible.rs` to additively read `prompt_tokens_details.cached_tokens` and `cache_write_tokens`. Generic improvement that also benefits any other OpenAI-compatible provider that emits these.
- Adds a dedicated gitleaks rule for `sk-or-v1-*` keys (the entropy-based redactor would otherwise miss them — pure hex tops out at 4.0 bits/byte, below the 4.5 threshold).
- Attribution headers (`HTTP-Referer`, `X-Title`) are NOT sent by default — self-hosted installations stay anonymous on OR's public leaderboard unless the operator opts in via `extra_headers`.

## Test plan

- [x] `cargo test -p fabro-model --lib` — 119 passing, including new `builtin_openrouter_provider_is_opt_in`
- [x] `cargo test -p fabro-llm --lib providers::openai_compatible` — 33 passing, including new streaming + non-streaming cached-token tests
- [x] `cargo test -p fabro-redact --lib` — passing, including new `redact_string_openrouter_api_key`
- [x] `cargo +nightly-2026-04-14 fmt --check --all` — clean
- [x] `cargo +nightly-2026-04-14 clippy --workspace --all-targets -- -D warnings` — clean
- [x] Manual: `fabro model list --provider openrouter` shows all 17 models
- [x] Manual: `fabro model test --model anthropic/claude-haiku-4-5` completes against live OR API
- [x] Live e2e: `OPENROUTER_API_KEY=... cargo test -p fabro-llm --test integration openrouter_complete -- --include-ignored` (gated, costs ~$0.001)

## Discussion points

A few decisions worth raising for the review conversation, each defensible but with a reasonable alternative.

1. **Attribution headers default to opt-in.** Self-hosted installations don't auto-advertise on OR's public leaderboard. The alternative (default to attribution with an opt-out) would give Fabro more leaderboard presence at the cost of installations leaking their existence. Privacy-by-default felt like the right call but happy to flip.

2. **`reasoning_effort` is silently dropped for OR-routed models.** Users wanting OR's unified reasoning control use `provider_options.openrouter.reasoning = { effort = "high" }` via the API; the typed `Request.reasoning_effort` flag is a no-op for any model routed through OR. Documented in the integration page. Alternative: translate it in the openai_compatible adapter (couples generic adapter to OR-specific shape) or emit a `tracing::warn!` once when set. Open to either.

3. **`billing_policy` defaults to OpenAi math for all 17 models.** `billing_policy` selects the billing algorithm (which tokens get billed and what derived prices apply). For Claude-via-OR under the default OpenAi algorithm, cache reads and regular input/output are billed correctly; cache *writes* are not billed at all (the OpenAi algorithm has no cache_write concept). Drift is exactly zero for non-caching workflows and meaningful only for heavy Anthropic-prompt-cache usage routed through OR. Fixing requires per-model `billing_policy` overrides, which `ModelCatalogSettings` doesn't currently allow — setting provider-level to `anthropic` would break the other 14 entries. Open to expanding scope to add the schema field, otherwise happy to file as a follow-up alongside the deferred `Response.cost_usd` work in point 5.

4. **`family = "openrouter"` for all 17 models.** Single value across the catalog (was per-vendor `or-claude-4` etc. mid-development). Trade-off: lose the ability for a future UI feature to group "all Claude 4 across providers" via the `family` field alone. Alternative: match direct-provider family names (`claude-4`, `gpt-5`, etc.).

5. **`provider_options` workflow-attr wiring deferred.** `provider_options` is settable via `POST /api/v1/completions` and the Rust/TypeScript SDKs, but the workflow LLM handler hardcodes `provider_options: None` at `fabro-workflow/src/handler/llm/api.rs:1078` and no Graphviz attr or workflow.toml field parses into it. So OR routing controls (provider sort, fallback models, transforms, plugins) aren't reachable from `workflow.fabro`/`workflow.toml` today. (Note: Anthropic thinking is reachable via the typed `reasoning_effort` attr, which translates to budget_tokens in the Anthropic adapter — different path. Gemini's `cached_content` and any other Gemini extras share the same workflow gap as OR routing.) Documented as a limitation in the integration page. Adding a generic `provider_options` workflow-attr parser is cross-cutting (~30–60 lines, benefits OR routing + Gemini cached_content + Anthropic non-thinking extras) and probably belongs in its own PR.

6. **No `Response.cost_usd` field.** OR returns inline `usage.cost` (authoritative billing) but adding the field across `Response` would require ~30 mechanical constructor updates. Catalog-derived cost (`tokens × catalog_price`) is close enough for OSS use; deferred to a focused follow-up if/when hosted billing accuracy becomes load-bearing.


brynary:
@swerner thanks! Let's discuss the open questions at your next conveience

swerner:
Closing related to what @brynary mentioned in #459 - we're doing a preparatory refactoring to make adding new multi-model providers much easier going forward. First PR of the refactor is #471

PR #450, opened 2026-05-28
feat: chat-driven workflow builder at /playground

swerner:
## Summary

Adds a new `/playground` route where users build a Fabro workflow by
chatting with Ask Fabro on the right while watching a live canvas
re-render on the left. The workflow can be downloaded as a `.fabro.zip`
or — eventually — launched as a real Fabro run; today the "Run for
real" button POSTs to `/api/v1/runs` and redirects to the resulting
`/runs/{id}` page, with a placeholder project/repo/folder picker.

The feature is built as a standalone component subtree under
`apps/fabro-web/app/components/playground/` with no `AppShell` or
`react-router` dependencies, so it can be re-embedded in other contexts
later by passing `chatEndpoint`, `authMode`, and an optional
`realRunRedirect` prop.

## What changed

**Frontend (`apps/fabro-web/`)**

- New `/playground` route + `<Playground>` component tree.
- Live SVG canvas via `@viz-js/viz` with click-to-inspect (read-only
  node detail panel), pan, zoom, fit-to-window, and a simulated walk
  through the graph driven by a Play button.
- Docked chat sidebar (assistant-ui) wired to the new
  `/api/v1/playground/chat` endpoint, with auto-retry on parse failure
  and a playground-specific tool-call summary that reads
  `Wrote workflow.fabro (N nodes, M edges)`.
- File tabs (`workflow.fabro` / `workflow.toml` / `README.md`),
  `.fabro.zip` download via `fflate`, and a "Run for real" toolbar
  button that POSTs an inline `RunManifest` to `/api/v1/runs`.
- Draft persists across page refreshes via `localStorage`.

**Backend (`lib/crates/fabro-server/`)**

- New `POST /api/v1/playground/chat` SSE endpoint. Server is stateless
  across turns: each request carries the full draft, the server runs
  the LLM with a single `write_workflow_file` tool, streams
  `StreamEvent` frames back, and lets the client own diffing/animating
  the result into the canvas.
- Request-size caps before the LLM call (50 messages, 100 nodes, 200
  edges) so a misbehaving or malicious client can't drag multi-MB
  transcripts through token billing.

**Spec / wire contract**

- OpenAPI: new `playground/chat` operation + four new schemas
  (`CreatePlaygroundChatRequest`, `PlaygroundWorkflowDraft`,
  `PlaygroundWorkflowNode`, `PlaygroundWorkflowEdge`).
- `lib/packages/fabro-api-client` not regenerated yet (the playground
  uses raw `fetch`); reviewers who want the TS client to pick up the
  new types can run `bun run generate` in that package.

## Key design decisions

1. **Single `write_workflow_file` tool, not six per-op tools.** The
   first cut exposed `add_node`/`update_node`/`connect`/etc. as
   discrete tool calls. The model would routinely add nodes without
   wiring them up, leaving the canvas in a broken half-state. Pivoted
   to a single tool that takes the full new `workflow.fabro` content;
   the browser parses the DOT, diffs it against the local draft, and
   animates the resulting reducer ops in. The model only has to "get
   the file right", and the canvas still paints node-by-node thanks
   to the client-side animator.

2. **Stateless server.** Each chat turn POSTs the full current draft;
   nothing is persisted server-side. Keeps the endpoint cheap, makes
   refresh-resumption trivial (browser owns the truth), and means the
   same endpoint can later sit behind a rate-limited anonymous variant
   without growing per-session state.

3. **Standalone component subtree.** `<Playground>` has no
   `AppShell`/router/store dependencies. All cross-cutting concerns
   flow in as props (`chatEndpoint`, `authMode`, `realRunRedirect`).
   This is the structural hook that makes future re-embedding possible
   without a refactor.

4. **Chat is the only mutation path.** Click-to-inspect on the canvas
   is read-only. Bi-directional canvas editing was explicitly cut from
   scope to keep one source of truth for "how the workflow changed."

5. **Inline `RunManifest` instead of temp-dir-then-clone.** The
   playground has no project to run against, so the `Run for real`
   modal builds a `RunManifest` that carries the full DOT and
   `workflow.toml` source inline (`workflows[key].{source, config}`).
   `cwd` is pinned to a fixed `/tmp/fabro-playground` constant — no
   LLM-controlled segment in a filesystem-looking field.

6. **React effects policy compliance.** All `useEffect` calls in
   playground component code go through the existing primitives in
   `app/hooks/effects.ts` (`useDocumentEvent`, `useInterval`) or a
   purpose-named hook (`useCanvasRender`).

## Still outstanding (planned follow-ups)

- [ ] **Actually kicking off the ad-hoc run.** "Run for real" today
      POSTs a manifest with a placeholder project/repo/folder
      fieldset. The intent is to reuse the project-picker pattern
      being introduced on the in-flight automations branch — once
      that pattern lands, the disabled inputs in
      `run-for-real-modal.tsx` become the live surface.
- [ ] **Header link to `/playground`.** No nav entry yet; users have
      to type the URL directly.
- [ ] **Live SSE-driven canvas overlay** via
      `GET /api/v1/runs/{id}/attach` — currently the modal redirects
      to the standard run-view page; the "watch it build on the
      playground canvas" experience comes when the `stage.*` events
      are wired through.
- [ ] **Regenerate `lib/packages/fabro-api-client`** so the new types
      ship to TS consumers.
- [ ] **Smoke test:** end-to-end download → unzip →
      `fabro run <name>` round-trip.
- [ ] **`scripts/build.ts` dist-symlink bug:** `pruneOldBuilds` can
      delete the directory `apps/fabro-web/dist` points at, which
      pins the dev server in 503 "build in progress" forever.
      Workaround documented; the real fix is a separate PR.

## Test plan

- [ ] `cd apps/fabro-web && bun run test app/components/playground/` — 111 tests pass
- [ ] `cd apps/fabro-web && bun run typecheck` — clean
- [ ] `cargo test -p fabro-server playground` — 6 tests pass
- [ ] Visit `/playground`; the canvas renders the welcome `start → ??? → exit` ghost.
- [ ] Type "build me a release-notes workflow" in chat; nodes/edges animate in; ack reads `Wrote workflow.fabro (N nodes, M edges)`.
- [ ] Click a node → inspector panel populates; click empty canvas → deselects.
- [ ] Click `Simulate`; nodes light up `start → ... → exit` along the resolved path.
- [ ] Click `Download .fabro`; unzip; `cd <unzipped> && fabro run <name>` runs locally.
- [ ] Click `Run for real` → modal opens → confirm → POST succeeds → redirected to `/runs/{id}` → run executes.
- [ ] Refresh the page; the draft persists from localStorage.
- [ ] Click `Start over` → `Yes`; canvas resets to welcome state.


docs/public/api-reference/fabro-api.yaml
```diff
@@ -7411,6 +7449,95 @@ components:
...
+      type: object
+      required: [messages, workflow]
+      properties:
+        messages:
+          type: array
+          description: assistant-ui-style message history for the turn.
+          items:
+            $ref: "#/components/schemas/CompletionMessage"
+        workflow:
+          $ref: "#/components/schemas/PlaygroundWorkflowDraft"
```
brynary: RE: this new `POST /api/v1/playground/chat` endpoint -- I am wondering how much this could share with the pre-existing Ask Fabro API.

In other words, Ask Fabro sessions are roughly just agent messages, user messages and tool calls. Is that enough to power Playground if e.g. `write_workflow_file` was just a regular tool -- _or_ if it became `Edit('workflow.toml', ...)`

lib/crates/fabro-server/src/server/handler/playground.rs
```diff
@@ -0,0 +1,503 @@
...
+        workflow.name.clone()
+    };
+    let goal = if workflow.goal.is_empty() {
+        "(not set yet)".to_string()
+    } else {
+        workflow.goal.clone()
+    };
+
+    format!(
+        "You are Ask Fabro, helping the user build a Fabro workflow inside \
```
brynary: Would externalize into a `*.md` file.

lib/crates/fabro-server/src/server/handler/playground.rs
```diff
@@ -0,0 +1,503 @@
...
+}
+
+fn convert_api_message(msg: &CompletionMessage) -> LlmMessage {
+    let role = match msg.role {
+        CompletionMessageRole::System => Role::System,
+        CompletionMessageRole::User => Role::User,
+        CompletionMessageRole::Assistant => Role::Assistant,
+        CompletionMessageRole::Tool => Role::Tool,
+        CompletionMessageRole::Developer => Role::Developer,
+    };
```
brynary: This completion mapping is usually a code smell for type sprawl -- where multiple similar types represent the same thing.

lib/crates/fabro-server/src/server/handler/playground.rs
```diff
@@ -0,0 +1,503 @@
...
+            error!(error = ?e, "playground: LLM stream call failed");
+            return ApiError::new(StatusCode::BAD_GATEWAY, format!("LLM error: {e}"))
+                .into_response();
+        }
+    };
+
+    // Forward StreamEvents as `stream_event` SSE frames. The browser-side
+    // adapter listens for the `tool_call_end` event carrying the
+    // `write_workflow_file` arguments, parses the DOT, diffs it against
+    // its current draft, and animates the diff into the canvas.
```
brynary, line 169: Just reading this comment:

`// Forward StreamEvents as `stream_event` SSE frames.`

Is that logic we may have elsewhere already?

lib/crates/fabro-server/src/server/handler/playground.rs
```diff
@@ -0,0 +1,503 @@
...
+    Ok(())
+}
+
+pub(super) fn routes() -> Router<Arc<AppState>> {
+    Router::new().route("/playground/chat", post(create_playground_chat))
+}
+
+/// Renders the workflow draft as a compact DOT-style summary the model
+/// can read for current state. Same shape as the file the model writes
+/// back via `write_workflow_file`, minus theming/comments.
```
brynary: This summarization concept is interesting to me. Is using the graph as-is too vebrose?

Issue #305, opened 2026-05-18
Feature request: AWS Bedrock and vendor API token support


depopry:
## Summary

Request to add two related provider/auth capabilities to Fabro's multi-model routing layer:

1. **AWS Bedrock as a model provider**
2. **Vendor API token authentication** (direct API keys for Anthropic, OpenAI, etc.)

---

## Motivation

Fabro's CSS-like model stylesheet is a great abstraction for routing nodes to different models and providers. Right now it appears provider support is tied to specific integrations. Expanding this to include AWS Bedrock and direct vendor tokens would significantly broaden where Fabro can run.

### AWS Bedrock

Many engineering teams — particularly in enterprise and regulated environments — access frontier models exclusively through AWS Bedrock rather than directly via vendor APIs. This is often mandated for compliance, data residency, or cost consolidation reasons. Supporting Bedrock as a provider target in the model stylesheet would allow these teams to adopt Fabro without routing traffic outside their AWS environment.

Example stylesheet syntax (illustrative):

```dot
graph [
    model_stylesheet="
        *       { provider: bedrock; model: anthropic.claude-sonnet-4-5; region: eu-west-1; }
        .heavy  { provider: bedrock; model: anthropic.claude-opus-4-7; }
    "
]
```

### Vendor API tokens

Direct API key authentication (e.g. `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`) would support teams that don't use AWS Bedrock and want to supply their own keys — useful for self-hosted deployments, cost attribution, or teams with existing vendor contracts. This also avoids requiring any cloud provider intermediary for simple local or CI setups.

---

## Expected behaviour

- Model stylesheets can target a `bedrock` provider, with region and credential configuration (IAM role, env vars, or explicit credentials)
- Stylesheets can alternatively accept a vendor API token or reference an env var for direct provider auth
- Fallback chains in the stylesheet work across both auth modes

---

## Context

- Using Fabro for multi-model workflow orchestration across a platform engineering team
- Current model routing is a key differentiator — extending provider support would make it practical in more deployment environments
- Happy to discuss implementation approach or test early builds

bokmann:
A useful update here: current Fabro already has a workable path for most of this via custom providers plus the `openai_compatible` adapter, so native Bedrock support is not the only viable route.

In practice, this means users can already do something like:

```toml
[llm.providers.bedrock]
adapter = "openai_compatible"
agent_profile = "openai"
[...]

PR #459, opened 2026-05-30
feat(llm): Amazon Bedrock provider — Converse codec, SigV4 + API-key auth

depopry:
Adds **Amazon Bedrock** as an opt-in built-in provider, over Bedrock's unified **Converse / ConverseStream** API. One codec serves every Converse-capable family — Claude, Amazon Nova, Meta Llama, Mistral, DeepSeek, Moonshot Kimi, Z.AI GLM, MiniMax, NVIDIA Nemotron, and OpenAI gpt-oss — because AWS translates the envelope to each model's native dialect server-side. Auth is either **AWS SigV4** (the default credential chain — env / profile / IMDS / IRSA / SSO, resolved per request so sessions refresh) or a **Bedrock API key** (`AWS_BEARER_TOKEN_BEDROCK`, bearer). Disabled by default (the Ollama / OpenRouter opt-in pattern).

This is the redo of #459's original Claude-only `InvokeModel` adapter, rebuilt on the gateway-refactor seams (#481–#497). @depopry's SigV4 signer, AWS event-stream frame decoder, `BedrockAuth`, the `aws_sigv4` credential grammar, `AdapterKind::Bedrock`, region-from-base_url, and the lean-deps decision are preserved and authored by him on the first two commits; the per-family `BedrockCodec` trait he wrote turned out to be the crate-wide `Codec` seam in miniature, so the refactor promoted exactly that shape. The original Claude-only description is preserved in a comment below.

## What's here

- **`AdapterKind::Bedrock` × `CodecKind::BedrockConverse`** on the route, plus the `aws_sigv4` credential source (no static secret — the adapter signs at request time; `fabro-auth` stays AWS-free). *(@depopry)*
- **SigV4 signer + AWS event-stream `FrameDecoder`** on the lean AWS stack (no `aws-sdk-bedrockruntime`; transport stays on `fabro-http`). Re-targeted at Converse's direct-JSON stream frames; the signer resolves credentials per request. *(@depopry)*
- **`bedrock_converse` codec** — Converse envelope (`system[]`, typed content blocks, `inferenceConfig`, `toolConfig`), prompt caching via `cachePoint`, thinking-signature round-trip through `reasoningContent`, usage mapped onto the disjoint `TokenCounts` buckets, `provider_options.bedrock` passthrough. Plus the adapter shell and an event-stream byte loop beside the transport's shared SSE loop.
- **Catalog**: `bedrock.toml` (Claude incl. Fable 5, Nova 2, Llama 4, Mistral, DeepSeek, Kimi, GLM, MiniMax, Nemotron, gpt-oss — cross-region inference-profile ids, per-model `billing_policy` so Claude bills Anthropic-style) and a companion **`bedrock-openai`** provider for GPT-5.5/5.4 over the `bedrock-mantle` Responses endpoint (pure config over the existing `openai_responses` codec, zero new code).
- Secrets registry (`AWS_BEARER_TOKEN_BEDROCK`), gitleaks rules for both Bedrock key formats, the `docs/integrations/bedrock` guide, and live e2e tests.

## Live verification (confirmed end-to-end against a real AWS account)

Verified on a real Bedrock account (us-east-2, SigV4 + bearer):

- **SigV4 + Converse** — multiple families (Claude, Nova, DeepSeek, …) via the full settings → catalog → route → adapter → codec path.
- **ConverseStream** — streaming deltas through the workflow engine.
- **Multi-turn tool use** — agent loop with tool calls round-tripping (no-arg tools included).
- **Multi-model routing** — Claude + DeepSeek pinned in one run through the single Converse codec.
- **mantle Responses** — `openai.gpt-5.5` answered via the `bedrock-openai` provider (bearer auth).

The exercise caught and fixed several issues that unit tests (static creds, mocked transports) could not — see the follow-up commits below.

## Follow-up fixes from live testing (commits on top of the foundation)

1. **Worker AWS env** — the workflow worker scrubs its env to an allowlist, so SigV4 (which re-resolves from the ambient chain per request) couldn't work through `fabro run`. The AWS credential-chain inputs now cross into the worker.
2. **Vault bearer key** — Bedrock was the only key-based provider missing a `vault:` credential ref, so `fabro secret set AWS_BEARER_TOKEN_BEDROCK` silently didn't feed it. Now resolves env → vault → SigV4.
3. **Converse tool-encoding hardening** — a no-arg tool call's `toolUse.input` is now a `{}` object (Bedrock rejects null), and every tool `inputSchema` gets a top-level `type: "object"` (strict families like DeepSeek reject a typeless schema Claude tolerates).
4. **Nova output cap** — `amazon.nova-2-lite` max_output 65536 → 65535 (Bedrock's per-request limit).

Earlier fixes already folded into the foundation commits: the `aws-config` sleep-impl (default chain panicked) and AWS error-body decoding (top-level `message`/`Message`/`__type` → proper messages instead of "Unknown error").

## Manual testing & setup

See `docs/integrations/bedrock` — now documents the non-obvious account setup that live testing surfaced: the per-Region Anthropic use-case approval, `aws-marketplace:Subscribe` for third-party models, the Fable 5 / Mythos-class data-sharing opt-in, and the bearer-vs-SigV4 precedence override for running Converse + mantle side by side.

## Open decision / discussion

- **Model-id naming** — Bedrock rows use dotted ids mirroring Bedrock's native inference-profile ids (`us.anthropic.claude-sonnet-4-6`, `openai.gpt-5.5`), which also makes them the wire `api_id`. Third scheme alongside bare ids and OpenRouter's `vendor/model` slashes. No collision risk (enforced at catalog build). Open to a uniform scheme if preferred.
- **`BEDROCK_API_KEY` alias** — see the comment thread; the AWS console hands some users `export BEDROCK_API_KEY=` while the SDK-standard var is `AWS_BEARER_TOKEN_BEDROCK`. Question of whether to accept both.

## Deferred (named follow-ups)

- **`qwen.qwen3-coder-next`** — omitted pending a verified Bedrock model/inference-profile id (its fabro id isn't a valid Bedrock identifier; needs an explicit `api_id`). Re-add once confirmed via `aws bedrock list-inference-profiles`.
- **Claude Mythos 5** — Anthropic-Messages-only on `bedrock-mantle` (limited preview).
- **Converse structured output** (`response_format` rejected with a clear error).
- **`reasoning_effort` on Converse rows** via `additionalModelRequestFields` (the `bedrock-openai` GPT rows already accept effort levels).
- **CountTokens** route (`count_input_tokens` returns `None`).

## Verification

`cargo nextest run --workspace`: green except the pre-existing environment-dependent fabro-workflow failures (identical on main). clippy `-D warnings` + pinned-nightly fmt clean. Codec unit tests + adapter httpmock tests + frame-decoder/signer locks.

### DIFF
--- a/docs/public/reference/user-configuration.mdx
+++ b/docs/public/reference/user-configuration.mdx
@@ -198,7 +198,7 @@ x-team-secret = { vault = "gateway_team_secret" }
 | `billing_policy` | `"openai"` \| `"anthropic"` \| `"gemini"` \| `"none"` | derived from `adapter` | Provider-owned billing algorithm for usage estimates. Override for exceptional providers such as local no-billing runtimes. |
 | `base_url` | string | built-in value or adapter runtime default | Provider API base URL. Required for most custom OpenAI-compatible providers. |
 | `auth` | table | omitted | API-key auth config. Omit the table entirely for providers that need no API key; any `extra_headers` are still attached. |
-| `auth.credentials` | array<string> | required when `auth` present | Ordered credential refs. Accepted forms are `vault:<NAME>` and `env:<NAME>`. Literal secret strings are rejected. |
+| `auth.credentials` | array<string> | required when `auth` present | Ordered credential refs. Accepted forms are `vault:<NAME>`, `env:<NAME>`, and `aws_sigv4` (sign requests from the AWS default credential chain — Bedrock). Literal secret strings are rejected. |
 | `auth.header` | `"bearer"` or `{ custom = "Header-Name" }` | `"bearer"` | Primary API-key header policy. Omit when the provider uses a standard bearer token. |
 | `extra_headers` | table | `{}` | Additional headers attached to provider requests. Values must be typed refs: `{ literal = "..." }`, `{ env = "NAME" }`, or `{ vault = "NAME" }`. |
 | `priority` | integer | `0` | Higher-priority configured providers win default selection; ties use canonical provider ID. |
--- a/lib/crates/fabro-auth/src/resolve.rs
+++ b/lib/crates/fabro-auth/src/resolve.rs
@@ -406,17 +406,18 @@ impl CredentialResolver {
                     .ok_or_else(|| ResolveError::NotConfigured(provider_id.clone()))?;
                 let auth_header = auth_header_for_catalog_provider(provider, key.clone())?;
                 let mut cred = ApiCredential {
-                    provider:      provider_id.clone(),
-                    auth_header:   Some(auth_header),
-                    extra_headers: HashMap::new(),
-                    base_url:      None,
-                    codex_mode:    false,
-                    org_id:        None,
-                    project_id:    None,
+                    provider: provider_id.clone(),
+                    auth_header: Some(auth_header),
+                    extra_headers: self.resolved_extra_headers_for_catalog(
+                        vault,
+                        provider_id,
+                        catalog,
+                    )?,
+                    base_url,
+                    codex_mode: false,
+                    org_id: None,
+                    project_id: None,
                 };
-                cred.base_url = base_url;
-                cred.extra_headers =
-                    self.resolved_extra_headers_for_catalog(vault, provider_id, catalog)?;
                 if provider_id == &ProviderId::openai() {
                     apply_openai_api_env_context(&mut cred, &*self.env_lookup);
                 }
--- a/lib/crates/fabro-auth/src/resolve.rs
+++ b/lib/crates/fabro-auth/src/resolve.rs
@@ -610,6 +610,7 @@ mod tests {
             r#"
 [providers.bedrock]
 adapter = "bedrock"
+enabled = true
 base_url = "https://bedrock-runtime.eu-west-1.amazonaws.com"
 
 [providers.bedrock.auth]
--- a/lib/crates/fabro-auth/src/resolve.rs
+++ b/lib/crates/fabro-auth/src/resolve.rs
@@ -28,6 +28,9 @@ pub(crate) enum ResolvedSecret {
         credential: Box<OAuthCredential>,
         vault_name: String,
     },
+    /// Opaque AWS SigV4 source: no static secret; the adapter signs requests
+    /// using the AWS default credential chain.
+    AwsSigv4,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -333,6 +336,9 @@ impl CredentialResolver {
                 Err(err) => Err(vault_lookup_error(provider, name, err)),
             },
             CredentialRef::Env(name) => Ok((self.env_lookup)(name).map(ResolvedSecret::ApiKey)),
+            // AWS SigV4 is an opaque source: it always "resolves" (the adapter
+            // signs at request time from the AWS chain), no vault/env lookup.
+            CredentialRef::AwsSigv4 => Ok(Some(ResolvedSecret::AwsSigv4)),
         }
     }
 
@@ -379,6 +385,21 @@ impl CredentialResolver {
     ) -> Result<ApiCredential, ResolveError> {
         let base_url = Self::provider_base_url_for_catalog(provider_id, catalog);
         match secret {
+            // Opaque AWS SigV4 source: carry the marker so the adapter signs
+            // with the AWS chain; no static secret resolved here.
+            ResolvedSecret::AwsSigv4 => Ok(ApiCredential {
+                provider: provider_id.clone(),
+                auth_header: Some(ApiKeyHeader::AwsSigv4),
+                extra_headers: self.resolved_extra_headers_for_catalog(
+                    vault,
+                    provider_id,
+                    catalog,
+                )?,
+                base_url,
+                codex_mode: false,
+                org_id: None,
+                project_id: None,
+            }),
             ResolvedSecret::ApiKey(key) => {
                 let provider = catalog
                     .provider(provider_id)
@@ -577,6 +598,37 @@ mod tests {
             api.base_url.as_deref(),
             Some("https://chatgpt.com/backend-api/codex")
         );
+    }
+
+    #[tokio::test]
+    async fn sigv4_provider_resolves_to_aws_sigv4_credential() {
+        let dir = tempfile::tempdir().unwrap();
+        let vault = Vault::load(dir.path().join("secrets.json")).unwrap();
+        // No env credentials configured: SigV4 must still resolve.
+        let resolver = test_resolver(vault, Arc::new(|_| None));
+        let catalog = catalog_with(
+            r#"
+[providers.bedrock]
+adapter = "bedrock"
+base_url = "https://bedrock-runtime.eu-west-1.amazonaws.com"
+
+[providers.bedrock.auth]
+credentials = ["aws_sigv4"]
+"#,
+        );
+
+        let resolved = resolver
+            .resolve(
+                ProviderId::from("bedrock"),
+                CredentialUsage::ApiRequest,
+                &catalog,
+            )
+            .await
+            .unwrap();
+
+        let ResolvedCredential::Api(api) = resolved;
+        assert_eq!(api.provider, ProviderId::from("bedrock"));
+        assert_eq!(api.auth_header, Some(ApiKeyHeader::AwsSigv4));
     }
 
     #[tokio::test]
--- a/lib/crates/fabro-llm/src/adapter_registry.rs
+++ b/lib/crates/fabro-llm/src/adapter_registry.rs
@@ -207,23 +207,6 @@ fn build_openai_compatible(config: AdapterConfig) -> Result<Arc<dyn ProviderAdap
     Ok(Arc::new(build_openai_compatible_adapter(config)?))
 }
 
-/// Placeholder until the Bedrock adapter lands later in this series; the
-/// adapter kind exists first so the auth/config grammar can be wired and
-/// tested ahead of it.
-#[expect(
-    clippy::needless_pass_by_value,
-    reason = "Adapter factories share the by-value AdapterFactory signature."
-)]
-fn build_bedrock(config: AdapterConfig) -> Result<Arc<dyn ProviderAdapter>, Error> {
-    Err(Error::Configuration {
-        message: format!(
-            "provider '{}': the bedrock adapter is not yet wired",
-            config.provider_id
-        ),
-        source:  None,
-    })
-}
-
 /// Return the factory for a known adapter kind.
 #[must_use]
 pub fn factory_for(adapter_kind: AdapterKind) -> AdapterFactory {
@@ -232,7 +215,7 @@ pub fn factory_for(adapter_kind: AdapterKind) -> AdapterFactory {
         AdapterKind::OpenAi => build_openai,
         AdapterKind::Gemini => build_gemini,
         AdapterKind::OpenAiCompatible => build_openai_compatible,
-        AdapterKind::Bedrock => build_bedrock,
+        AdapterKind::Bedrock => providers::bedrock::build,
     }
 }
 
--- a/lib/crates/fabro-llm/src/adapter_registry.rs
+++ b/lib/crates/fabro-llm/src/adapter_registry.rs
@@ -90,7 +90,9 @@ fn apply_primary_auth_header(
             extra_headers.insert(name, value);
             None
         }
-        None => None,
+        // SigV4 is not a static header; only the Bedrock adapter consumes
+        // the marker (it signs at request time).
+        Some(ApiKeyHeader::AwsSigv4) | None => None,
     }
 }
 
@@ -205,6 +207,23 @@ fn build_openai_compatible(config: AdapterConfig) -> Result<Arc<dyn ProviderAdap
     Ok(Arc::new(build_openai_compatible_adapter(config)?))
 }
 
+/// Placeholder until the Bedrock adapter lands later in this series; the
+/// adapter kind exists first so the auth/config grammar can be wired and
+/// tested ahead of it.
+#[expect(
+    clippy::needless_pass_by_value,
+    reason = "Adapter factories share the by-value AdapterFactory signature."
+)]
+fn build_bedrock(config: AdapterConfig) -> Result<Arc<dyn ProviderAdapter>, Error> {
+    Err(Error::Configuration {
+        message: format!(
+            "provider '{}': the bedrock adapter is not yet wired",
+            config.provider_id
+        ),
+        source:  None,
+    })
+}
+
 /// Return the factory for a known adapter kind.
 #[must_use]
 pub fn factory_for(adapter_kind: AdapterKind) -> AdapterFactory {
@@ -213,6 +232,7 @@ pub fn factory_for(adapter_kind: AdapterKind) -> AdapterFactory {
         AdapterKind::OpenAi => build_openai,
         AdapterKind::Gemini => build_gemini,
         AdapterKind::OpenAiCompatible => build_openai_compatible,
+        AdapterKind::Bedrock => build_bedrock,
     }
 }
 
--- a/lib/crates/fabro-llm/src/providers/bedrock/mod.rs
+++ b/lib/crates/fabro-llm/src/providers/bedrock/mod.rs
@@ -9,7 +9,7 @@
 pub(crate) mod eventstream;
 pub(crate) mod sigv4;
 
-use std::collections::VecDeque;
+use std::collections::{HashMap, VecDeque};
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -22,6 +22,8 @@ use tokio::sync::OnceCell;
 use tokio::time;
 
 use crate::adapter_registry::AdapterConfig;
+#[cfg(test)]
+use crate::adapter_registry::AdapterKindOptions;
 use crate::attachments::{self, AttachmentPolicy};
 use crate::codec::bedrock_converse::BedrockConverse;
 use crate::codec::{Codec, CodecCtx, CodecParams, EncodedRequest, RawEvent, StreamDecoder};
@@ -61,8 +63,16 @@ pub(crate) fn build(config: AdapterConfig) -> Result<Arc<dyn ProviderAdapter>, E
         })?;
     let adapter = match config.auth_header {
         Some(ApiKeyHeader::AwsSigv4) => Adapter::new_sigv4(base_url)?,
-        Some(ApiKeyHeader::Bearer(token) | ApiKeyHeader::Custom { value: token, .. }) => {
-            Adapter::new_api_key(token, base_url)?
+        Some(ApiKeyHeader::Bearer(token)) => Adapter::new_api_key(token, base_url)?,
+        Some(ApiKeyHeader::Custom { name, .. }) => {
+            return Err(Error::Configuration {
+                message: format!(
+                    "bedrock provider '{}' does not support custom auth header '{}' (use bearer \
+                     credentials or aws_sigv4)",
+                    config.provider_id, name
+                ),
+                source:  None,
+            });
         }
         None => {
             return Err(Error::Configuration {
@@ -76,6 +86,9 @@ pub(crate) fn build(config: AdapterConfig) -> Result<Arc<dyn ProviderAdapter>, E
         }
     };
     let mut adapter = adapter.with_name(config.provider_id);
+    if !config.extra_headers.is_empty() {
+        adapter = adapter.with_default_headers(config.extra_headers);
+    }
     if let Some(catalog) = config.catalog {
         adapter = adapter.with_catalog(catalog);
     }
@@ -133,6 +146,12 @@ impl Adapter {
     }
 
     #[must_use]
+    pub fn with_default_headers(mut self, headers: HashMap<String, String>) -> Self {
+        self.http = self.http.with_default_headers(headers);
+        self
+    }
+
+    #[must_use]
     pub fn with_timeout(self, timeout: AdapterTimeout) -> Self {
         Self {
             http: self.http.with_timeout(timeout),
@@ -182,6 +201,9 @@ impl Adapter {
         for (key, value) in &self.http.default_headers {
             req = req.header(key, value);
         }
+        for (key, value) in &encoded.headers {
+            req = req.header(key, value);
+        }
 
         req = match &self.auth {
             BedrockAuth::ApiKey(token) => req.bearer_auth(token).body(body),
@@ -189,7 +211,7 @@ impl Adapter {
                 let signer = cell
                     .get_or_try_init(Sigv4Signer::from_default_chain)
                     .await?;
-                signer.sign_post(req, &self.region, &url, &body).await?
+                signer.sign_post(req, &self.region, &url, body).await?
             }
         };
 
@@ -378,11 +400,12 @@ fn region_from_base_url(base_url: &str) -> Result<String, Error> {
         ),
         source:  None,
     };
-    let host = base_url
-        .strip_prefix("https://")
-        .or_else(|| base_url.strip_prefix("http://"))
-        .unwrap_or(base_url);
-    let host = host.split('/').next().unwrap_or(host);
+    #[expect(
+        clippy::disallowed_types,
+        reason = "Bedrock region derivation needs URL host parsing; the raw URL is not logged or rendered."
+    )]
+    let parsed = fabro_http::Url::parse(base_url).map_err(|_| invalid())?;
+    let host = parsed.host_str().ok_or_else(invalid)?;
     let rest = host
         .strip_prefix("bedrock-runtime-fips.")
         .or_else(|| host.strip_prefix("bedrock-runtime."))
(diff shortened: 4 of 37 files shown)


lib/crates/fabro-model/src/catalog.rs
```diff
@@ -1256,6 +1273,14 @@ fn build_providers(
         let auth = settings.auth.clone();
         validate_provider_auth(&provider_id, auth.as_ref())?;
 
+        // Bedrock is region-scoped: the region derives the base URL and the
+        // SigV4 signing region, so it must be configured.
+        if adapter == AdapterKind::Bedrock && settings.region.is_none() {
+            return Err(CatalogBuildError::BedrockMissingRegion {
+                provider: provider_id,
+            });
+        }
```
brynary: This code is leaking adapter-specific behavior into the generic `catalog.rs`. We may want to add a concept of validation to the adapter trait so that an adapter has the opportunity to return an `Err` if it is not in a valid configuration.

lib/crates/fabro-model/src/catalog.rs
```diff
@@ -477,6 +489,8 @@ pub struct CatalogProvider {
     pub billing_policy: BillingPolicy,
     pub api_key_url:    Option<String>,
     pub base_url:       Option<String>,
+    /// AWS region for region-scoped providers (Bedrock). `None` for others.
+    pub region:         Option<String>,
```
brynary: I'm weary of adding AWS-specific `region` to the generic `CatalogProvider `. One possibility would be adding a generic way for key/value pairs to be configured at the catalog level.

Alternatively, can the region be encoded into the `base_url`? If so that would remove the need for an extra field enitrely.

lib/crates/fabro-llm/src/providers/bedrock/mod.rs
```diff
@@ -0,0 +1,580 @@
...
+/// How the adapter authenticates to Bedrock.
+pub(crate) enum BedrockAuth {
+    /// Long-lived Bedrock API key, sent as an `Authorization: Bearer` token.
+    ApiKey(String),
+    /// SigV4 with an already-resolved signer.
+    Sigv4(Sigv4Signer),
+    /// SigV4 where credentials are resolved from the AWS default chain on first
+    /// use and cached. Lets the synchronous adapter factory defer the async
+    /// chain resolution to request time.
+    Sigv4Lazy(OnceCell<Sigv4Signer>),
```
brynary: Do we need both the eager and lazy variants?

lib/crates/fabro-llm/src/providers/anthropic.rs
```diff
@@ -231,7 +231,7 @@ struct CountTokensResponse {
     input_tokens: i64,
 }
 
-fn token_counts_from_api_usage(usage: &ApiUsage) -> TokenCounts {
+pub(crate) fn token_counts_from_api_usage(usage: &ApiUsage) -> TokenCounts {
```
brynary: If I am reading it properly, this PR takes a number of Anthropic-related functions and moves them from private to crate-public so the logic can be shared with Bedrock.

There are two concerns I have with this:

- First, from a pure code organization standpoint, I am wondering if we can better encapsulate this into an abstraction (e.g. an Anthropic "codec").

- Second, this Bedrock supports many [models](https://docs.aws.amazon.com/bedrock/latest/userguide/model-cards.html) including many non-Anthropic models. When someone uses a non-Anthropic model via Bedrock, will this break?

lib/crates/fabro-model/src/catalog.rs
```diff
@@ -217,9 +221,17 @@ pub enum CredentialRefParseError {
...
+    /// schemes that do not use a static credential (e.g. Bedrock `aws_sigv4`,
+    /// which signs with the AWS default credential chain).
+    #[serde(default)]
     pub credentials: Vec<CredentialRef>,
     #[serde(default)]
     pub header:      ApiKeyHeaderPolicy,
+    /// Authentication scheme for adapters that support more than one (Bedrock).
+    /// `None` means the adapter's default.
+    #[serde(default)]
+    pub scheme:      Option<AuthScheme>,
```
brynary: My initial reaction here is that `scheme` should be folded into the pre-existing `credentials` by expanding (and renaming) `CredentialRef`, probably to a `CredentialSource`. In AWS, in the Sigv4 system, the credentials are coming from _somewhere_, it just may be opaque.

The goal would be to make it so that invalid states are not representable in the type system (within reason).

lib/crates/fabro-llm/src/adapter_registry.rs
```diff
@@ -193,9 +199,43 @@ pub fn factory_for(adapter_kind: AdapterKind) -> AdapterFactory {
         AdapterKind::OpenAi => build_openai,
         AdapterKind::Gemini => build_gemini,
         AdapterKind::OpenAiCompatible => build_openai_compatible,
+        AdapterKind::Bedrock => build_bedrock,
     }
 }
 
+fn build_bedrock(config: AdapterConfig) -> Result<Arc<dyn ProviderAdapter>, Error> {
```
brynary: Leaking Bedrock-specific code into files outside of the `bedrock/` folder seems like a code smell.

lib/crates/fabro-auth/src/resolve.rs
```diff
@@ -209,6 +211,25 @@ impl CredentialResolver {
...
+            return Ok(ResolvedCredential::Api(ApiCredential {
+                provider: provider_id.clone(),
+                auth_header: None,
+                extra_headers,
+                base_url: Self::provider_base_url_for_catalog(&provider_id, catalog),
+                codex_mode: false,
+                org_id: None,
+                project_id: None,
+            }));
+        }
```
brynary: We're leaking Bedrock-specific code here outside of the `bedrock/` folder, it seems like we should attempt to iterate the adapter interface to allow for a "tell, don't ask"-based interface.

lib/crates/fabro-auth/src/resolve.rs
```diff
@@ -303,6 +324,11 @@ impl CredentialResolver {
                 .resolved_extra_headers_for_catalog(vault, &provider.id, catalog)
                 .is_ok();
         };
+        // SigV4 providers rely on the AWS credential chain, not a static secret,
+        // so treat them as configured whenever the provider is declared.
+        if auth.scheme == Some(AuthScheme::AwsSigv4) {
+            return true;
+        }
```
brynary: Same note about leaking here.

brynary (commented):
Hi @depopry -- Thank you for this PR... It's a good start.

I reviewed the whole PR (by hand, no robot!) and left comments. Ultimately the tricky part here is going to be refactoring a bit to make it so Bedrock logic is well encapsulated and the type system is clean.

Please let me know what you think of the comments.

lib/crates/fabro-llm/src/providers/anthropic.rs
depopry: Very fair point, and my bias towards anthropic appears transparent here 😅 I'll work on making this more model agnostic

lib/crates/fabro-model/src/catalog.rs
depopry: Removed. Region moved into `base_url` (the adapter parses the signing region from the Bedrock runtime endpoint), so the catalog no longer special-cases bedrock region validation.

lib/crates/fabro-model/src/catalog.rs
depopry: Dropped the `region` field. The endpoint already encodes the region (`bedrock-runtime.<region>.amazonaws.com`), so the adapter parses it from `base_url` and no AWS-specific catalog field is needed.

lib/crates/fabro-llm/src/providers/bedrock/mod.rs
depopry: Collapsed to a single `Sigv4(OnceCell<Sigv4Signer>)` variant: lazy chain resolution in production, tests pre-seed the cell. Eager variant removed.

lib/crates/fabro-llm/src/providers/anthropic.rs
depopry: Added a `BedrockCodec` trait (request body / response parse / streaming decode) selected by model family; `AnthropicCodec` encapsulates the Anthropic wire reuse. Non-Anthropic Bedrock families now return a clear error instead of emitting a malformed Anthropic payload.

lib/crates/fabro-model/src/catalog.rs
depopry: Folded the scheme into the credential list: added `CredentialRef::AwsSigv4` (the opaque AWS-chain source) and removed `auth.scheme`/`AuthScheme`. The mode is now implied by which credential resolves, so invalid combinations are unrepresentable. Note: I kept the name `CredentialRef` because `CredentialSource` is already a trait in fabro-auth — happy to rename to e.g. `CredentialOrigin` if you'd prefer.

lib/crates/fabro-llm/src/adapter_registry.rs
depopry: Moved construction into `providers::bedrock::build`; the registry now just dispatches `AdapterKind::Bedrock => providers::bedrock::build`.

lib/crates/fabro-auth/src/resolve.rs
depopry: Removed. With `CredentialRef::AwsSigv4`, resolution is generic: the source resolves to an `ApiKeyHeader::AwsSigv4` marker like any other credential, so there's no bedrock/scheme special-casing in fabro-auth.

lib/crates/fabro-auth/src/resolve.rs
depopry: Removed the special-case — an `aws_sigv4` credential resolves like any other, so `has_credential_material` no longer needs a bedrock-specific branch.

depopry:
Pushed a refactor pass addressing the review (commits on top of the original set):

- **Encapsulation / "tell, don't ask"**: the `scheme` vs `credentials` split is gone — `CredentialRef` gains an `AwsSigv4` variant (the opaque AWS-chain source), so the auth mode is implied by which credential resolves and invalid combinations are unrepresentable. This removed the bedrock/sigv4 special-casing from both `fabro-auth` resolution and catalog validation; SigV4 resolves generically to an `ApiKeyHeader::AwsSigv4` marker.
- **No AWS-specific `region` on the generic type**: dropped the field; the adapter parses the signing region from `base_url` (the Bedrock runtime endpoint).
- **Bedrock code out of generic files**: construction moved to `providers::bedrock::build`.
- **Model-agnostic**: added a per-family `BedrockCodec` trait; `AnthropicCodec` encapsulates the Anthropic wire reuse, and non-Anthropic Bedrock models now return a clear error rather than a malformed request.
- **SigV4 variants**: collapsed eager/lazy into one `OnceCell`-backed variant.

One naming note left inline: I kept `CredentialRef` rather than `CredentialSource` since the latter is already a trait in `fabro-auth` — glad to pick another name if you'd like.

A couple of items still need your environment's authoritative tooling (I couldn't run them here): `cargo +nightly-2026-04-14 fmt --all` and `cargo nextest run --workspace`.

brynary:
@depopry -- Fantastic, thanks. I really appreciate your thoughtful processing of the code review feedback.

I'll do a follow-up pass on this today!

depopry:
@brynary - Thanks for the feedback so far!

As to try and contain the scope, I've currently only enabled anthropic support for Bedrock, but the pattern should now be in place to expand this to other models in the future

brynary:
@depopry thanks for the updates on this.

I chatted this morning with @swerner who has #438 open with OpenRouter support. OpenRouter has some similarities to Bedrock, as a single provider offering multiple model families. I think if we can nail OpenRouter and Bedrock, then we will end up with a nice, extensible architecture that can support other systems like Google Vertex.

What I'd like to do here is a bit of [preparatory refactoring](https://x.com/kentbeck/status/250733358307500032) where we make the change easy to make before making the change. In this case, I'd like to make the abstraction of a "Codec" that speaks the HTTP requests and responses a first-class concept and move all the existing Adapters to be Codec-based. This is going to churn a lot of the provider-related code, but it should help us increase confidence in our abstractions.

This may add a couple of days before we can merge Bedrock but in return I think we should have a more durable and elegant implementation which is lees likely to have bugs and easier to evolve. Will a few days cause you an issue on your side?

depopry:
Hey @brynary thanks for the update, no concerns my side!

swerner:
Hey @depopry thanks again for this PR! We just completed a big refactor over the last week or so, starting with PR #481 and ending with #497 adding OpenRouter, and are now ready to pull in support for Bedrock.

We've restructured the branch a bit so it conforms to the new pattern for providers and model codecs, and we were able to build off of a lot of your work like the SigV4 and eventstream additions. Those should mostly be in there, and we've also built out the whole Codec pattern crate-wide now.

I'm going to be doing another manual pass to verify everything works and then amend this PR - though you'll still be listed as author/co-author on the commits. Your original branch head will be preserved at `35b2f86f2` if you want to diff.

docs/public/reference/user-configuration.mdx
```diff
@@ -193,18 +193,84 @@ x-team-secret = { vault = "gateway_team_secret" }
...
 | `enabled` | boolean | `true` | Set `false` to disable a provider after lower-precedence layers define it. |
 | `aliases` | array<string> | `[]` | Additional provider names accepted by model routing and fallback config. |
 
+### Amazon Bedrock
+
+The `bedrock` adapter calls Bedrock's `InvokeModel` API with native Anthropic
+payloads (Anthropic-family models). Bedrock is not configured by default because
+it needs a regional endpoint, IAM/model-access setup, and model ids specific to
+your account; define the provider and its models in your config. The signing
+region is taken from `base_url`.
```
brynary: Given that [OpenAI](https://aws.amazon.com/bedrock/openai/) now has equal footing on Bedrock, this seems overly Anthropic-specific

docs/public/reference/user-configuration.mdx
```diff
@@ -193,18 +193,84 @@ x-team-secret = { vault = "gateway_team_secret" }
...
+**Long-lived Bedrock API key.** For environments without role-based credentials,
+use a Bedrock API key sent as a bearer token:
+
+```toml title="settings.toml"
+[llm.providers.bedrock]
+adapter = "bedrock"
+base_url = "https://bedrock-runtime.us-east-1.amazonaws.com"
+
+[llm.providers.bedrock.auth]
+credentials = ["env:AWS_BEARER_TOKEN_BEDROCK", "vault:AWS_BEARER_TOKEN_BEDROCK"]
```
brynary: Since Bedrock is an AWS service, does it support a single token variable like this?

Or, does it require an IAM Access Key ID and Secret Access Key pairing?

lib/crates/fabro-llm/src/providers/bedrock/codec.rs
```diff
@@ -0,0 +1,204 @@
...
+
+use crate::error::Error;
+use crate::providers::{anthropic, common};
+use crate::types::{
+    ContentPart, FinishReason, Message, RateLimitInfo, Request, Response, Role, StreamEvent,
+};
+
+/// The model-native wire format for one Bedrock model family.
+#[async_trait::async_trait]
+pub(crate) trait BedrockCodec: Send + Sync {
```
brynary: @swerner this `trait BedrockCodec` caught me off guard. My naive assumption was that Bedrock would re-use existing e.g. Anthropic codecs when accessing Claude models on Anthropic. Can you please clarify?

swerner:
@brynary I'm still working through some manual tests. Amended PR will be up shortly

swerner:
📦 **Preserving the original PR description for history.** The description above has been updated to reflect the expanded multi-model scope (Converse codec covering all hosted families, not just Anthropic), but @depopry's original write-up — and the design it described — is below verbatim. The SigV4 signer, event-stream decoder, and `aws_sigv4` auth grammar from it are kept and authored by him in this branch's first two commits.

---

## Summary

Adds a native `bedrock` adapter to `fabro-llm` so Anthropic-family models on Amazon Bedrock can be used directly, without an OpenAI-compatible translation proxy. Implements the approach agreed in #305.

The adapter speaks Bedrock's `InvokeModel` / `InvokeModelWithResponseStream` API with **native Anthropic Messages payloads** and supports two authentication modes:

- **`aws_sigv4`** (default) — SigV4 request signing via the standard AWS credential chain (environment, IRSA/web-identity, EC2/ECS instance profile, SSO, assume-role). This is the primary deployment target: Fabro on EKS/EC2 with an IAM role granting Bedrock access, never handling a static secret.
- **`bedrock_api_key`** — a long-lived Bedrock API key sent as an `Authorization: Bearer` token, for environments without role-based credentials.

Closes #305 (pending review).

## Design notes

- **Lean AWS dependencies** (`aws-config`, `aws-credential-types`, `aws-sigv4`, `aws-smithy-eventstream`, `aws-smithy-types`, `aws-smithy-runtime-api`). Transport stays on the in-house `fabro-http` client, so signed Bedrock requests still flow through the existing retry / middleware / redaction / `no_proxy` layers. The full `aws-sdk-bedrockruntime` (and its parallel hyper client) is intentionally **not** pulled in.
- **Reuse, no drift.** The Anthropic Messages translation, response parsing, and streaming accumulator are exposed `pub(crate)` from the Anthropic adapter and reused. Bedrock's streaming responses (`application/vnd.amazon.eventstream`) are decoded frame-by-frame into the **same** inner Anthropic event JSON and fed through the shared `StreamAccumulator`, so streaming behavior matches the direct Anthropic adapter.
- **Provider-level region.** `region` derives both the endpoint (`https://bedrock-runtime.{region}.amazonaws.com`) and the SigV4 signing region, and is validated as `[a-z0-9-]`. Multiple regions = multiple provider entries; cross-region inference profiles work via inference-profile model ids.
- **Explicit `auth.scheme`** config (defaults to `aws_sigv4`). SigV4 providers carry no static credential, so they resolve to a credential with no auth header and sign at request time; the credential chain is resolved lazily on first request and cached.
- **No built-in Bedrock provider.** Bedrock needs operator-specific region / IAM / model-access configuration with no universal default, so it is configured in `fabro.toml` (documented under `[llm.providers.<id>]`) rather than shipped enabled in the embedded catalog.

## Security

- No hardcoded secrets; API keys via env/vault credential refs, SigV4 via the AWS chain.
- The adapter logs no headers/credentials, and `Sigv4Signer` / `BedrockAuth` / `Adapter` do not derive `Debug`, so the bearer token, SigV4 `Authorization`, and `x-amz-security-token` cannot leak through logs. `fabro-http` does not log headers; the pattern redactor covers AWS access keys in incidental log strings.
- `region` is validated before being used as a URL host component.

## Testing

- Unit tests for region/URL building, blocking `complete()` and streaming (mocked HTTP, including event-stream frames split across chunks), SigV4 header generation, AWS error → `ProviderErrorKind` mapping, the factory, and config parsing/validation.
- Opt-in live e2e test gated on `AWS_BEDROCK_E2E` (uses real AWS credentials + Bedrock access; never runs in the default profile).
- All touched crates pass: `fabro-llm`, `fabro-model`, `fabro-auth`, `fabro-config`, `fabro-agent`.

> Note for reviewers/CI: this branch was prepared in an environment without the pinned nightly toolchain, so please run `cargo +nightly-2026-04-14 fmt --all` and `cargo nextest run --workspace` as the authoritative format/test gate.

lib/crates/fabro-auth/src/resolve.rs
```diff
@@ -379,6 +385,21 @@ impl CredentialResolver {
...
+            ResolvedSecret::AwsSigv4 => Ok(ApiCredential {
+                provider: provider_id.clone(),
+                auth_header: Some(ApiKeyHeader::AwsSigv4),
+                extra_headers: self.resolved_extra_headers_for_catalog(
+                    vault,
+                    provider_id,
+                    catalog,
+                )?,
+                base_url,
+                codex_mode: false,
```
brynary, line 399: Definitely not introduced by this PR. but the `codex_mode` especially and also the `org_id`/`project_id` fields on this `ApiCredential` are over-fit to OpenAI

brynary (commented):
This looks really good! The Bedrock Converse codec really drove this home.

lib/crates/fabro-auth/src/resolve.rs
swerner: Good point, I can do a followup PR for something on fabro-auth for this - adding probably per-adapter credential variants?

swerner:
**Discussion: should we also accept `BEDROCK_API_KEY` as a credential alias?**

We name the bearer credential `AWS_BEARER_TOKEN_BEDROCK` — the official, SDK-recognized variable ([AWS docs](https://docs.aws.amazon.com/bedrock/latest/userguide/api-keys-use.html): boto3 and the JS SDK auto-detect it, and `Authorization: Bearer $AWS_BEARER_TOKEN_BEDROCK` is the documented cURL usage). That's deliberate: anyone who's already wired Bedrock for the AWS CLI/SDK has it set, and our credential chain picks it up with zero extra config.

But in testing, the AWS console's copy-paste snippet sometimes hands you `export BEDROCK_API_KEY=...` (and the mantle/OpenAI quickstart uses `OPENAI_API_KEY` for the *same* key). Same key string, different env-var label depending on which console page you're on.

Question: do we want to accept `BEDROCK_API_KEY` as an additional alias in both providers' credential lists (so a copy-paste works regardless), keeping `AWS_BEARER_TOKEN_BEDROCK` primary? It's a one-line catalog addition + the static registry entry. Slight cost: a second supported name to document. Leaning toward adding it for ergonomics, but wanted a read before baking in a second public env var.

brynary:
I think we can accept `BEDROCK_API_KEY` as well, that seems harmless

swerner:
Ran `/simplify` over this branch with Claude Opus 4.8 (1M) — a quality-only pass (reuse / simplification / efficiency / altitude), no correctness-bug hunting. Four review agents fanned out over the diff in parallel.

**Net: the new code was already in good shape.** One small readability fix applied:

- **`fabro-auth/src/resolve.rs`** — collapsed a struct-then-mutate in the `ApiKey` credential arm. `ApiCredential` was built with `extra_headers: HashMap::new()` / `base_url: None` and then both were immediately overwritten on the next two lines; inlined them into the constructor to match the `AwsSigv4`/`OAuth` arms right above it. No behavior change; `cargo build -p fabro-auth` clean.

Notable findings we **looked at and deliberately skipped**:

- *Efficiency — flagged clones in `bedrock_converse/encode.rs` and `stream.rs`:* false positives. The codec builds an owned wire body from a borrowed `&Request`, so those copies are required; the streamed `ToolCall` genuinely needs to live in both `self.parts` and the emitted `ToolCallEnd` event (one clone = the minimum).
- *Reuse — `EventStreamLoop`/`decode_eventstream` overlaps the shared SSE transport loop:* real, but intentional and already noted in the code. AWS binary `vnd.amazon.eventstream` framing isn't text SSE; unifying them means extending `SseFraming` with a binary variant and relocating `FrameDecoder` into the transport — a design-level refactor, out of scope for a cleanup pass.
- *Reuse — `merge_provider_options` mirrors the Anthropic codec's version:* genuine duplication, but the signatures differ and real reuse would require also rewriting the Anthropic codec (outside this diff).
- *Altitude — clean pass.* The earlier "Bedrock leaking into generic files" feedback is fully addressed: `AwsSigv4` is a correctly-generalized "non-static request-time auth" seam, dispatch is first-class enum variants, and AWS-specific logic stays confined to `providers/bedrock/` and `codec/bedrock_converse/`.

swerner:
Ran **/simplify** on this branch with GPT-5.5 at xhigh reasoning, using three parallel review agents over the full PR diff:

- **Reuse review** looked for existing helpers/utilities and duplicate logic.
- **Quality review** looked for abstraction leaks, redundant state, and stringly-typed or dropped config paths.
- **Efficiency review** looked for unnecessary copies, hot-path work, and avoidable allocations.

Findings fixed in `e0a4e6bc1`:

- Bedrock adapter now preserves shared adapter config: `extra_headers` are installed, codec headers are applied, and unsupported custom auth headers are rejected instead of silently becoming bearer tokens.
- `aws_sigv4` credentials are now catalog-validated as Bedrock-only, so non-Bedrock providers cannot build an unauthenticated route from that marker.
- Worker env passthrough no longer forwards Bedrock bearer API-key env vars; those remain optional LLM provider secrets and should be read through the vault in server runtime.
- SigV4 request signing now takes ownership of the serialized request body and sends the same `Vec<u8>`, avoiding an extra full-body copy.
- Shared codec helpers now cover adapter-name-keyed `provider_options` merging and malformed tool-argument fallback to `{}`, reusing that behavior across Bedrock/OpenAI-compatible/OpenAI/Anthropic paths.
- Bedrock region parsing now uses URL host parsing rather than manual prefix/split handling.
- Prompt-cache placement avoids allocating an index vector.
- Bedrock MIME format mapping now emits the formats Converse expects instead of splitting arbitrary MIME suffixes.
- Updated the stale catalog comment around Claude Fable 5.

Findings intentionally left as follow-ups:

- Feature-gating the AWS SDK dependency stack would be a larger packaging decision.
- One-at-a-time Bedrock event-stream frame yielding and parallel local attachment reads are real efficiency opportunities, but not worth expanding this cleanup commit.

Validation:

- `cargo +nightly-2026-04-14 clippy -p fabro-llm -p fabro-model -p fabro-server --all-targets -- -D warnings`
- `cargo test -p fabro-llm --lib`
- `cargo test -p fabro-model --lib`
- `cargo test -p fabro-server --lib spawn_env::tests::worker_allowlist_is_fail_closed`

One broader unsandboxed run of `cargo test -p fabro-llm -p fabro-model -p fabro-server --lib` failed only in three unrelated `fabro-server` object-store tests whose behavior depends on host storage/AWS env assumptions; the touched suites above pass.

docs/public/reference/user-configuration.mdx
swerner: It turns out it supports both, and you can configure whichever you like with the new setup.

PR #467, opened 2026-06-02
feat(workflow): support overriding cwd for local sandbox provider

amuraru:
## Problem

The `local` sandbox uses the run's `source_directory` (the CLI's cwd at invocation time) as its working directory and `create_dir_all`s it on the server (`LocalSandbox::initialize` in `fabro-sandbox`). That is correct when the CLI and the server share a host — the agent operates directly on the user's project tree.

When the server is **remote** from the CLI — e.g. `fabro serve` running in a container in Kubernetes, driven over HTTP with the `local` sandbox — the client's cwd (e.g. `/Users/alice/project`) does not exist on the server. The sandbox then tries to create that path as the (often unprivileged) server user and fails at init:

```
sandbox.failed provider="local" error="Failed to create working directory" causes=["Permission denied (os error 13)"]
```

and the run dies with `workflow_error` before the agent starts.

## Fix

When `source_directory` is absent or does not exist on the server, fall back to a server-writable `workspace` directory under the run's scratch dir instead of recreating the client path. **Same-host behavior is unchanged**: an existing `source_directory` is still used as-is.

The selection is extracted into a small pure helper, `local_working_directory(source_directory, run_dir)`, so it can be unit-tested directly.

## Testing

- `cargo test -p fabro-workflow local_working_directory` — 3 new tests (existing source dir → used; absent → fallback; present-but-missing-on-server → fallback)
- `cargo check -p fabro-workflow`

Thanks for fabro @brynary!

### DIFF
--- a/docs/public/api-reference/fabro-api.yaml
+++ b/docs/public/api-reference/fabro-api.yaml
@@ -6371,6 +6371,10 @@ components:
           example: 0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef
         provider:
           $ref: "#/components/schemas/EnvironmentProvider"
+        cwd:
+          type: ["string", "null"]
+          description: Runtime command working directory for this environment.
+          example: /srv/fabro/workspaces/team-a
         image:
           $ref: "#/components/schemas/EnvironmentApiImageSettings"
         resources:
@@ -6406,6 +6410,10 @@ components:
           example: docker
         provider:
           $ref: "#/components/schemas/EnvironmentProvider"
+        cwd:
+          type: ["string", "null"]
+          description: Runtime command working directory for this environment.
+          example: /srv/fabro/workspaces/team-a
         image:
           $ref: "#/components/schemas/EnvironmentApiImageSettings"
         resources:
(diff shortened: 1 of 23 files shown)


brynary:
Hi @amuraru -- Thanks for the kind words and this report!

The `local` sandbox provider was envisioned for use when the Fabro server and Fabro CLI are on the same host -- typically a developer laptop. I agree that this use case of using `local` sandbox with a split client/server host seems like something that we should support cleanly, and am focused on how to do that elegantly.

My concern with this fallback behavior is that it is implicit and could be surprising. I'd prefer to find something that is explicit and clear. Here is what I am thinking...

The recently added `Environment` abstraction encapsulates the idea of a sandbox provider (e.g. `local` or `docker`) combined with any additional settings -- e.g. env vars, CPU limits, etc.

Right now the Environment does not include a `cwd` setting, however that appears to be an omission. The Environment should allow customizing the CWD where commands are run by default.

Therefore, I think the best solution is to add `cwd` support to environment. Once we have that, you will be able to define an environment like this:

```toml
[environments.my-server]
provider = "local"
cwd = "/any/path/on/the/server"
```

Then assign your runs to use that environment from the `project.toml` or `workflow.toml`:

```toml
[run.environment]
id = "my-server"
```

This makes it totally explicit _and_ has the benefit you can have different Fabro runs execute in different server directories, all independent of the CLI-side CWD.

Does this sound like it would work for you?

amuraru:
Hi @brynary!

i like the Environment construct and adding the cwd at that level would solve my usecase as well. Thanks again for opensourcing fabro - I am piloting it for a team shared instance. The local runner works fine for the moment - i am also thinking a more k8s native where workflow runs are using better isolation would also be useful

brynary:
@amuraru update: I am working on this actively now.

PR #471, opened 2026-06-03
test(llm): pin provider wire behavior with per-dialect snapshot tests

swerner:
## What

Adds wire snapshot tests pinning the exact encode/decode/stream behavior of all four provider adapters — **109 tests / 117 insta snapshots** in `fabro-llm/tests/it/wire/{anthropic,openai_compatible,openai_responses,gemini}.rs`, driven by a shared canonical request corpus in `tests/it/support.rs`. Tests only; no `src/` changes.

Each test points a real adapter at a local httpmock server, side-channels the full received request (method, path, headers, body) out of an `is_true` matcher closure, responds with a canned provider body or scripted SSE transcript, and snapshots both the captured wire request and the decoded canonical `Response` / `Vec<StreamEvent>`.

## Why

This is the behavior-pinning net for an upcoming refactor series that separates fabro-llm's wire translation (codec) from transport/auth concerns. The refactor must be behavior-preserving; these snapshots make that checkable per PR instead of asserted. The anthropic and gemini dialects have no twin coverage, so these tests are the only net for those paths.

httpmock matcher-capture is used for all four dialects (rather than twin request-logs for the OpenAI ones): the corpus deliberately exercises shapes a strict twin would reject (provider_options merges, response_format variants, bad-file-path attachment parts), one mechanism is cheaper to maintain than two, and the twin already validates the OpenAI dialects via `parity_matrix` and the server scenario tests.

## Coverage

Per dialect:

- **Encode** — multi-turn/system mapping, `tool_choice` ×4, tool round-trips (incl. error results), thinking round-trips, attachments (inline data, URL passthrough, silent bad-file-path drop, audio fallback), `response_format` (json + json_schema), sampling params, per-dialect `provider_options` merges (incl. the adapter-name-keyed compat case), catalog-driven reasoning effort and prompt cache (beta header), and the count-tokens wire route.
- **Decode** — finish-reason mappings, each dialect's distinct usage arithmetic (anthropic direct cache reads with `reasoning_tokens: 0`; openai-responses cached/reasoning subtraction; gemini `(prompt − cached) + tool_use_prompt`; compat prompt/completion only), thinking/tool/opaque items, dual-id (`fc_…`/`call_…`) preservation.
- **Stream** — tool-call and reasoning deltas, error events (pinning `retryable`/`failover_eligible`), and each dialect's stream-end contract: anthropic emits no `Finish` without `message_stop`; openai_compatible synthesizes one only if content started (both halves of the minimax tolerance pinned); gemini synthesizes unconditionally.

Notable current behaviors pinned as-is (documented divergences, not changed here): `ToolResult.image_data` is dropped by every encoder; `ToolChoice::None` drops the whole `tools` array on anthropic only; canonical `Thinking` parts are dropped by openai-responses/gemini; `Request.metadata` is dropped by compat/gemini; gemini ignores `reasoning_effort` and mints synthetic UUID tool-call/response ids (normalized to `[UUID]` in snapshots).

## Test plan

- `cargo nextest run -p fabro-llm` — 498 passed (new `it` target run twice to verify snapshot determinism incl. UUID normalization)
- `cargo +nightly fmt --check --all` / `cargo +nightly clippy -p fabro-llm --all-targets -- -D warnings` — clean
- `fabro-llm/tests/integration.rs` and `fabro-agent/tests/it/parity_matrix.rs` untouched

### DIFF
--- a/lib/crates/fabro-llm/tests/it/main.rs
+++ b/lib/crates/fabro-llm/tests/it/main.rs
@@ -1,0 +1,2 @@
+mod support;
+mod wire;
--- a/lib/crates/fabro-llm/tests/it/main.rs
+++ b/lib/crates/fabro-llm/tests/it/main.rs
@@ -1,2 +1,7 @@
+#![allow(
+    clippy::absolute_paths,
+    reason = "This test module prefers explicit type paths over extra imports."
+)]
+
 mod support;
 mod wire;
--- a/lib/crates/fabro-llm/tests/it/support.rs
+++ b/lib/crates/fabro-llm/tests/it/support.rs
@@ -1,8 +1,20 @@
-//! Shared helpers for capturing the wire requests adapters send.
+//! Shared helpers for capturing the wire requests adapters send, plus the
+//! canonical request corpus pinned across all four provider dialects.
 
 use std::sync::{Arc, Mutex};
 
+use fabro_llm::provider::ProviderAdapter;
+use fabro_llm::types::{
+    AudioData, ContentPart, DocumentData, ImageData, Message, Request, ResponseFormat, Role,
+    ThinkingData, ToolCall, ToolChoice, ToolDefinition, ToolResult,
+};
+use fabro_model::Catalog;
+use fabro_model::catalog::LlmCatalogSettings;
 use httpmock::prelude::*;
+
+// ---------------------------------------------------------------------------
+// Wire capture
+// ---------------------------------------------------------------------------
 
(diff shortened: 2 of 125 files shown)


brynary:
@swerner can we externalize the snapshots out of the `*.rs` files? The agents do better with smaller files.

I _think_ we already have a library in place to do snapshot-based testing with external snapshots. If not, I have used https://insta.rs/ for that in the past.

PR #472, opened 2026-06-03
Interpolation foundation (InterpString v2)

swerner:
# Interpolation foundation (InterpString v2)

First step of unifying config-string interpolation across Fabro. This PR is the
**behavior-neutral foundation** only — it introduces the type machinery and a
clippy gate, but changes no field's interpolation behavior. The actual field
work follows as separate stacked PRs, sequenced **reduce-first**: narrowing
changes (demote fields that shouldn't interpolate, de-template DOT attrs) land
before capability additions (resolve env in MCP / prepare / hooks).

## Why

Config strings interpolate `{{ ... }}` inconsistently today — some fields
resolve `{{ env.X }}`, others are typed as if they do but silently pass the
literal template text downstream. We're converging on three field types
(`String`, `InterpString`, and later an importable template for prompts/goals)
with four namespaces (`env`, `vars`, `secrets`, `inputs`). This PR lays the
`InterpString` foundation; it does not migrate any field.

## What's in it

- Segments generalize to `Token { namespace, name }` with a `Namespace` enum
  (`env`/`vars`/`secrets`/`inputs`). `secrets`/`inputs` are **reserved** —
  parsed as tokens ahead of their resolvers.
- `ResolveCtx` with per-namespace lookups. `resolve_with()` fails loudly
  (`Unavailable`) for a token whose namespace isn't provided in context;
  `substitute_with()` substitutes provided namespaces and preserves the rest.
  `resolve()` / `substitute_variables()` are thin wrappers over one core path.
- `ResolveEnvError` → `ResolveError { namespace, name, kind: Missing | Unavailable }`
  (message text unchanged for env/vars; the kind no longer bakes the namespace
  in, so it scales to four namespaces without an enum explosion).
- `Provenance` tracks secret-sourced names alongside env-sourced, for uniform
  redaction later.
- **`as_source()` is clippy-gated** (`disallowed-methods`). It keeps its name;
  every call site carries an `#[expect(..., reason)]` classifying it
  (serialization, error display, known-leak-pending-fix, demotion-pending,
  test). The lint turns the leak surface into a greppable, reasoned work-list
  and the method stays for its permanent uses (serde round-trip of the
  unresolved template + diagnostics).
- fabro-server: five duplicate `process_env_var` facades and two duplicate
  `resolve_interp` helpers consolidated into one `crate::interp` module.

## Behavior changes (honest list)

- **`{{ secrets.* }}` / `{{ inputs.* }}` are now reserved.** On main they
  weren't recognized as tokens → silent literal passthrough. Now, at
  `resolve()` consumers they **fail loud** (`Unavailable`) instead of passing
  the literal string through (nobody wants the literal characters as a value —
  strictly better, but technically a change). At `as_source` sites they
  round-trip unchanged. Actual resolution lands in later enhancing PRs.
- Some fabro-server resolution errors gain a `"failed to resolve <source>"`
  context line.

Otherwise behavior-neutral: every field resolves exactly as it did on main.

## What's deferred to follow-up PRs (reduce-first order)

- **Reducing / cleanup (next):** demote leak fields to `String`
  (`run.model.*`, `cli.exec.model.*`, `run.git.author.*`,
  `run.scm.owner/repository`); de-template `condition`/`label`/`model`/
  `provider`/`speed` and `output_schema`.
- **Enhancing (after):** resolve `{{ env.* }}` in MCP transports, prepare
  steps, and hooks; wire `secrets`/`inputs`.

## Verification

- `cargo build --workspace`
- `cargo nextest run --workspace` → 6449 passed, 181 skipped

## Reviewer notes

- The reserved-namespace `Unavailable` error for `secrets`/`inputs` is
  **intentional**, not a missing case — they're parsed ahead of their
  resolvers so misuse fails loud instead of leaking.
- `as_source` is clippy-gated but keeps its name deliberately — the gate is
  the enforcement; renaming was avoided as unnecessary churn.

### DIFF
--- a/clippy.toml
+++ b/clippy.toml
@@ -31,6 +31,7 @@ disallowed-methods = [
   { path = "reqwest::blocking::Client::new", reason = "Use fabro_http::blocking_http_client() or fabro_http::blocking_test_http_client()", allow-invalid = true },
   { path = "reqwest::blocking::Client::builder", reason = "Use fabro_http::BlockingHttpClientBuilder::new()", allow-invalid = true },
   { path = "reqwest::get", reason = "Build a fabro_http client and send the request explicitly", allow-invalid = true },
+  { path = "fabro_types::settings::interp::InterpString::as_source", reason = "Returns the unresolved template source, which leaks {{ ... }} tokens as literal text downstream. Resolve via resolve()/resolve_with() or substitute via substitute_with() instead; document intentional raw-source access (serialization, error messages, deliberate source preservation) with #[expect(clippy::disallowed_methods, reason = \"...\")]", allow-invalid = true },
 ]
 disallowed-types = [
   { path = "std::io::Read", reason = "Blocking trait; prefer tokio::io::AsyncReadExt on Tokio paths. Document intentional sync I/O with #[expect(clippy::disallowed_types, reason = \"...\")]" },
--- a/lib/crates/fabro-checkpoint/src/author.rs
+++ b/lib/crates/fabro-checkpoint/src/author.rs
@@ -51,6 +51,11 @@ impl GitAuthor {
     }
 }
 
+#[expect(
+    clippy::disallowed_methods,
+    reason = "raw source is today's behavior; run.git.author.* is slated for demotion to plain String in the \
+              interpolation unification (D2)"
+)]
 impl From<&GitAuthorLayer> for GitAuthor {
     fn from(value: &GitAuthorLayer) -> Self {
         Self::from_options(
@@ -60,6 +65,11 @@ impl From<&GitAuthorLayer> for GitAuthor {
     }
 }
 
+#[expect(
+    clippy::disallowed_methods,
+    reason = "raw source is today's behavior; run.git.author.* is slated for demotion to plain String in the \
+              interpolation unification (D2)"
+)]
 impl From<&GitAuthorSettings> for GitAuthor {
     fn from(value: &GitAuthorSettings) -> Self {
         Self::from_options(
--- a/lib/crates/fabro-cli/src/commands/exec.rs
+++ b/lib/crates/fabro-cli/src/commands/exec.rs
@@ -276,6 +276,11 @@ impl ProviderAdapter for AuthenticatedFabroServerAdapter {
     }
 }
 
+#[expect(
+    clippy::disallowed_methods,
+    reason = "raw source is today's behavior; cli.exec.model.* is slated for demotion to plain String in the \
+              interpolation unification (D2)"
+)]
 pub(crate) async fn execute(mut args: ExecArgs, ctx: &CommandContext) -> AnyResult<()> {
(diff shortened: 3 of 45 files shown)


swerner:
~This PR still ended up pretty large, I could break it up into two if you're like - one for the clippy warnings on the as_source() and another for the MCP fix - which will be closer to the size of the incremental fixes in the PRs going forward for the different fields, but thought it would be nice to have one bug fix come in this PR as well.~

Ended up breaking it apart

swerner:
Ran `/simplify` over this PR and the stacked demote branch (4 parallel review agents: reuse, simplification, efficiency, altitude). The fixes were amended into the commits they belong to; nothing changes the PR's behavior envelope.

**Applied in this PR (foundation commit):**
- `Namespace` derives `strum::EnumString` instead of the hand-written string→variant match in `parse_token` — one canonical variant↔string mapping for both directions, per the repo's strum convention.
- New `InterpString::resolve_or_source(lookup)` carries the D3 raw-source-fallback justification once; `resolve_permissions`, `resolve_env`, and `display_interp` had the identical `map_or_else(as_source)` expression and `#[expect]` triplicated. The eventual D3 hard-error flip is now a single-function change.
- `substitute_variables_in_str` parses once and short-circuits when there are no vars tokens (callers were parsing the same string twice).
- fabro-server: new `interp::resolve_interp_with(value, lookup)` core; the process-env free function and `AppState::resolve_interp` both delegate to it. Error text is unchanged — the "failed to resolve" context stays on the process-env path only.
- Deleted `run_manifest.rs::resolve_working_directory`, a byte-for-byte copy of `fabro_config::project::resolve_working_directory_from_run` including its duplicated "known leak (Phase 2 keep-rows)" annotation — one home, one work-list row.
- Hoisted the per-test `#[expect(clippy::disallowed_methods, reason = "test asserts the raw template source")]` boilerplate to module/file level in the multi-occurrence files.

**Applied on the stacked demote branch (not this PR):** parser-accurate `warn_if_demoted_template` (no more false warnings on never-claimed `{{ ... }}` text), the missing `cli.exec.model.*` warn coverage, the `substitute_string` double-parse removal, and test-assert `.clone()` cleanups.

**Flagged but deliberately skipped:**
- Collapsing the four `with_*` builders into one `with(namespace, lookup)` — the named builders are the clearer public API.
- Removing the secrets/inputs machinery that has no production callers yet — that's this PR's point (reserving the namespaces ahead of Phase 2).
- Making `ResolveError` carry the template source, which would eliminate the "raw source shown in the error message" expect category — real error-type design change; Phase 2 candidate.
- Routing the remaining process-env resolve call sites (`get_github_repo`, runs storage root, `server_storage_dir`) through the `AppState::env_lookup` test seam — changes which lookup is consulted; Phase 2 candidate.

Verified: workspace clippy `-D warnings` clean, full nextest pass (only the 5 known env-dependent fabro-workflow credential tests fail locally, identical on the clean tree), CI green on the amended commit.

PR #474, opened 2026-06-03
Limit DOT templates to prompt + goal

swerner:
# Limit DOT templates to prompt + goal

Part of unifying config interpolation in Fabro. Per the field taxonomy, full
MiniJinja templates (`ImportableTemplate`) should be limited to **`prompt`
(node) and `goal` (graph)** — the content fields that legitimately need
`{{ inputs.* }}` / `{{ goal }}`. Every other graph/node/edge attribute should
be a plain value, not a Turing-complete template.

This is a **behavior-reducing** slice and is **independent of the InterpString
foundation PR** (it touches the MiniJinja/template-engine path, not the
`InterpString` config path), so it branches off `main` and can be reviewed on
its own.

## What changes

- `TemplateTransform::render_attrs` still renders node `prompt` (unchanged) and
  the graph `goal` (rendered separately, as before), but **no longer renders**
  `label`, `model`, `provider`, `speed`, edge `label`, or `condition`. Those
  are left as literal text.
- When a now-demoted attribute still contains `{{ … }}` / `{% … %}`, a
  `detemplated_attribute` **warning** is emitted so authors can migrate (the
  syntax is now literal, not rendered).
- `condition` keeps its dedicated routing-expression evaluator
  (`evaluate_condition` / `parse_condition_expr`); only the Jinja pre-render is
  removed, so routing still works exactly as before.
- `output_schema` becomes a string-or-`@file` value, not a template:
  `FileInliningTransform` still resolves an `@file` reference but loads its
  contents **verbatim**, and neither the inline value nor the loaded file is
  MiniJinja-rendered.

`prompt` and `goal` are unaffected — both inline and `@file` forms are still
MiniJinja-rendered (the `@` only selects whether the template is in-band or
loaded from a file).

## Behavior change

`{{ … }}` in a demoted attribute (`label`/`model`/`provider`/`speed`/
`condition`/`output_schema`) is now **literal text** instead of being rendered.
A parse-time `detemplated_attribute` warning flags any remaining occurrences so
they're not silently dropped. This was rarely a sensible thing to do anyway
(e.g. `label = "{{ goal }}"` would splat the entire goal into a short display
label).

## Verification

- `cargo build --workspace`
- `cargo nextest run -p fabro-workflow` → 1164 passed
- No pending `insta` snapshots

## Tests

- `template_transform_renders_prompt_and_leaves_other_attrs_literal` — `prompt`
  still renders; node/graph/edge `label` stay literal; one migration warning per
  demoted label.
- `file_inlining_transform_does_not_render_templates_in_output_schema` and
  `file_inlining_transform_loads_output_schema_file_verbatim` — `output_schema`
  inline and `@file` contents are used verbatim, no Jinja.

### DIFF
--- a/lib/crates/fabro-workflow/src/transforms/file_inlining.rs
+++ b/lib/crates/fabro-workflow/src/transforms/file_inlining.rs
@@ -192,45 +192,52 @@ impl FileInliningTransform {
             .with_inputs(self.inputs.clone());
 
         for (node_id, node) in &mut graph.nodes {
-            for attr_name in ["prompt", "output_schema"] {
-                let Some(AttrValue::String(attr_value)) = node.attrs.get(attr_name) else {
-                    continue;
-                };
+            // `prompt` is an importable template: MiniJinja-render the value,
+            // then inline any `@file` reference (whose contents are rendered
+            // too). Clone up front so the immutable borrow ends before we
+            // re-insert.
+            let prompt = match node.attrs.get("prompt") {
+                Some(AttrValue::String(value)) => Some(value.clone()),
+                _ => None,
+            };
+            if let Some(attr_value) = prompt {
                 let target = TemplateRenderTarget::node_attr(
                     self.source_name.clone(),
                     node_id.clone(),
-                    attr_name,
+                    "prompt",
                 )
-                .with_source_origin(self.source_text.as_deref(), attr_value)
+                .with_source_origin(self.source_text.as_deref(), &attr_value)
                 .with_template_store(template_render_store(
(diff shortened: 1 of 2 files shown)


brynary:
@swerner I'm planning to merge this one on green, LMK if there's another order you think is preferrable

swerner:
@brynary sounds good - yeah this is the only one that could be done in parallel - the rest of the InterpString ones all kind of fall into a sequence (though I'll follow up first with the behavior reducing PR first rather than adding the env support just yet)

PR #481, opened 2026-06-09
refactor(llm): introduce Codec trait seam + extract openai_compatible

swerner:
## What

Introduces the `Codec` / `StreamDecoder` trait seam in `fabro-llm` and extracts the OpenAI Chat Completions wire logic behind it as the first conforming codec. Two commits:

1. **`codec/mod.rs`** — the pure translation contract (`encode` / `decode_response` / `stream_decoder`, plus defaulted `encode_count_tokens` / `decode_count_tokens` / `decode_error`) and its data types (`CodecCtx`, `CodecParams`, `EncodedRequest`, `RawEvent`). A codec knows *what the bytes say*; it owns no HTTP, auth, or base URL.
2. **`codec/openai_compatible/`** — the Chat Completions codec split into `wire` / `translate` / `request` / `response` / `stream`. `providers/openai_compatible.rs` shrinks from 1,608 → ~330 lines: a thin transport shell that keeps the public struct/builders/auth/`validate_request`, owns the streaming byte loop + SSE `data:` framing, and delegates all translation to the codec. The two hand-rolled stream unfolds collapse into one.

This is the first step of a gateway refactor that separates codec (wire dialect) from transport/auth/route, so later work (Bedrock, OpenRouter) becomes mostly config rather than parallel adapters.

## Behavior

No behavior change. The public adapter API (`OpenAiCompatibleAdapter::new` / `with_name` / `with_catalog` / …) is unchanged, and **all 126 wire snapshots pass without edits** — the parity proof that the extracted codec produces byte-identical output. The 29 in-module unit tests move into the codec submodules alongside the code they exercise.

## On the trait

`openai_compatible` is the simplest dialect, so its `impl Codec` is just three methods — count-tokens and error mapping inherit the defaults. The contract is defined in full now (a scoped `dead_code` allow on `codec/mod.rs` covers the seams the anthropic/openai/gemini codecs will exercise in follow-up PRs) so those extractions only *override* methods, never extend the trait.

Extracting a real codec refined two trait signatures vs. the initial sketch: the canonical `Request` lives in `CodecCtx` (decoders need it for tool-argument parsing and the stream model fallback), and the header-parsed `rate_limit` threads into `decode_response` / `stream_decoder`. `on_event` returns `Result` so dialect error events propagate as stream errors.

## Tests

- `cargo nextest run -p fabro-llm` — 515 passed (incl. 126 wire snapshots, unmodified)
- `cargo nextest run --workspace` — green
- fabro-agent `parity_matrix` (the frozen `OpenAiCompatibleAdapter` contract) — green
- `cargo +nightly fmt --check` / `clippy --all-targets -- -D warnings` — clean

### DIFF
--- a/lib/crates/fabro-llm/src/codec/mod.rs
+++ b/lib/crates/fabro-llm/src/codec/mod.rs
@@ -1,0 +1,145 @@
+//! The codec seam: pure, sync translation between the canonical core
+//! (`Request`/`Response`/`StreamEvent`) and a provider wire dialect.
+//!
+//! A codec knows *what the bytes say*. It does NOT know how they travel
+//! (auth, base URL, retries, streaming transport) — that's the adapter/
+//! transport layer. Everything a codec varies on arrives as data in
+//! [`CodecCtx`] / [`CodecParams`]; codecs hold no per-request state.
+//!
+//! The trait is intentionally complete (count-tokens + error mapping have
+//! defaults) so the per-dialect codecs that follow only ever *override*
+//! methods, never extend the contract.
+
+#![allow(
+    dead_code,
+    reason = "Codec seam introduced ahead of its first implementation; \
+              openai_compatible conforms in the next commit of PR 2."
+)]
+
+use fabro_model::Model;
+
+use crate::error::{Error, error_from_status_code};
+use crate::providers::common::parse_error_body;
+use crate::types::{Request, Response, StreamEvent};
+
+/// Per-request context. Borrowed — the codec reads what it needs and returns.
+pub(crate) struct CodecCtx<'a> {
+    /// Identity stamped into `Response.provider`, and the `provider_options`
+    /// namespace key for the openai_compatible codec (kimi/zai/…).
+    pub provider_name: &'a str,
+    /// The model id to send on the wire — catalog `api_id`, resolved by the
+    /// route (today `api_id == id` everywhere).
+    pub deployment_id: &'a str,
+    /// Model row for capability lookups (prompt_cache, reasoning levels,
+    /// max_output). `None` when no catalog is injected.
+    pub model:         Option<&'a Model>,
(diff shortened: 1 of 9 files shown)


swerner:
Ran a `/simplify` cleanup pass over the extraction (4 review angles: reuse / simplification / efficiency / altitude). Applied in 7cc654d, no behavior change — all 126 wire snapshots still pass unmodified, clippy/fmt clean.

**What changed:**

- **`Codec::decode_error` now takes the transport-parsed `retry_after`** (same pattern as `rate_limit` threading into `decode_response`/`stream_decoder`). This let the `stream()` error path delegate to `codec.decode_error(...)` instead of hand-rolling `parse_error_body` + `error_from_status_code` in the shell — the one place the adapter was still doing codec-layer work. Later codecs now get dialect error mapping by overriding one method instead of patching their adapters. The `dead_code` allow comment was narrowed to the count-tokens routes accordingly.
- **Deduped `complete()`/`stream()`** via two private helpers: `codec_ctx()` (the borrowed context setup) and `encoded_request()` (encode → URL → body → dialect headers). The request timeout stays in `complete()` only, since `stream()` uses the line reader's read timeout.
- **`StreamLoop.pending` is now a `VecDeque`** — drops the stored-reversed-Vec invariant and both `events.reverse()` calls in the unfold (`Vec → VecDeque` is O(1)).
- **`stream.rs` `test_state` builds through `StreamState::new`** with a minimal request (the `minimal_request` pattern from the other provider tests) instead of mirroring the constructor field-by-field, so new `StreamState` fields don't require a parallel test edit.

**Reviewed and intentionally skipped:** `StreamState::new`'s string clones (required by the `'static` decoder box), the clone-heavy `finish_events()` assembly (pre-existing, imposed by the `StreamEvent::Finish` shape), and a `StreamLoop::new` constructor (single construction site).

PR #487, opened 2026-06-10
refactor(llm): extract codec/openai_responses behind the Codec trait

swerner:
## Summary

Next dialect extraction in the gateway refactor series (after #481 / #485, sibling of the anthropic extraction): the OpenAI Responses API wire translation moves out of `providers/openai.rs` into `codec/openai_responses/`, behind the `Codec` / `StreamDecoder` traits. The adapter becomes a thin transport shell (2,784 → 692 lines) owning auth (bearer + org/project headers), base URL, the streaming byte loop, and route config; all translation is in the codec.

Two commits, each independently green:
2. **Rewire the adapter** to it and migrate the ~54 unit tests into the codec submodules they now cover.

Key moves:
- **Codex mode splits along the codec seam**: encode-side param omission (`temperature`/`top_p`/`max_output_tokens` omitted, `instructions` always sent) rides on a new `CodecParams::openai_codex` flag; the transport-side half (blocking requests served via streaming) is route config on the adapter. No provider-name branching — codex is OpenAI's only route split.
- **`translate_input` goes sync**: its only async-ness was file-path image loading, now handled by the shared `attachments::resolve` (#485) in the adapter before encode (images only; audio/documents render as text placeholders in the codec without I/O).
- The invariant-dense pieces move wholesale, already pure: opaque `openai_reasoning`/`openai_message` item round-trip, the `fc_…`/`call_…` dual-id preservation via `provider_metadata`, custom-tool (apply_patch) emission and raw-input accumulation, `store: false` + `include: ["reasoning.encrypted_content"]`.
- The SSE state machine becomes `SseAccumulator` behind `StreamDecoder`: the transport owns byte reading + framing; the decoder is fed framed `RawEvent`s, resolves the event type from the SSE `event:` line or the JSON `type` field, and `finish()` synthesizes nothing (`response.completed`/`incomplete` are the finishers — matching the old EOF behavior exactly).

Coordination note: this PR makes the same unit→fielded `CodecParams` change as the sibling anthropic extraction (each adds only its own fields) — whichever lands second resolves a trivial field-union conflict in `codec/mod.rs`.

## Behavior preservation

No behavior change. The 33 openai_responses wire snapshots from #471 (codex mode, dual-id round-trip, opaque items, attachment drop-on-error, response_format, streaming happy path / tool deltas / reasoning deltas / failure events) pass unmodified, and the full fabro-llm suite is back to count (516: all 54 migrated tests plus one new test pinning the count-tokens endpoint + filtered body on the codec).

## Testing

- `cargo nextest run -p fabro-llm` — 516 passed (126 wire snapshots included)
- `cargo check --workspace`

### DIFF
--- a/lib/crates/fabro-llm/src/codec/anthropic_messages/encode.rs
+++ b/lib/crates/fabro-llm/src/codec/anthropic_messages/encode.rs
@@ -610,7 +610,8 @@ reasoning = true
     fn direct_params() -> CodecParams {
         CodecParams {
             anthropic_version: AnthropicVersion::Header("2023-06-01"),
-            anthropic_beta:    true,
+            anthropic_beta: true,
+            ..CodecParams::default()
         }
     }
 
--- a/lib/crates/fabro-llm/src/codec/mod.rs
+++ b/lib/crates/fabro-llm/src/codec/mod.rs
@@ -12,6 +12,7 @@
 
 pub(crate) mod anthropic_messages;
 pub(crate) mod openai_compatible;
+pub(crate) mod openai_responses;
 
 use fabro_model::Model;
 
@@ -42,7 +43,8 @@ pub(crate) struct CodecCtx<'a> {
 /// Per-route dialect knobs, expressed as data so one codec can serve several
 /// routes. The default is inert ("nothing special"); a route that needs a
 /// dialect quirk sets the relevant field. Grows as codecs need it — #459 adds
-/// `ModelPlacement` for Bedrock.
+/// `ModelPlacement` for Bedrock. Inert for codecs that don't read a given
+/// field.
 #[derive(Debug, Default, Clone)]
 pub(crate) struct CodecParams {
     /// Where/whether to place the Anthropic API version. Direct Anthropic uses
@@ -52,6 +54,12 @@ pub(crate) struct CodecParams {
     /// Whether to emit Anthropic beta headers (prompt-caching / fast-mode /
     /// 1M-context). True on the direct route, false for Kimi-over-anthropic.
     pub anthropic_beta:    bool,
+    /// Codex-endpoint dialect for the openai_responses codec: omit the
+    /// sampling params (`temperature`/`top_p`/`max_output_tokens`) the Codex
+    /// endpoint rejects and always send `instructions` (empty string when the
+    /// request has none). The transport-side half of codex mode (forced
+    /// streaming) is route config, not codec data.
+    pub openai_codex:      bool,
 }
 
 /// Placement of the Anthropic API version on the wire.
--- a/lib/crates/fabro-llm/src/codec/mod.rs
+++ b/lib/crates/fabro-llm/src/codec/mod.rs
@@ -91,8 +91,9 @@ pub(crate) struct EncodedRequest {
 
 /// One framed item off the byte stream, handed to a [`StreamDecoder`].
 pub(crate) struct RawEvent<'a> {
-    /// SSE `event:` type — `Some` for anthropic; `None` for the data-only
-    /// framing openai/gemini use.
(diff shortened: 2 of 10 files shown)


swerner:
## Simplify pass (cc8039b61)

Ran a 4-angle review (reuse / simplification / efficiency / altitude) over the codec extraction and applied the cleanups. Net −128 lines; all 516 fabro-llm tests pass, emitted stream events are byte-identical.

### Applied

**Adapter (`providers/openai.rs`)**
- Replaced the 4× hand-built `CodecCtx` literals with a `codec_ctx()` helper, matching the gemini (#489) and openai_compatible adapters. `stream()` now builds the ctx once — the two scope blocks and deferred `let rate_limit;` are gone.
- Deleted `RouteConfig`: both fields were copies of `self.codex_mode`. `complete()` checks it directly; a small `codec_params()` covers the encode side.
- `stream()`'s non-2xx branch now routes through `Codec::decode_error` instead of inlining the trait default's body — the module doc already claimed it did, and the sibling adapters do.

**Codec (`codec/openai_responses/`)**
- `decode_response` parses the body once (was two full JSON parses per blocking response).
- The function_call/custom_tool_call → `ToolCall` construction (the call-id/item-id round-trip rules) existed in four near-identical copies across `decode.rs` and `stream.rs`; now one shared `tool_call_from_item()`.
- `handle_tool_call_delta` had two O(n²)-per-stream patterns (rebuilding custom-tool `arguments` from a clone of the full accumulated string each delta; a redundant second scan ending in a full `ToolCall` clone with an unreachable fallback). Now one lookup, in-place accumulation, identical event payloads.
- `parse_output` moves output items into content parts instead of cloning them (encrypted reasoning blobs made those clones non-trivial); final stream response moves (`mem::take`) tool calls + accumulated text; dead `raw_response` field removed.
- `tool_call_types: HashMap<String, (String, String)>` carried a never-read tuple half → `custom_call_ids: HashSet<String>`.
- Count-tokens body filtered in place via `retain` instead of cloning every kept field. This reordered keys in one wire snapshot (`model` first) — same fields/values.

### Noted but deferred (cross-PR concerns for the #487/#488/#489 joint pass)

- `StreamLoop` + the unfold driver + the SSE framer are per-adapter copies; the codec module comment defers transport consolidation to a later PR, so this copy was kept byte-compatible rather than extracted early.
- `attachments::resolve` (#485 infra) deep-clones the whole `Request` even with no file-backed attachments — the common case. A short-circuit there would benefit all three codec PRs.
- The `apply_patch` name hardcoded in the codec to classify orphaned tool results — real fix is carrying `tool_type` on `ToolResult` (canonical-type change, not a cleanup).
- Blocking paths still pass the `"type"` error-field literal to `send_and_read_response` rather than `decode_error`; gemini's `send_for_body` shape is the better pattern, but adopting it here would lose the operation-tagged error logging this adapter's tests pin. Candidate for hoisting into `providers/common`.
- Audio/document placeholder strings duplicated with the openai_compatible codec; `minimal_request()` test fixture now has a fourth copy — both follow existing precedent, better unified across the three branches at once.

swerner:
Rebased onto main now that #488 has landed. Beyond the clean replay, the adaptations are isolated in the top commit (06712f7d5):

- **`CodecParams` union** (the expected conflict in `codec/mod.rs`): now carries both the anthropic route knobs (`anthropic_version`/`anthropic_beta`, from #488) and this PR's `openai_codex`. The per-route constructors spread `..CodecParams::default()` for the fields they don't set.
- **Adopted #488's `Cow`-returning `attachments::resolve`**: the openai adapter's `resolve_request` now borrows the request untouched when no file-backed attachments need loading, instead of deep-cloning the conversation per call.

All 526 fabro-llm tests green (including #488's anthropic wire pins, unmodified), clippy/fmt clean, workspace compiles.

PR #488, opened 2026-06-10
refactor(llm): extract codec/anthropic_messages behind the Codec trait

swerner:
## Summary

Dialect extraction in the gateway refactor series (after #481 / #485, sibling of #487): the Anthropic Messages wire translation moves out of `providers/anthropic.rs` into `codec/anthropic_messages/`, behind the `Codec` / `StreamDecoder` traits. The adapter becomes a thin transport shell owning auth, base URL, the streaming byte loop, and route config; all translation is in the codec.

Three commits, each independently green:
2. **Rewire the adapter** to it and migrate the ~70 unit tests into the codec submodules they now cover.
3. **Port #482's Claude Fable 5 handling into the codec layout** (see below).

Key moves:
- **Route config replaces the request-time `provider_name == "anthropic"` branches**: auth scheme (x-api-key vs bearer), version/beta headers, the count-tokens availability gate, and Kimi-over-anthropic forced streaming resolve once per call into a `RouteConfig`. Dialect headers ride on `CodecParams` (`AnthropicVersion::Header("2023-06-01")` + beta-header emission for the direct route; inert defaults for Kimi).
- **`build_api_request`'s `(ApiRequest, RequestBuilder)` dual-return dies**: codec `encode` produces body + headers as data (`EncodedRequest`); the transport applies them. This also kills the duplicated header rebuild in `count_input_tokens`.
- **Encode goes sync**: file-backed Image/Document attachments resolve to inline data via the shared `attachments::resolve` (#485) in the adapter before encode (drop-on-error preserved; audio stays a text placeholder in the codec).
- The SSE state machine becomes `SseAccumulator` behind `StreamDecoder`: the transport owns byte reading + `\n\n` framing; the decoder is fed framed `RawEvent`s. `finish()` returns nothing — `message_stop` is the only finisher, matching today's no-synthesis contract.
- json_schema synthetic-tool machinery (encode injection, decode extraction, stream rewrite) moves intact around the shared `SYNTHETIC_TOOL_NAME`.

### The #482 (Claude Fable 5) port

#482 modifies the old-layout `anthropic.rs` directly, so this branch re-homes its behavior into the codec structure (commit 3): `stop_details` on the wire type, the Fable encode gates keyed off the deployment id (no default adaptive `thinking`, no `temperature`/`top_p`, no legacy 1M-context beta header — which now lands **once** instead of twice, since both routes share `build_headers`), refusal → failover-eligible content-filter errors in decode and stream, and the `validate_request` rejection of manual thinking configs. The port is inert until the Fable catalog entry lands. Validated by merging #482's head into this branch on a scratch branch: the only conflict is `anthropic.rs` (resolved as this branch's version), and **all of #482's Fable/refusal tests pass against the codec implementation** (521 fabro-llm tests + fabro-model/fabro-workflow 1286 green on the merged tree). If #482 merges first, this PR's rebase resolves the same single-file conflict the same way.

Coordination note: this PR makes the same unit→fielded `CodecParams` change as #487 (each adds only its own fields) — whichever lands second resolves a trivial field-union conflict in `codec/mod.rs`.

## Behavior preservation

No behavior change. The anthropic wire snapshots from #471 (direct route, Kimi-over-anthropic bearer/no-version pin, prompt-cache with catalog, json_schema, count-tokens wire, streaming happy path / tool deltas / error events / no-message_stop-no-Finish) pass unmodified, and the full fabro-llm suite is back to count (515).

## Testing

- `cargo nextest run -p fabro-llm` — 515 passed (126 wire snapshots included)
- Scratch-merge validation against #482's head — 521 passed incl. its 6 Fable/refusal tests; `cargo nextest run -p fabro-model -p fabro-workflow` — 1286 passed
- `cargo build --workspace`

### DIFF
--- a/lib/crates/fabro-llm/src/attachments.rs
+++ b/lib/crates/fabro-llm/src/attachments.rs
@@ -6,18 +6,9 @@
 //! Loads that fail drop the part silently — the long-standing contract — and
 //! non-file URLs and already-inline data pass through untouched.
 //!
-//! Shared infra introduced ahead of its consumers: the per-dialect codecs
-//! (anthropic/openai_responses/gemini) each construct their own
-//! [`AttachmentPolicy`] and call [`resolve`] from their adapter shells when
-//! they are wired. Until the first of those lands, nothing here is reachable.
-
-// Each dialect adapter that wires this in removes the allow as part of its
-// rewire; harmless if it lingers when several land in parallel.
-#![allow(
-    dead_code,
-    reason = "Attachment-resolution infra added ahead of the dialect codecs (PRs 3-5) that \
-              construct an AttachmentPolicy and call resolve from their adapter shells."
-)]
+//! Shared infra for the per-dialect codecs (anthropic/openai_responses/gemini):
+//! each constructs its own [`AttachmentPolicy`] and calls [`resolve`] from its
+//! adapter shell.
 
 use crate::providers::common;
 use crate::types::{AudioData, ContentPart, DocumentData, ImageData, Request};
--- a/lib/crates/fabro-llm/src/attachments.rs
+++ b/lib/crates/fabro-llm/src/attachments.rs
@@ -10,6 +10,8 @@
 //! each constructs its own [`AttachmentPolicy`] and calls [`resolve`] from its
 //! adapter shell.
 
+use std::borrow::Cow;
+
 use crate::providers::common;
 use crate::types::{AudioData, ContentPart, DocumentData, ImageData, Request};
 
@@ -23,9 +25,15 @@ pub(crate) struct AttachmentPolicy {
     pub audio:     bool,
 }
 
-/// Return a copy of `request` with file-path attachments (per `policy`)
-/// resolved to inline data. Parts whose file fails to load are dropped.
-pub(crate) async fn resolve(request: &Request, policy: AttachmentPolicy) -> Request {
+/// Resolve file-path attachments (per `policy`) to inline data. Parts whose
+/// file fails to load are dropped. Borrows the request untouched in the common
+/// case where nothing needs loading; only requests with policy-matching
+/// local-file parts pay for a copy.
+pub(crate) async fn resolve(request: &Request, policy: AttachmentPolicy) -> Cow<'_, Request> {
+    if !needs_resolution(request, policy) {
+        return Cow::Borrowed(request);
+    }
+
     let mut resolved = request.clone();
     for message in &mut resolved.messages {
         let mut new_content = Vec::with_capacity(message.content.len());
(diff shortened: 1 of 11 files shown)


swerner:
## Simplify pass (bd08ea7cf)

Ran the same 4-angle review (reuse / simplification / efficiency / altitude) as #487 and applied the cleanups. All 525 fabro-llm tests pass; clippy and fmt clean.

Note: this lands on top of the rebase onto #482 — the port commit (f59671f32) was realigned to the **merged** #482 semantics first (catalog-feature gating via `supports_sampling_params`/`AlwaysAdaptive` instead of `deployment_id == "claude-fable-5"` string matching, default adaptive `thinking` kept alongside `output_config.effort`, 1M beta header removed entirely, refusal errors labeled with the response model). #482's wire tests pinned all of it.

### Applied

**Adapter (`providers/anthropic.rs`)**
- Added the `codec_ctx()` helper and build the ctx once in `stream()` — same 4× `CodecCtx` literal (twice in `stream()` with scope-block gymnastics) that #487/#489 factored out; all three adapters now share the shape.
- `stream()`'s non-2xx branch routes through `Codec::decode_error` instead of inlining the trait default's body — the codec docs already claimed this; now it's true.
- `parse_sse_block` borrows from the block (`(&str, Cow<str>)`) instead of allocating ~3 `String`s per SSE event — the borrowed `RawEvent` boundary made this a one-function fix on the streaming hot path.
- The adapter's `provider_options.anthropic.thinking` walk shares a new `anthropic_option()` accessor with the codec, which also replaced four hand-written `.get("anthropic")` traversals in `encode.rs`.

**Codec (`codec/anthropic_messages/`)**
- Deleted the `Built` carrier struct — `build_headers` derives `auto_cache`/`is_fast` from the ctx it already receives; `build_request` returns a plain `ApiRequest`.
- Replaced the `omit_tools` set-inside-a-closure side effect with a plain `matches!` predicate.
- `decode_response` deserializes `ApiResponse` from `&raw` instead of `serde_json::from_value(raw.clone())` — that clone copied the entire response tree per non-streaming completion.
- Collapsed the copy-paste Image/Document arms into one `media_block()` helper; 3× `block_{index}` derivation became `block_text_id()`.
- `ApiRequest.stop_sequences` lost its dead `Option` (the only constructor always wrapped `Some`).
- `SseAccumulator::new` takes plain data instead of a full `CodecCtx`, letting the test module's duplicate 12-field literal collapse into a call to the real constructor; `take_response` moves fields out instead of cloning; the JSON-schema delta rewrite moves the string instead of cloning per delta.

**Shared infra**
- `attachments::resolve` returns `Cow<Request>` and borrows untouched when no policy-matching local-file parts exist — previously every `complete`/`stream`/`count_tokens` deep-cloned the entire conversation even with zero attachments. Benefits #487/#489 when they pick it up.
- `codec/mod.rs`: removed the module-wide `#![allow(dead_code)]` (this PR exercises every seam it deferred) and corrected the `decode_error` doc, which claimed anthropic overrides it.

### Noted but deferred

- **`RouteConfig` kept** (unlike #487's removal): it carries four genuinely distinct knobs for two real routes (direct vs Kimi-over-anthropic), not one duplicated bool.
- **`CountTokensRequest` → filter-body-in-place** (#487's pattern): switching would also apply the provider-options merge to the count body — a behavior change pinned by wire snapshots. The two assembly paths staying separate is a known drift hazard worth a deliberate decision.
- **Moving the always-adaptive thinking rejection from `validate_request` into codec `encode`**: the deepest altitude finding — `encode` is declared fallible but never fails, while the rejection sits adapter-side where future Bedrock/OpenRouter shells must re-duplicate it. Relocates behavior #482 just placed, so deferred to a series-level decision.
- **Cross-adapter `StreamLoop`/SSE-framer consolidation, `AttachmentPolicy` on the codec, `send_and_read_response`'s `"type"` error-field param, test fixture dedup**: series-level items, same as noted on #487; deferred to the planned transport-consolidation PR.

PR #489, opened 2026-06-10
refactor(llm): extract codec/gemini_generate behind the Codec trait

swerner:
## Summary

Final dialect extraction in the gateway refactor series (after #481 / #485, sibling of #487 and #488): the Gemini `generateContent` wire translation moves out of `providers/gemini.rs` into `codec/gemini_generate/`, behind the `Codec` / `StreamDecoder` traits. The adapter becomes a thin transport shell (1,607 → ~400 lines) owning auth (`x-goog-api-key`), base URL, and the streaming byte loop; all translation is in the codec.

Two commits, each independently green:
2. **Rewire the adapter** to it, migrate the ~22 unit tests, and add the previously missing stream-decoder tests.

Gemini is the simplest route story in the series — no provider-name branching, no mode flags, count-tokens always available, no forced streaming — so there is no route config and **no `CodecParams` changes** (this PR is conflict-free with #487/#488 apart from one `mod` line; if it lands after them, the unit-struct `CodecParams` literals become `::default()` on rebase, mechanical).

It does exercise two trait seams the other codecs don't:
- **Fully-formed endpoints from the codec**: model-in-path `:generateContent` / `:streamGenerateContent?alt=sse` / `:countTokens` ride on `EncodedRequest.endpoint` (the count body wraps the request in `generateContentRequest`).
- **The first `decode_error` override**: Gemini's gRPC-status mapping (`error_from_grpc_status` with HTTP-status fallback) moves behind the codec; the adapter feeds it status + body + retry-after. The send-side timeout mapping stays transport-side.

Other moves, wholesale and already pure: synthetic-UUID tool-call/response ids, the id→name recovery map for `functionResponse`, usage arithmetic (cache subtraction + tool-use addition + thoughts→reasoning), default `safety_settings` injection (flagged profile-ish in a comment, unchanged), `thoughtSignature` round-trip, and the `provider_options.gemini` merge. `translate_messages` goes sync: file-backed Image/Audio/Document attachments resolve via the shared `attachments::resolve` (#485) before encode.

The streaming decoder preserves Gemini's distinct stream-end contract exactly: data-only SSE (no event types, no `[DONE]`), and `finish()` synthesizes the `Finish` from accumulated state unconditionally at byte-stream end — there is no terminal wire event.

## Behavior preservation

No behavior change. The 32 gemini wire snapshots from #471 (encode round-trips, attachments, response_format, provider_options merges, streaming happy path / tool deltas / reasoning deltas / the unconditional-Finish stream-end pin) pass unmodified, and the full fabro-llm suite is green at 525: all 22 migrated tests plus 10 new ones — 9 stream-decoder unit tests (gemini previously had **zero**: text/thought deltas, reasoning→text transition, single-chunk function calls, finish-reason handling, Finish synthesis with and without a wire finish reason, ToolCalls inference, malformed-chunk errors) and 1 pinning the three model-in-path endpoints.

## Testing

- `cargo nextest run -p fabro-llm` — 525 passed (126 wire snapshots included)
- `cargo check --workspace`

### DIFF
--- a/lib/crates/fabro-llm/src/codec/gemini_generate/decode.rs
+++ b/lib/crates/fabro-llm/src/codec/gemini_generate/decode.rs
@@ -169,3 +169,196 @@ pub(super) fn decode_count_tokens(body: &str) -> Result<i64, Error> {
         })?;
     Ok(response.total_tokens)
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn token_counts_disjoint_with_cache_thoughts_and_tool_use() {
+        let body = serde_json::json!({
+            "promptTokenCount": 200,
+            "cachedContentTokenCount": 180,
+            "candidatesTokenCount": 200,
+            "thoughtsTokenCount": 300,
+            "toolUsePromptTokenCount": 400
+        });
+        let meta: UsageMetadata = serde_json::from_value(body).unwrap();
+        let usage = parse_usage(Some(&meta));
+
+        assert_eq!(usage.input_tokens, 420);
+        assert_eq!(usage.cache_read_tokens, 180);
+        assert_eq!(usage.output_tokens, 200);
+        assert_eq!(usage.reasoning_tokens, 300);
+        assert_eq!(usage.cache_write_tokens, 0);
+        assert_eq!(usage.total_tokens(), 1100);
+    }
+
+    #[test]
+    fn gemini_error_uses_grpc_status_when_available() {
+        let err = gemini_error(
+            400,
+            "model not found".into(),
+            "gemini",
+            Some("NOT_FOUND".into()),
+            None,
+            None,
+        );
+        assert!(matches!(err, Error::Provider {
+            kind: ProviderErrorKind::NotFound,
+            ..
+        }));
+
+        let err = gemini_error(
+            400,
+            "bad args".into(),
+            "gemini",
+            Some("INVALID_ARGUMENT".into()),
+            None,
+            None,
+        );
+        assert!(matches!(err, Error::Provider {
+            kind: ProviderErrorKind::InvalidRequest,
+            ..
+        }));
+
+        let err = gemini_error(
+            429,
+            "rate limited".into(),
+            "gemini",
+            Some("RESOURCE_EXHAUSTED".into()),
+            None,
+            None,
+        );
+        assert!(matches!(err, Error::Provider {
+            kind: ProviderErrorKind::RateLimit,
+            ..
+        }));
+
+        let err = gemini_error(
+            401,
+            "bad key".into(),
+            "gemini",
+            Some("UNAUTHENTICATED".into()),
+            None,
+            None,
+        );
+        assert!(matches!(err, Error::Provider {
+            kind: ProviderErrorKind::Authentication,
+            ..
+        }));
+
+        let err = gemini_error(
+            403,
+            "denied".into(),
+            "gemini",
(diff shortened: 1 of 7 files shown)


swerner:
Ran a `/simplify` pass (4 parallel review agents: reuse / simplification / efficiency / altitude) over this PR. Verdict: the codec extraction itself is a clean move — the cleanup debt was concentrated in the adapter shell. Applied in 1dc732b (net −14 lines, 525 tests still green):

- **`CodecCtx` was copy-pasted 4× in the adapter, twice inside `stream()`** (flagged by all four agents). Now a `codec_ctx()` helper matching `openai_compatible.rs`, and `stream()` builds one ctx — the comment claiming the decoder had to be built "while the ctx is still in scope" described a borrow constraint that doesn't exist. Also drops the redundant second catalog lookup per streaming call, and means the planned `CodecParams` growth touches one site instead of four.
- **Triplicated attachment arms in `encode.rs`**: once #485 removed file loading, the Image/Audio/Document arms were character-identical except the default MIME. Collapsed into `media_part()`.
- **Double JSON parse in `decode_response`** (pre-existing, moved): body was parsed into `ApiResponse` and again into `Value` for `raw`. Now one `Value` parse + `ApiResponse::deserialize(&raw)`.

Deliberately **not** fixed, for the record:

- The ~100-line streaming loop / pre-stream error branch / `send_for_body` duplication vs `openai_compatible.rs` — the headline reuse finding, but `codec/mod.rs` explicitly defers transport consolidation to a later PR in the series, and hoisting shared helpers now would conflict with #487/#488. Two notes for that PR: `send_for_body` drops the `warn!` instrumentation `common::send_and_read_response_with_operation` emits, and the loops have small non-dialect deltas (`trim_end_matches('\r')` vs `trim()`, empty-`data:` skip) to reconcile.
- `attachments::resolve` deep-clones the full `Request` even when nothing needs resolving (worst case duplicating inline attachment bytes). Real cost, but it's #485 shared infra that #487/#488 also wire into — a `Cow`/pre-scan change is a good follow-up after the siblings land.
- `SseAccumulator.finished` overlapping the loop's `finished_emitted` — kept: it makes `finish()` idempotent independent of whichever transport drives it.
- Per-chunk `Vec<StreamEvent>` allocation in `on_event` — imposed by the shared `StreamDecoder` signature, not fixable locally.
- `stream.rs` re-implementing `decode::parse_part`'s functionCall parsing — pre-existing, and unifying isn't behavior-preserving (missing-name handling differs: drop vs `""`).
- `ctx.model` populated here but `None` in the openai_compatible adapter — kept populated (now in one place); aligning the sibling is outside this PR's files.

swerner:
Rebased onto main now that #488 and #487 have landed. Beyond the clean replay, the adaptations are isolated in the top commit (ecab07a9a):

- **`codec/mod.rs`**: kept main's state through the conflict — `CodecParams` stays the merged anthropic+openai union and the module-wide `#![allow(dead_code)]` stays removed (#488 dropped it); this PR just adds `mod gemini_generate`.
- **`CodecParams` construction**: this branch predated the struct gaining fields, so its `let params = CodecParams;` unit-struct call sites (3 in the adapter, 3 in test helpers) became `CodecParams::default()` — gemini sets no dialect knobs.
- **Adopted #488's `Cow`-returning `attachments::resolve`**: the gemini adapter's `resolve_request` now borrows the request untouched when no file-backed attachments need loading, instead of deep-cloning the conversation per call.

All 536 fabro-llm tests green (including the anthropic and openai wire pins from main, unmodified), clippy/fmt clean, workspace compiles.

PR #491, opened 2026-06-10
refactor(llm): consolidate the HTTP transport behind the codec seam

swerner:
PR 6 of the gateway refactor series (after #481, #488, #487, #489): collapse the four per-adapter transport copies into one `transport` module. Net −157 lines, and every cross-adapter duplication flagged in the #487/#488 simplify findings is resolved here.

## What moved where

**`transport.rs` (new)** — how bytes travel, dialect-blind:
- `HttpTransport` (promoted from `providers::http_api::HttpApi`): client, auth key, base URL, timeouts
- `LineReader` + `parse_retry_after` + `parse_rate_limit_headers` (moved from `providers::common`, re-export shims kept there for the frozen fabro-cli imports; `LineReader::new` keeps its 2-arg signature)
- `complete_via_http` / `send_for_body`: blocking send with the shared timeout/error/status warn logs, non-2xx mapped through `Codec::decode_error`
- `stream_via_http` + one SSE decode loop, parameterized by `SseFraming::{EventBlocks, DataLines}` — replaces the four verbatim `StreamLoop` + unfold copies and the four divergent framers (anthropic's `parse_sse_block`, openai's `parse_sse_message`, the inline data-line handling in openai_compatible/gemini, and fabro_server's private block parser)

**`codec/mod.rs`** — gains the dialect-neutral pure helpers `parse_error_body` and `extract_system_prompt` (moved from `providers::common`), so the codec layer no longer imports from the transport-side providers module.

**Adapters** — shrink to auth + route config + codec composition. `send_and_read_response` and its `error_code_field` parameter are deleted: the dialect error-body key now lives only in the codecs, and any future `decode_error` override applies to blocking and streaming paths alike.

## Unified SSE framing semantics (deliberate decisions)

The four framers disagreed on edge cases; the shared framer picks one behavior, stated here rather than chosen silently:
- data payloads are trimmed; multi-line `data:` payloads join with `\n`; CRLF tolerated in both modes
- comment (`:`), blank, and non-data lines are skipped
- events with an **empty payload are dropped** rather than handed to the decoder — previously anthropic would error the whole stream on a bare `data:` line and openai_compatible would feed the decoder an empty string (also an error); openai/gemini already skipped

All streaming wire snapshots pass unmodified through the shared loop, and the framer has direct unit tests for these cases.

## Behavior notes (beyond the framing edge cases)

- **Error values are byte-identical**: `Codec::decode_error`'s default is exactly the `parse_error_body("type")` + `error_from_status_code` path the deleted call sites inlined; gemini's gRPC-aware override is what its paths already used.
- **Logging only**: gemini's blocking paths gain the shared timeout/error/status warn logs (they had none); count-tokens requests are uniformly tagged `operation="input_token_count"` (previously only openai's was). The openai count-tokens logging pin passes unchanged.
- gemini's timeout error message now uses the configured provider name instead of a hardcoded `gemini:` prefix (visible only on custom-named gemini routes).

## Verification

- `cargo nextest run --workspace`: green except the 5 pre-existing fabro-workflow failures that fail identically on main (environment-dependent, unrelated)
- fabro-llm: 545 passed — all PR 0 wire snapshots unmodified
- fabro-cli compiles against the frozen `providers::common::{LineReader, parse_retry_after}` paths

Next in the series: PR 7 (codec on the route in fabro-model) — route vocabulary + the route-equivalence table test.


swerner:
Ran a `/simplify` pass over the PR diff (4 parallel review agents: reuse, simplification, efficiency, altitude). Net result: the consolidation held up well — one real cleanup out of seven raw findings.

**Applied** (32f810c37): removed a redundant inner `{ ... }` block in `fabro_server.rs`'s stream loop, left over when the old nested `if event_type == "stream_event"` check became a pattern match, and consolidated the stale skip-comments it stranded.

**Findings reviewed and skipped, with reasoning:**

- *Reuse*: clean — no existing helpers duplicated; `HttpApi` → `HttpTransport` and the `providers::common` moves are pure relocations.
- *Efficiency*: clean — no new allocations on the SSE hot path. The `Cow`-borrowing `parse_sse_block` is actually an improvement over the old per-adapter `.to_string()` copies. The full `.trim()` in DataLines framing (vs gemini's old `trim_end_matches('\r')`) is required for the unified leading-whitespace behavior, not waste.
- *Altitude: "transport can't pass an error-field name to `decode_error`, breaking gemini"* — false positive; gemini overrides `Codec::decode_error`, and the transport delegating to the trait method is exactly the seam this PR builds.
- *Altitude: "DataLines parsing should be codec-owned"* — the trim/skip/empty rules are SSE-spec-level, not dialect-specific; the unified framing semantics are a documented, deliberate decision in this PR.
- *Altitude: "`parse_sse_block` export encodes adapter filtering into transport"* — backwards; the `stream_event` filter lives in the fabro_server adapter, which is the right layer.
- *Altitude: "move `extract_system_prompt` into a codec helpers submodule"* — relocation churn with no maintenance payoff.

Verification after the fix: 545 fabro-llm tests pass, clippy `-D warnings` and pinned-nightly fmt clean.

PR #492, opened 2026-06-11
Demote non-interpolating config fields to plain String

swerner:
# Demote non-category leak fields to plain `String`

Second slice of the interpolation unification — the first **reducing** PR
stacked on the foundation (#472), per the reduce-first sequencing: narrowing
changes land before capability additions. (The other reducing slice, the DOT
de-templating, already landed independently as #474.)

## Why

The target model gives `InterpString` to fields in five categories —
`command` / `script` / `headers` / `env` / `url` — wherever they appear. A
handful of fields were typed `InterpString` but are *identifiers or commit
content*, not in any category:

- `run.model.provider` / `run.model.name`
- `cli.exec.model.provider` / `cli.exec.model.name`
- `run.git.author.name` / `run.git.author.email`
- `run.scm.owner` / `run.scm.repository`

Their consumers never resolved them — they leaked raw source text via
`as_source()`. This PR demotes them to plain `String` (layer and resolved
structs) with **no interpolation**.

## The principle: only `InterpString` fields access variables

These fields are dropped from the variable substitute pass entirely, so both
`{{ vars.* }}` and `{{ env.* }}` are now literal text. This **removes an
incidental behavior**: run-scoped plain-`String` fields used to get
`{{ vars.* }}` substituted via the String pass (a lucky accident), while
`env` always leaked literally. Variable access becomes deliberate and typed
rather than accidental; if any of these fields should support variables
later, that's a controlled promotion back to `InterpString`.

## Behavior changes (honest list)

- **The incidental run-scoped `{{ vars.* }}` substitution on these eight
  fields stops working.** To keep the removal visible rather than silent, a
  `tracing::warn!` fires at resolve time when a demoted field still contains
  claimed template tokens (`warn_if_demoted_template`). Unclaimed `{{ ... }}`
  text (jq programs, Go templates) never interpolated and does not warn.
- `{{ env.* }}` / `{{ secrets.* }}` / `{{ inputs.* }}` never resolved on
  these fields, so nothing else changes.

## Added in review: D11 demotions (separate commit, revertable)

The rule got refined during review: a field is `InterpString` iff it is in one
of the five categories **and resolved at the run boundary** (the only point
where `vars`/`secrets`/`inputs` exist — they're server state, so connect-time
and startup-time fields can't reach them even in principle). A separate commit
applies the clean subset so it can be cherry-picked out if we change course:

- `cli.target.http.url` / `cli.target.unix.path` — consumed at CLI connect
  time; consumers only ever leaked raw source, so nothing working is removed.
- `run.working_dir` — **the outlier; see the PR comment.** Its `{{ vars.* }}`
  substitution worked; demoted on the category test alone.

## What's deliberately NOT here

- The **control-plane fields** (`server.storage.root` / `listen.unix.path` /
  S3 fields / `github.app_id/client_id/slug` / `server.api.url` /
  `server.web.url`) — untouched here, demoted in a follow-up PR.
  **Resolved during review** (see the resolution comment): `InterpString` was
  conflating the user-facing workflow language with the internal control
  plane. Control-plane fields never interpolate; the few deployment knobs
  that need late binding (e.g. `FABRO_WEB_URL`, whose only real usage is the
  split-web PoC ferrying a compose env var across a file mount) become
  explicit native `EnvVars` reads, and `fabro-server/src/interp.rs` shrinks
  to deletion. `slack.default_channel` stays `InterpString` (consumed with
  run context).

## Implementation notes

- Consumers move from `as_source()` to direct `String` access; the
  foundation's `#[expect(disallowed_methods, ... demotion pending ...)]`
  annotations for these fields are removed (no longer `InterpString`).
- `fabro-checkpoint`'s author plumbing and `fabro-manifest`'s scm fields
  simplify accordingly.

## Verification

- `cargo build --workspace`
- `cargo nextest run --workspace` → 6684 passed, 181 skipped

### DIFF
--- a/lib/crates/fabro-checkpoint/src/author.rs
+++ b/lib/crates/fabro-checkpoint/src/author.rs
@@ -1,7 +1,6 @@
 use std::fmt::Write;
 
 use fabro_config::GitAuthorLayer;
-use fabro_types::settings::InterpString;
 use fabro_types::settings::run::GitAuthorSettings;
 
 /// Resolved git author identity for checkpoint commits.
@@ -51,30 +50,14 @@ impl GitAuthor {
     }
 }
 
-#[expect(
-    clippy::disallowed_methods,
-    reason = "raw source is today's behavior; run.git.author.* is slated for demotion to plain String in the \
-              interpolation unification (D2)"
-)]
 impl From<&GitAuthorLayer> for GitAuthor {
     fn from(value: &GitAuthorLayer) -> Self {
-        Self::from_options(
-            value.name.as_ref().map(InterpString::as_source),
-            value.email.as_ref().map(InterpString::as_source),
-        )
+        Self::from_options(value.name.clone(), value.email.clone())
     }
 }
 
-#[expect(
-    clippy::disallowed_methods,
-    reason = "raw source is today's behavior; run.git.author.* is slated for demotion to plain String in the \
-              interpolation unification (D2)"
-)]
 impl From<&GitAuthorSettings> for GitAuthor {
     fn from(value: &GitAuthorSettings) -> Self {
-        Self::from_options(
-            value.name.as_ref().map(InterpString::as_source),
-            value.email.as_ref().map(InterpString::as_source),
-        )
+        Self::from_options(value.name.clone(), value.email.clone())
     }
 }
--- a/lib/crates/fabro-cli/src/commands/exec.rs
+++ b/lib/crates/fabro-cli/src/commands/exec.rs
@@ -282,8 +282,8 @@ pub(crate) async fn execute(mut args: ExecArgs, ctx: &CommandContext) -> AnyResu
     let cli = &ctx.user_settings().cli;
     #[cfg(feature = "sleep_inhibitor")]
     let _sleep_guard = sleep_inhibitor::guard(cli.exec.prevent_idle_sleep);
-    let provider_str = cli.exec.model.provider.clone();
-    let model_str = cli.exec.model.name.clone();
+    let provider_str = cli.exec.model.provider.as_deref();
+    let model_str = cli.exec.model.name.as_deref();
     let permissions = cli.exec.agent.permissions.map(|p| match p {
         AgentPermissions::ReadOnly => AgentPermissionLevel::ReadOnly,
         AgentPermissions::ReadWrite => AgentPermissionLevel::ReadWrite,
@@ -293,12 +293,8 @@ pub(crate) async fn execute(mut args: ExecArgs, ctx: &CommandContext) -> AnyResu
         SettingsOutputFormat::Text => OutputFormat::Text,
         SettingsOutputFormat::Json => OutputFormat::Json,
     });
-    args.agent.apply_cli_defaults(
-        provider_str.as_deref(),
-        model_str.as_deref(),
-        permissions,
-        output_format,
-    );
+    args.agent
+        .apply_cli_defaults(provider_str, model_str, permissions, output_format);
     let server_target = user_config::exec_server_target(&args.server)?;
     // v2 MCPs live under `cli.exec.agent.mcps` (owner-specific) or
     // `run.agent.mcps`. For `fabro exec` we use the cli.exec path, falling
--- a/lib/crates/fabro-cli/src/commands/exec.rs
+++ b/lib/crates/fabro-cli/src/commands/exec.rs
@@ -16,7 +16,6 @@ use fabro_llm::types::{
 };
 use fabro_mcp::config::McpServerSettings;
 use fabro_model::ProviderId;
-use fabro_types::settings::InterpString;
 use fabro_types::settings::cli::OutputFormat as SettingsOutputFormat;
 use fabro_util::exit::{self, ErrorExt, ExitClass};
 use futures::stream;
@@ -276,11 +275,6 @@ impl ProviderAdapter for AuthenticatedFabroServerAdapter {
     }
 }
 
-#[expect(
-    clippy::disallowed_methods,
-    reason = "raw source is today's behavior; cli.exec.model.* is slated for demotion to plain String in the \
-              interpolation unification (D2)"
-)]
 pub(crate) async fn execute(mut args: ExecArgs, ctx: &CommandContext) -> AnyResult<()> {
     use fabro_agent::cli::PermissionLevel as AgentPermissionLevel;
     use fabro_types::settings::run::AgentPermissions;
@@ -288,13 +282,8 @@ pub(crate) async fn execute(mut args: ExecArgs, ctx: &CommandContext) -> AnyResu
     let cli = &ctx.user_settings().cli;
     #[cfg(feature = "sleep_inhibitor")]
     let _sleep_guard = sleep_inhibitor::guard(cli.exec.prevent_idle_sleep);
-    let provider_str = cli
-        .exec
-        .model
-        .provider
-        .as_ref()
-        .map(InterpString::as_source);
-    let model_str = cli.exec.model.name.as_ref().map(InterpString::as_source);
+    let provider_str = cli.exec.model.provider.clone();
+    let model_str = cli.exec.model.name.clone();
     let permissions = cli.exec.agent.permissions.map(|p| match p {
         AgentPermissions::ReadOnly => AgentPermissionLevel::ReadOnly,
         AgentPermissions::ReadWrite => AgentPermissionLevel::ReadWrite,
--- a/lib/crates/fabro-cli/src/user_config.rs
+++ b/lib/crates/fabro-cli/src/user_config.rs
@@ -11,7 +11,7 @@ use fabro_config::{
 use fabro_static::EnvVars;
 use fabro_types::settings::cli::CliTargetSettings;
 use fabro_types::settings::server::LogDestination;
-use fabro_types::settings::{CliNamespace, InterpString, RunNamespace};
+use fabro_types::settings::{InterpString, RunNamespace};
 use fabro_types::{ServerSettings, UserSettings};
 use fabro_util::error::SharedError;
 use fabro_util::version::FABRO_VERSION;
@@ -220,21 +220,12 @@ fn value_at_path<'a>(document: &'a toml::Value, path: &[&str]) -> Option<&'a tom
     Some(current)
 }
 
-/// Pull the resolved CLI target configuration out of `[cli.target]`.
-/// Returns either an http(s) URL or a unix socket path.
-fn cli_target_from_settings(settings: &CliNamespace) -> Option<String> {
-    let target = settings.target.as_ref()?;
-    match target {
-        CliTargetSettings::Http { url } => Some(url.clone()),
-        CliTargetSettings::Unix { path } => Some(path.clone()),
-    }
-}
-
 fn configured_server_target(settings: &UserSettings) -> Result<Option<ServerTarget>> {
-    let Some(value) = cli_target_from_settings(&settings.cli) else {
-        return Ok(None);
-    };
-    parse_server_target(&value).map(Some)
+    match settings.cli.target.as_ref() {
+        Some(CliTargetSettings::Http { url }) => ServerTarget::http_url(url).map(Some),
+        Some(CliTargetSettings::Unix { path }) => ServerTarget::unix_socket_path(path).map(Some),
+        None => Ok(None),
+    }
 }
 
 pub(crate) fn default_server_target() -> ServerTarget {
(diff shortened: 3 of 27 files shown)


swerner:
## Heads-up for the follow-up: `run.working_dir` is the outlier in the next demotion batch

The refined rule (`InterpString` if *five categories* **and** *resolved at the run boundary*) sweeps `cli.target.*`, `server.storage.root`, `listen.unix.path`, and the S3 endpoints into a follow-up demotion PR. For all of those, demotion removes nothing real: they're consumed at CLI connect time or server startup, where `vars`/`secrets`/`inputs` are unreachable *in principle* (the var store and vault are server state — and the var store literally lives inside the storage that `storage.root` locates).

**`run.working_dir` is different, and worth calling out honestly:**

- It **does** cross the run boundary — it's substituted in `RunNamespace::substitute_variables` deliberately (`substitute_option(&mut self.working_dir, ...)`), so `{{ vars.* }}` **works on it today**.
- Its demotion rests entirely on the **category test**: it's a path/identifier of *where* to run, not command/script/headers/env/url *content*.
- So unlike every other field in both batches, this is a real capability removal — a product call, not an impossibility.

Example usages that stop working when it demotes:

```toml
# monorepo: per-run service directory via a server-side variable
[run]
working_dir = "{{ vars.SERVICE_DIR }}"
```

```toml
# shared workflow, repo subpath varies by team
[run]
working_dir = "services/{{ vars.TEAM }}/api"
```

(`{{ env.* }}` in `working_dir` never resolved, so only the `vars` forms are affected.)

Mitigations: the demotion ships the same `warn_if_demoted_template` resolve-time warning this PR introduces, so existing configs fail loudly rather than silently running in a literal `{{ vars.* }}` directory; and promotion back to `InterpString` later is a controlled, typed decision if the monorepo use case turns out to be load-bearing.

swerner:
## Ran `/simplify` on this diff — cleanups applied in a3a70a151

Four parallel review agents (reuse / simplification / efficiency / altitude) over the PR diff. Findings applied:

- **Borrows over clones** where values are only consumed as `&str`: `cli.exec.model.*` in `exec.rs`, `scm.owner`/`repository` in `configured_repo_origin_url`, `provider` in `resolve_start_llm`. `materialize_run` now `take()`s the model fields it unconditionally overwrites instead of cloning.
- **`resolve_git` flattened** — the two `warn_if_demoted_template` calls moved out of the `and_then(map(...))` closure into a plain `if let`, so the struct-building chain is side-effect-free again.
- **Comment accuracy** (the altitude finding): the comments in `substitute_variables`, the `warn_if_demoted_template` doc, and the test lead asserted "only `InterpString` fields interpolate" in *present* tense — false today, since a dozen plain-`String` fields (e.g. `model.controls.*`, `scm.provider`) still substitute via the String pass. Reworded to state the actual rule: these fields were removed from the vars pass (D2); other String fields keep substituting until the String pass is retired in a later slice.

Noted but deliberately skipped: merging the now-byte-identical `GitAuthorLayer`/`GitAuthorSettings` `From` impls (type unification is out of scope for this slice), and table-driving the four `warn_if_demoted_template` call sites (temporary shim, all sites greppable via the helper name).

Reuse review came back clean — `warn_if_demoted_template` correctly delegates to `InterpString::parse(..).is_literal()` rather than re-implementing token detection.

Verification: clippy `-D warnings` clean, fmt clean, touched-crate tests 2674 passed (5 failures are pre-existing local-env credential leakage, identical on the base commit; CI green).

The upcoming non-run-field demotions (`cli.target.*`, `working_dir`, server paths) will be a separate reducing PR — we'll run `/simplify` on that diff independently rather than re-running here.

swerner:
## D11 demotions: what landed, what's paused, and the iffy cases

Commit `1afa06a4a` (kept separate so it's trivially revertable / movable to its own PR) demotes `cli.target.http.url`, `cli.target.unix.path`, and `run.working_dir` under the refined rule: **`InterpString` iff in one of the five categories AND resolved at the run boundary** — the run boundary being the only point where `vars` (server `VariableStore`), `secrets` (server vault), and `inputs` (run request) all exist. For `cli.target.*` it's circular by construction: resolving `{{ vars.SERVER_URL }}` would mean querying the server whose address is the value being resolved. Their consumers only ever leaked raw source, so nothing working was removed.

### The iffy cases, honestly

**1. `run.working_dir` (demoted here — the one real capability removal).**
It *does* cross the run boundary, and its `{{ vars.* }}` substitution genuinely worked. Demotion rests on the category test alone — it's a path, not command/script/headers/env/url content. Configs like these stop interpolating (and now warn loudly at resolve time):

```toml
[run]
working_dir = "{{ vars.SERVICE_DIR }}"          # monorepo, per-run service dir
working_dir = "services/{{ vars.TEAM }}/api"    # shared workflow, per-team subpath
```

Promotion back is a controlled, typed decision if this turns out to be load-bearing.

**2. `server.api.url` / `server.web.url` (NOT demoted — paused).**
Implementation surfaced that this isn't a "leak" at all: server-scope `{{ env.* }}` resolution **works today** via `fabro-server/src/interp.rs`, and it's load-bearing:
- `docker/split-web/settings.toml` ships `url = "{{ env.FABRO_WEB_URL }}"` — the packaged split-web deployment depends on it.
- The `server.web.url` validation error literally instructs operators to use "the FABRO_WEB_URL environment variable" (`canonical_origin.rs`).
- Containers can't substitute env into mounted TOML files without entrypoint gymnastics, so `{{ env.* }}`-in-server-config earns its keep in Docker deployments — the "deploy tooling covers env natively" argument is weakest exactly here.

**3. `github.app_id` / `client_id` / `slug` (NOT demoted — paused).**
Mixed consumption: in-run (clone-token minting, PR creation) AND no-run-context (OAuth login resolves `client_id` per request in `web_auth.rs`, webhook/diagnostics paths). Under a strict run-boundary rule they'd demote — but their env resolution also works today through the same `interp.rs` machinery, and the OAuth flow actively uses it.

### The open question this raises

The original D11 cost-benefit assumed we'd be *avoiding building* startup-time resolution machinery. We wouldn't — it already exists (built in the foundation PR as the rogue-resolver consolidation), it's small, audited, and shipped artifacts use it. So the real choice for server-scope fields is:

- **(a) Demote anyway** — strictest rule, but deletes working `{{ env.* }}` support and breaks `docker/split-web` + the documented FABRO_WEB_URL pattern (fixable with entrypoint envsubst, but that's moving complexity, not removing it), or
- **(b) Keep server scope as a second sanctioned resolution point** — rule becomes "InterpString iff resolved at the run boundary **or** by the server's own env at startup/request time (env+secrets only)" — which is the original D1, now with a mechanical justification.

Storage/listen/S3 fields ride on the same ruling, so they're paused with it.

swerner:
## Resolution: the (a)/(b) question dissolves into (c)

Closing the loop on the open ruling from the previous comment. Digging into how `FABRO_WEB_URL` is actually consumed settled it.

### The finding: one type, two jobs, two env channels

`InterpString` was conflating two unrelated jobs:

1. **The user-facing workflow config language** — commands/scripts/headers/env/urls that execute as part of a run, resolved at the run boundary where `vars`/`secrets`/`inputs` exist.
2. **The internal control plane** — deployment knobs (CLI connection, server identity/storage/listen, GitHub App identifiers) consumed at connect/startup/request time, where workflow variables can't exist.

Job 2 leaked into job 1's type because `{{ env.* }}` tokens were the only late-binding mechanism in a TOML file. The tell: Fabro **already runs a native env channel** for exactly this job — `fabro-static::EnvVars` with direct reads (`FABRO_STORAGE_DIR` is a clap env arg, `FABRO_SERVER_MAX_CONCURRENT_RUNS` / `FABRO_LOG*` read directly, and the GitHub App *secrets* come from env via the secret registry). Meanwhile `FABRO_WEB_URL` is declared in the same registry but **never read directly anywhere** — it only works if a `{{ env.FABRO_WEB_URL }}` token happens to sit in your settings file. Two channels, inconsistently applied; the GitHub App config is split across both (secrets native, identifiers token-mediated). The `canonical_origin.rs` error message advertises FABRO_WEB_URL as if it were native — it's silently false for any user whose settings lack the token.

### FABRO_WEB_URL's entire footprint

One internal PoC. `docker-compose.split-web.yaml:19` sets it — **using compose's own native `${SPLIT_WEB_PORT:-8080}` substitution on the same line** — and `docker/split-web/settings.toml:5` carries the token whose only job is ferrying that value across a read-only file mount into the process. No user docs, no install flow, nothing else. The shipped dependency is on the *env var*, not the *token mechanism*.

### The resolution (c)

- **`InterpString` is the user-facing workflow language, full stop**: five categories, resolved at the run boundary, all four namespaces. Everything a user can template behaves identically.
- **The control plane never interpolates**: plain `String` in TOML, written literally. Deployment knobs that need late binding get explicit native `EnvVars` reads (env override > settings literal > default) — initially just `FABRO_WEB_URL`; others promoted on demand.
- `github.app_id/client_id/slug` demote — the mixed in-run/no-run consumption dilemma dissolves (plain strings need no resolution point), and the identifiers finally match their sibling secrets on the native channel.
- `slack.default_channel` is the one server-defined survivor (consumed with run context). `llm.providers.base_url` stays a D9 promotion candidate.
- `docker/split-web/settings.toml` drops its token (the compose env var keeps working — natively), the error message becomes true for everyone, and `fabro-server/src/interp.rs` shrinks to deletion.

This lands as a follow-up control-plane demotion PR, not here. It also answers both earlier review comments at the root: CLI targeting stays impossible-by-construction with credentials on their one audited path, and "complexity breeds vulns" cuts the right way — today's two-channel state, where which-mechanism-works depends on the contents of a mounted TOML file, was the complex one. End state: one interpolation mechanism at one boundary, plus one greppable allowlist of native env knobs.

(Plan doc updated with the full writeup — D11 resolution (c), 2026-06-11.)

brynary:
@swerner is this one good to go?

swerner:
## 🧹 Ran `/simplify` (Opus 4.8)

Quality-only pass over the demotion diff (reuse / simplification / efficiency / altitude) — no behavior changes.

**Fixed** (`fabro-config/src/resolve/cli.rs`):

1. **`cli.target` warn labels now use the real config paths.** `warn_if_demoted_template` was passing `cli.target.http.url` / `cli.target.unix.path`, but `http`/`unix` is the serde tag *value*, not a TOML path segment — the actual config path is `cli.target.url` / `cli.target.path` (which is also what the adjacent `require_string` error paths already use, and matches the convention for the other demoted-field warnings like `run.model.provider` / `cli.exec.model.*`). Aligned them so one field doesn't surface under two different names across warn vs. error output.
2. **Reused the `model` binding** in `resolve_exec` instead of re-deriving `exec.model.as_ref()` twice when building `CliExecModelSettings`.

**Skipped** (with reasons):

- *Collapse `require_string` / `require_interp` into a generic `require<T: Clone + Default>`* — `InterpString` doesn't derive `Default`, so this would mean adding a new trait impl to a type outside this diff just to dedup a 12-line near-duplicate. Cost > benefit.
- *Add a tracing-capture test asserting the demotion warning actually fires* — a genuine coverage gap, but it's net-new test infrastructure rather than a simplification of the diff. Better as a follow-up / `/code-review` item if we want the warn contract pinned executably.

Efficiency pass came back clean — the demotion *removes* `InterpString` wrapper allocations on the read paths.

**Verification:** `cargo nextest run --workspace` → 6684 passed, 181 skipped · `cargo +nightly clippy -p fabro-config --all-targets -- -D warnings` → clean.

swerner:
Ran a simplify pass on PR 492 with three parallel review agents: code reuse, code quality, and efficiency.

Findings addressed in commit `971a7945f`:

- Reuse: factored duplicated `require_interp` / `require_string` missing-value handling behind a shared helper.
- Reuse/quality: removed duplicate preflight model/provider resolution and now use the already-materialized run settings.
- Quality: removed the `[cli.target]` typed-setting to `String` to `ServerTarget` round trip.
- Efficiency: added cheap guards before parsing demoted strings for migration warnings.
- Efficiency: avoided cloning unchanged plain-string settings on false-positive `{{ vars.` scans by using a borrowed `Cow` path.

Findings intentionally not changed:

- `ProviderId` / `ModelId` for demoted model fields: left as plain `String` because the interpolation-unification D11 plan explicitly scopes PR 492 as a reducing demotion slice with these fields as literal strings.
- Structured resolver warnings: left as `tracing::warn!` because that is the migration-warning shape documented for this slice.

Validation run:

- `cargo check -p fabro-types -p fabro-config -p fabro-workflow -p fabro-checkpoint -p fabro-cli -p fabro-manifest -p fabro-server --all-targets`
- `cargo +nightly-2026-04-14 clippy -p fabro-types -p fabro-config -p fabro-cli -p fabro-server --all-targets -- -D warnings`
- `cargo nextest run -p fabro-types -p fabro-config`
- `cargo nextest run -p fabro-workflow -- materialize_run`
- `cargo nextest run -p fabro-server -- preflight_resolves_model_aliases_from_app_state_catalog`
- `cargo nextest run -p fabro-server -- preflight_probes_configured_llm_model_availability`
- `cargo nextest run -p fabro-cli -- user_config`

PR #493, opened 2026-06-11
refactor(model): put the codec on the route

swerner:
PR 7 of the gateway refactor series (after #481, #488, #487, #489, #491) — the series capstone: the wire dialect becomes route vocabulary in fabro-model config instead of a structural implication of the adapter type.

## What's here

**`fabro-model/src/codec.rs` (new)** — `CodecKind` (`anthropic_messages`, `openai_responses`, `openai_compatible`, `gemini_generate`; strum per house style). `CodecKind::default_for(AdapterKind)` reproduces the historical adapter→dialect fusion exactly.

**Catalog schema** — optional `codec` on provider rows and model rows (the multiplexer case), sparse-merged with the existing `.or()` pattern. Omitted everywhere in the built-in catalog, so **all defaults reproduce today's routes**. Explicit pairings outside the adapter's default are rejected at catalog build (`UnsupportedProviderCodec` / `UnsupportedModelCodec`) so no new route combination is silently enabled by configuration — the field is vocabulary for the OpenRouter/Bedrock feature PRs, not a new capability. `Catalog::effective_codec` mirrors `effective_agent_profile`. fabro-config mirrors the field through `LlmLayer` (`ProviderSettings.codec`, `ModelSettings.codec`) and the catalog-settings conversion.

**Route resolution** — `adapter_registry::resolve_route(catalog, model)` assembles `(provider row, model row)` into `Route { provider, transport, codec, deployment_id, billing_policy, agent_profile }`.

**Route-equivalence table test** — every built-in model row pinned to its resolved tuple as an executable table (23 rows), with a coverage assert so a new built-in model can't land without a deliberate table edit. This is the "compat mapping as an executable table, not a comment" test from the plan.

**`AdapterConfig` cleanup** — the OpenAI-only fields (`codex_mode`, `org_id`, `project_id`) move out of the shared struct into `AdapterKindOptions::OpenAi(OpenAiAdapterOptions)`; the client populates them only for OpenAi-kind routes, which is the only factory that ever read them.

## Deliberate scope cuts

- **No per-model `billing_policy`** — that schema change exists solely for the OpenRouter redo, which owns it.
- **`codec_params` and `supports_count_tokens` stay adapter-internal** — the registry `Route` carries what the catalog defines; the per-route knobs in the adapters' `RouteConfig` move out when a second codec/transport pairing actually exists (OpenRouter's anthropic skin / Bedrock). Wiring `resolve_route` into `Client` request dispatch is the optional PR 8 and is likewise deferred.
- **No user-facing docs for `codec`** — every accepted value equals the default, so there is nothing actionable to document yet; docs land with the first feature PR that enables a non-default pairing.

## Verification

- `cargo nextest run --workspace --no-fail-fast` (re-run post-rebase onto #491's merge): 6701 passed; the only failures are the same 5 pre-existing environment-dependent fabro-workflow failures noted in #491, identical on main
- fabro-llm: 548 passed — all wire snapshots unmodified

This ends the refactor series: the seams exist. Next up are a standalone cost PR (`cost.rs` + `Response.cost_usd`/`CostSource`, pulled forward from the #438 triage as its own pre-OpenRouter step) and then the feature redos — OpenRouter (#438: one TOML + typed codec params) and Bedrock (#459: sigv4/eventstream transport + config, private codec layer deleted).

### DIFF
--- a/lib/crates/fabro-config/src/builders.rs
+++ b/lib/crates/fabro-config/src/builders.rs
@@ -330,6 +330,7 @@ fn provider_settings_to_catalog(
     model_catalog::ProviderCatalogSettings {
         display_name:   settings.display_name,
         adapter:        settings.adapter,
+        codec:          settings.codec,
         agent_profile:  settings.agent_profile,
         auth:           settings.auth,
         billing_policy: settings.billing_policy,
@@ -346,6 +347,7 @@ fn model_settings_to_catalog(settings: ModelSettings) -> model_catalog::ModelCat
     let ModelSettings {
         provider,
         api_id,
+        codec,
         agent_profile,
         display_name,
         family,
@@ -365,6 +367,7 @@ fn model_settings_to_catalog(settings: ModelSettings) -> model_catalog::ModelCat
     model_catalog::ModelCatalogSettings {
         provider,
         api_id,
+        codec,
         agent_profile,
         display_name,
         family,
--- a/lib/crates/fabro-config/src/layers/combine.rs
+++ b/lib/crates/fabro-config/src/layers/combine.rs
@@ -1,6 +1,6 @@
 use std::collections::{BTreeMap, HashMap};
 
-use fabro_model::{AgentProfileKind, BillingPolicy, ProviderAuthConfig};
+use fabro_model::{AgentProfileKind, BillingPolicy, CodecKind, ProviderAuthConfig};
 use fabro_types::settings::cli::{CliAuthStrategy, OutputFormat, OutputVerbosity};
 use fabro_types::settings::run::{
     AgentPermissions, ApprovalMode, EnvironmentNetworkMode, EnvironmentProvider, MergeStrategy,
@@ -93,6 +93,7 @@ impl_combine_or_option!(
     LogFilter,
     AgentProfileKind,
     BillingPolicy,
+    CodecKind,
     ProviderAuthConfig,
     ReasoningEffortFeature,
 );
--- a/lib/crates/fabro-config/src/layers/llm.rs
+++ b/lib/crates/fabro-config/src/layers/llm.rs
@@ -29,7 +29,7 @@
 use std::collections::{BTreeMap, HashMap};
 
 use fabro_model::catalog::deserialize_knowledge_cutoff;
-use fabro_model::{AgentProfileKind, BillingPolicy, ProviderAuthConfig};
+use fabro_model::{AgentProfileKind, BillingPolicy, CodecKind, ProviderAuthConfig};
 pub use fabro_model::{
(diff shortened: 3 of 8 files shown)


swerner:
Ran `/simplify` on this PR (4 review agents: reuse, simplification, efficiency, altitude). The diff came back largely clean — one fix applied, the remaining findings didn't survive verification.

## Fixed

**Test boilerplate in `lib/crates/fabro-llm/src/adapter_registry.rs`** — four tests built full `AdapterConfig` struct literals, each re-spelling `kind_options: AdapterKindOptions::None` and `catalog: None` by hand. They now use struct-update syntax over the existing `AdapterConfig::new()` constructor, so only the fields each test actually exercises (`base_url`, `extra_headers`) are spelled out. Verified: fmt clean, clippy `-D warnings` clean, all 10 `adapter_registry` tests pass.

## Skipped, with reasons

- **Remove the `AsRef<str>` impl on `CodecKind` as unused** — false positive. `AdapterKind` in the same crate has the identical `as_str()` + `AsRef<str>` pattern (`adapter.rs:39-46`); `CodecKind` is deliberately symmetric with its sibling enum, and the house strum guidance endorses the `as_str` wrapper.
- **Consolidate `resolve_provider_codec`/`resolve_model_codec`** — the shared core (`CodecKind::default_for`) is already extracted; what remains differs in error variant and in the model function's `None => Ok(provider.codec)` inheritance arm, which documents intent for when the only-default-pairing constraint relaxes. A generic helper would trade the small duplicated match for `map_err` boilerplate in both callers.
- **Drop the fabro-config TOML parse tests as subsumed by catalog tests** — false positive. They exercise serde on fabro-config's `ProviderSettings`/`ModelSettings`, a different layer than fabro-model's catalog settings, and they mirror the existing `model_agent_profile_parses_from_toml` pattern next to them.
- **Restructure the 23-row route-equivalence table by provider grouping** — the flat one-row-per-model table is this PR's stated design ("compat mapping as an executable table"); grouping would weaken the per-row pinning that makes a reroute show up as a deliberate table edit.
- **Double model-index lookup in `resolve_route`** (`model_settings()` calls `get()` internally) — real but negligible: two HashMap hits in config-time code, and avoiding it would require a new `Catalog` accessor in fabro-model for no measurable gain.
- **`CodecKind` enum/round-trip test mirrors `AdapterKind`** — house style per the strum guidance; per-enum round-trip tests are the established convention here.

Altitude review found no layering issues: validation lives at catalog build where the decision is made, `resolve_route` consumes already-resolved values, and the `AdapterKindOptions` split keeps OpenAI-only knobs out of shared infrastructure.

PR #494, opened 2026-06-11
feat(llm): catalog-estimated completion cost on Response

swerner:
Standalone pre-OpenRouter step, pulled forward from the #438 triage (the gateway-refactor plan's "additive feature PR alongside the redo"): completion responses carry a USD cost with provenance.

## What's here

**`Response.cost_usd` + `Response.cost_source`** — new optional fields (`skip_serializing_if` keeps the wire shape byte-identical when unset). `CostSource` (`authoritative` | `estimated`) lives in fabro-model's billing vocabulary next to `UsdMicros`/`TokenCounts`, since the API layer reuses it.

**`fabro-llm/src/cost.rs`** — `estimate_cost_usd`, a thin wrapper over the existing `Catalog::price_tokens` billing machinery (billing-policy- and speed-aware), ported from #438's prototype with attribution. One fix over the prototype: model aliases and provider names are canonicalized before building the `ModelRef` — `ModelPricing::bill` rejects non-canonical refs, so the original would silently skip cost on alias requests (caught by a new test).

**Client-level stamping** — one generic post-decode site instead of #438's ~8 per-adapter sites (which predate the codec refactor): `Client::complete` stamps blocking responses and `Client::stream` stamps `Finish` events, beneath the middleware chain so middleware observes final responses. Codecs stay wire-translation-only — zero wire-snapshot churn — and every registered adapter (including custom `register_provider` ones) gets the same treatment. Stamping never overwrites an existing cost, so future authoritative in-band costs (OpenRouter) take precedence by construction.

**API surface** — `cost_usd`/`cost_source` on `CompletionResponse` (OpenAPI spec + handler + regenerated TS client). The streaming endpoint already carries cost implicitly since `Finish` events serialize the `Response` verbatim; this makes the blocking surface match. `CostSource` reuses the canonical fabro-model type via `with_replacement`, with the standard round-trip test pinning type identity and JSON parity.

## Deliberately not here (stays with the OpenRouter redo per the plan's hard rule)

- Authoritative `usage.cost` parsing in the `openai_compatible` codec wire structs
- Cached-token usage parsing (changes observable usage values)
- Per-model `billing_policy` schema field

## Verification

- `cargo nextest run --workspace --no-fail-fast`: 6701 passed; only the known 5 pre-existing environment-dependent fabro-workflow failures (identical on main)
- All fabro-llm wire snapshots unmodified; new pins: cost estimation unit tests (incl. alias canonicalization), Client stamping tests (blocking, streaming, beneath middleware, no-catalog), fabro-api `CostSource` round-trip
- clippy `-D warnings` + pinned-nightly fmt clean; `bun run typecheck` clean in fabro-web

Independent of the route-vocabulary work in #493 — branches directly off main. After both land, the OpenRouter redo shrinks to config + typed codec params + authoritative-cost decode.


swerner:
Ran `/simplify` over this PR (four parallel review agents — reuse, simplification, efficiency, altitude — findings deduped and applied). Two commits pushed:

## `refactor(llm): single-site cost stamping and simplify the estimate seam`

- **`estimate_cost_usd` returns `Option<f64>`** instead of `(Option<f64>, Option<CostSource>)` — the second element was always derivable from the first, and the tuple made the illegal `(Some, None)` state representable. `apply_estimated_cost` now derives the source.
- **`cost` module is `pub(crate)`** — the client is its only caller in the workspace, so no public API surface to keep stable.
- **Stamping is single-sited per mode.** It previously existed at 4 sites (empty-middleware fast path + chain base, in both `complete` and `stream`); shared `complete_stamped`/`stream_stamped` helpers now serve both paths, so future stamping changes (e.g. merging authoritative provider data for OpenRouter) happen in one place per mode. `stamp_stream_costs` also mutates the `Finish` response in place instead of destructuring and rebuilding the variant.
- **Test fixtures deduped**: three near-identical ~30-line TOML catalogs in `cost.rs` collapsed into one parameterized fixture, plus a new test pinning that no estimate leaves `cost_source` unset.

## `feat(llm): carry server-stamped costs through the fabro-server adapters`

Altitude finding with a small behavior change: the PR added `cost_usd`/`cost_source` to the server's `CompletionResponse` wire format, but both in-repo consumers of that format (the `fabro_server` provider adapter and the CLI exec adapter) hardcoded `None` — discarding the server's value and re-estimating from whatever catalog the local client has, or losing the cost entirely without one. They now parse and carry the fields; client-side stamping already never overwrites an existing cost, so the server value takes precedence by construction. Pinned in `complete_parses_response`.

## Flagged but deliberately skipped

- **Alias pre-canonicalization in `cost.rs`** — one agent flagged it as redundant with `price_tokens`' internal resolution; verified false positive (`ModelPricing::bill` rejects non-canonical `ModelRef`s, and the alias test pins it).
- **`Response` constructor / defaulted metadata tail** — this PR hand-edited ~45 struct-literal sites across 6 crates to add two `None` fields; a constructor or `#[derive(Default)]` metadata sub-struct would absorb future field additions. Cross-crate refactor, worth a follow-up.
- **Shared micros→USD helper** — `micros as f64 / 1_000_000.0` is now the workspace's 5th hand-rolled copy; a `UsdMicros::to_usd_f64()` in fabro-model plus migration of the pre-existing sites belongs in its own change.
- **`priced_mock_catalog` vs `catalog_with` in client tests** — documented as deliberately hermetic, kept.

Verification: full workspace suite 6702 passed (same 5 known environment-dependent fabro-workflow failures as main, confirmed identical with these changes stashed); clippy `-D warnings` and pinned-nightly fmt clean.

PR #496, opened 2026-06-11
refactor(llm): resolve request dispatch through the catalog route

swerner:
PR 8 of the gateway refactor series (after #493) — the optional closer: `Client` dispatch goes through the route machinery #493 introduced, instead of an inline ad-hoc lookup.

## What's here

`Client::resolve_provider`'s hand-rolled catalog hop (`catalog.get(model)` → provider id) becomes `adapter_registry::resolve_route`. Fallback order is byte-identical: explicit `request.provider` wins, then the model's catalog route, then the default provider, then the existing configuration error.

This puts route resolution on the live request path, so the route-equivalence table from #493 now pins actual dispatch rather than a helper nothing calls: a new live-dispatch sweep asserts every built-in model's request lands on the provider its route names, alongside explicit-provider-wins and unknown-model-default pins.

## Scope notes

- **No public API change** — `resolve_provider` is private; all frozen `Client` methods are untouched.
- The route's `codec`/`deployment_id` still aren't handed to adapters: `ProviderAdapter::complete(&Request)` is frozen (prod-implemented in fabro-cli), and every allowed pairing equals the adapter's built-in codec until the feature PRs. This PR is deliberately just the dispatch seam, so the OpenRouter redo's Client-side wiring is a no-op.

## Verification

- `cargo nextest run --workspace --no-fail-fast` (post-rebase onto #493's merge): green except the same 5 pre-existing environment-dependent fabro-workflow failures, identical on main
- Wire snapshots untouched

This closes the refactor series. Remaining: the already-open cost PR (#494), then the feature redos — OpenRouter (#438) and Bedrock (#459).

### DIFF
--- a/lib/crates/fabro-llm/src/client.rs
+++ b/lib/crates/fabro-llm/src/client.rs
@@ -6,7 +6,7 @@ use fabro_model::{AdapterKind, Catalog, ProviderId};
 use tracing::debug;
 
 use crate::adapter_registry::{
-    AdapterConfig, AdapterKindOptions, OpenAiAdapterOptions, factory_for,
+    self, AdapterConfig, AdapterKindOptions, OpenAiAdapterOptions, factory_for,
 };
 use crate::error::{Error, ProviderErrorKind};
 use crate::middleware::{Middleware, NextFn, NextStreamFn};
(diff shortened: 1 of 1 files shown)


swerner:
## `/simplify` review — no changes applied

Ran the four-angle cleanup review (reuse, simplification, efficiency, altitude) over this diff. Every finding was either a false positive, intentional scope, or outside the diff, so the branch is unchanged. Accounting below.

**Reuse — clean.** `client_with_all_catalog_providers` doesn't duplicate an existing helper and is correctly shared across the three new tests; the production change itself removes a hand-rolled lookup in favor of the existing `resolve_route`.

**Simplification — 2 findings, both skipped.**
- Closure shadowing at `client.rs:272` (`route.as_ref().map(|route| ...)`) — same-name shadowing in a short closure is idiomatic and matches the file's style.
- Copy-paste across the three new tests — shared setup is already in the helper; what remains is two lines per test, and parametrizing three clearly-named behavioral pins would trade clarity for nothing.

**Efficiency — 2 findings, both skipped.**
- `resolve_route` runs eagerly even when `request.provider` is explicitly set — real, but the old code had the same eager shape, the cost is a few HashMap lookups and two small clones immediately before an HTTPS call, and the eager `route` binding is the seam the feature PRs (#438, #459) extend to hand `codec`/`deployment_id` to adapters, where the route is likely needed even with an explicit provider. Making it lazy now would be churn undone next PR.
- `resolve_route` internally double-fetches the model (`catalog.get` then `model_settings`) — that's `adapter_registry.rs` from #493, outside this diff.

**Altitude — 1 finding, skipped (noted for later).** The explicit → catalog route → default precedence chain lives in `Client::resolve_provider`, while fabro-server's completions handler implements a similar-but-different chain inline (no default-provider fallback). Unifying them in `adapter_registry` is a behavior decision, not a cleanup, and outside this PR — but worth revisiting if the precedence logic starts spreading during the feature redos.

PR #497, opened 2026-06-12
feat(llm): add OpenRouter as an opt-in built-in provider

swerner:
The first feature payoff of the gateway refactor series (#481–#496): OpenRouter lands as **pure configuration over the `openai_compatible` codec** — no new adapter, no new `AdapterKind`, no OpenRouter codec fork. Redone from #438, which prototyped this pre-refactor as ~2,500 lines including a dedicated adapter and parallel codec plumbing; this PR's fabro-llm diff is the usage-superset decode plus a TOML file.

## What's here (3 commits)

**Per-model `billing_policy` override (fabro-model)** — a model row may override its provider's billing family: the aggregator case, where Claude served through an OpenAI-compatible provider bills Anthropic-style cache reads/writes. `pricing_for`/`billing_facts_for` and the resolved `Route` read the model-effective policy; unknown passthrough model ids keep the provider policy. Pinned by a pricing test (cache writes bill at 1.25× input under the override, $0 under the provider's OpenAI default).

**Aggregator usage superset in the `openai_compatible` codec** — the wire usage struct gains tolerant optional fields:
- `prompt_tokens_details.cached_tokens` / `cache_write_tokens` and `completion_tokens_details.reasoning_tokens` normalize into their disjoint `TokenCounts` buckets with the same subtraction convention as the `openai_responses` codec
- in-band `usage.cost` (OpenRouter returns it on every response) surfaces as `Response.cost_usd` with `cost_source = authoritative`, on both blocking and streamed responses — #494's client-side estimate stamping already defers to it by construction
- **deliberate behavior change owned here**: compat providers that report cached-token details now see them split out of `input_tokens` (previously ignored — the wire pin placed in PR 0 anticipating exactly this change flips, and two new OpenRouter-shaped wire pins land)

**The provider package** — `openrouter.toml` (disabled by default, the Ollama opt-in pattern; curated vendor-namespaced model list; Claude rows set `billing_policy = "anthropic"`; attribution headers deliberately not sent unless the operator opts in via `extra_headers`), `OPENROUTER_API_KEY` env/secret registry entries, a gitleaks rule for `sk-or-v1-` keys, a live e2e test asserting authoritative cost, and docs (integration guide + models concept + config reference).

## Deliberate scope cuts (fidelity follow-ups, per the plan)

- `reasoning_details[]` parse + verbatim multi-turn echo, `cache_control` multipart emission, `provider`/`native_finish_reason` field reads — the new wire pin proves they're tolerated and ignored today
- Typed reasoning-param-style / routing codec params — no catalog row can request reasoning effort yet (no `controls.reasoning_effort` declared), and routing prefs already pass through `provider_options.openrouter` verbatim via the existing adapter-name-keyed merge; typed params land when an operator-level knob actually needs them
- The OpenRouter Anthropic skin (`/api/v1/messages`) — a future pure config row pairing the existing `anthropic_messages` codec with bearer transport

## Verification

- `cargo nextest run --workspace --no-fail-fast`: 6724 passed; only the known 5 pre-existing environment-dependent fabro-workflow failures (identical on main)
- Wire snapshots: one deliberate flip (`decode_usage_ignores_token_details` → `decode_usage_parses_token_details`) + two new OpenRouter pins (blocking cost/cache-write, streamed cost); all other snapshots unmodified
- Builtin catalog unchanged for existing providers: OpenRouter is `enabled = false`, so the #493 route-equivalence table is untouched

Credit to #438 for the provider research, catalog curation, gitleaks rule, and docs structure.


swerner:
Ran a `/simplify` pass (4 parallel review agents: reuse, simplification, efficiency, altitude) over this PR's diff. Applied the findings in b18a02dac; verified with `cargo nextest` (698 passed), pinned-nightly fmt, and clippy `-D warnings`.

**Applied:**

- **`Catalog::effective_billing_policy`** (flagged independently by all 4 agents) — `billing_facts_for` hand-rolled the "model row overrides provider, unknown/passthrough ids fall back" chain that `effective_agent_profile` and `effective_codec` already implement as named helpers. It's now the third sibling in that family, and `billing_facts_for` collapses to two lines. Also drops a redundant double `get()` lookup the inline version paid by going through the public `model_settings()`.
- **Single owner for cost provenance** — the "`cost_source = Authoritative` iff in-band `cost` is present" invariant was stamped verbatim in both the blocking and stream decode paths of the `openai_compatible` codec. It now lives once as `translate::authoritative_cost_source()`. (Hoisting it higher, into `cost.rs`, would be the wrong altitude — the wire pins assert provenance at the codec level, and `apply_estimated_cost` already enforces the estimate-never-overwrites rule at its own single chokepoint.)
- **Minor readability** — `if usage.cost.is_some() { self.cost_usd = usage.cost; }` → `self.cost_usd = usage.cost.or(self.cost_usd);` in the stream decoder, and `decode_response` reads `api_resp.usage.as_ref()` once instead of twice.

**Noted but deliberately skipped:**

- The `.as_ref().and_then().unwrap_or(0)` chains in `ApiUsage::token_counts` mirror the `openai_responses` codec's normalization style on purpose; diverging from that convention to save a line isn't an improvement.
- `price_tokens` resolves model/provider/settings twice (via `billing_facts_for` + `pricing_for`) — real but cleanup-grade (a few HashMap hits per priced stage row on a cold read path), and partly pre-existing `cost.rs` structure outside this PR's scope.
- Efficiency review found nothing blocking: `openrouter.toml` parses once behind the `LazyLock` builtin catalog, and all new per-response work is plain field reads.

The reviews otherwise came back clean — notably, the altitude agent confirmed the per-model billing override resolves once at catalog build (`build_model`) with all consumers reading the resolved field, the authoritative-vs-estimate precedence has a single chokepoint in `apply_estimated_cost`, and nothing special-cases OpenRouter by name outside its generic provider TOML.

PR #505, opened 2026-06-13
design: ChatService trait + Slack refactor + Mattermost + MS Teams

marshally:
## Human Author's Note

Hi @brynary,

I have never written a line of Rust in my life, and I am in way over my head.

I thought it would be good for us to get aligned first on some of the fundamental design decisions that are driving this spec. The spec doc was driven by a Superpowers Brainstorming session with Sonnet 4.6 + Opus 4.8 advisor. I tried to lean into what I remember about your code aesthetics from a decade ago and the idiomatic patterns already expressed in the fabro repo.

Any feedback that you have on these decision points will drive significant change through the spec and plan files. So there's not much point in reviewing those until we have aligned on the philosophy/approach.

## What this is

A design spec for abstracting all chat platform integrations (Slack, Mattermost, MS Teams)
behind a shared `ChatService` trait. This replaces the earlier Mattermost-only design that
Bryan's review identified as not scaling.

**Spec:** [`docs/superpowers/specs/2026-06-13-chat-service-trait-and-providers.md`](https://github.com/marshally/fabro/blob/chat-service-spec/docs/superpowers/specs/2026-06-13-chat-service-trait-and-providers.md)

---

## Approach

The design is structured in three phases:

1. **Slack refactor** — move `SlackService` from `server.rs` to `fabro-slack` and implement the
   new trait. No behavior change; purely structural.
2. **Mattermost implementation** — full parity with Slack: lifecycle notifications and
   interactive interview prompts via buttons and thread replies.
3. **MS Teams stub** — a real but non-functional `TeamsService` that proves the trait contract
   holds for a pure-inbound-HTTP provider before any production Teams code is written.

---

#### 1. `ChatService` trait ([spec](https://github.com/marshally/fabro/blob/chat-service-spec/docs/superpowers/specs/2026-06-13-chat-service-trait-and-providers.md#fabro-chat--trait-and-shared-types) · [Decision 1](https://github.com/marshally/fabro/blob/chat-service-spec/docs/superpowers/specs/2026-06-13-chat-service-trait-and-providers.md#decision-1-chatservice-trait))

Each provider (Slack, Mattermost, Teams) implements four methods: `start()` for the inbound
connection loop, `handle_event()` for outbound Fabro events, `handle_webhook()` for inbound
HTTP, and `kind()` as a discriminant. Adding a fourth provider requires zero changes to
`fabro-server`.

#### 2. New `fabro-chat` crate ([spec](https://github.com/marshally/fabro/blob/chat-service-spec/docs/superpowers/specs/2026-06-13-chat-service-trait-and-providers.md#crate-structure) · [Decision 2](https://github.com/marshally/fabro/blob/chat-service-spec/docs/superpowers/specs/2026-06-13-chat-service-trait-and-providers.md#decision-2-fabro-chat-crate-placement))

The trait lives in a dedicated crate that sits between `fabro-types`/`fabro-interview` and the
provider crates. This mirrors the existing `fabro-sandbox` / `SandboxProvider` pattern exactly.
Putting the trait in `fabro-types` (no async) or `fabro-server` (circular dep) were ruled out.

#### 3. `ChatEventContext` breaks the circular dependency ([spec](https://github.com/marshally/fabro/blob/chat-service-spec/docs/superpowers/specs/2026-06-13-chat-service-trait-and-providers.md#chateventcontext-trait--decision-3) · [Decision 3](https://github.com/marshally/fabro/blob/chat-service-spec/docs/superpowers/specs/2026-06-13-chat-service-trait-and-providers.md#decision-3-chateventcontext-for-circular-dependency))

`handle_event` needs run projections and env-var lookup — both live in `AppState`. Putting
`AppState` in the trait signature would require `fabro-chat` → `fabro-server`, which depends
on `fabro-chat`. Solution: a narrow `ChatEventContext` trait defined in `fabro-chat`;
`AppState` implements it in `fabro-server`. Provider crates never see `AppState`.

#### 4. `handle_webhook` method keeps `axum` out of provider crates ([spec](https://github.com/marshally/fabro/blob/chat-service-spec/docs/superpowers/specs/2026-06-13-chat-service-trait-and-providers.md#webhook-dispatch--decision-4) · [Decision 4](https://github.com/marshally/fabro/blob/chat-service-spec/docs/superpowers/specs/2026-06-13-chat-service-trait-and-providers.md#decision-4-handle_webhook-method-vs-router-fragment))

Providers return a lightweight `WebhookOutcome { status, body }` instead of Axum router
fragments. The server owns a single `POST /api/v1/webhooks/:provider/*rest` route and
dispatches to the right service. Slack's `handle_webhook` always returns 404 — it has no
inbound HTTP surface.

#### 5. `Vec<Arc<dyn ChatService>>` in `AppState` ([spec](https://github.com/marshally/fabro/blob/chat-service-spec/docs/superpowers/specs/2026-06-13-chat-service-trait-and-providers.md#appstate-changes) · [Decision 5](https://github.com/marshally/fabro/blob/chat-service-spec/docs/superpowers/specs/2026-06-13-chat-service-trait-and-providers.md#decision-5-vec-of-chatservice-in-appstate))

Named optional fields (`slack_service`, `mattermost_service`, …) require `server.rs` edits for
every new provider. A `Vec` makes adding a fourth provider a zero-change operation in the
server, mirroring `SandboxProviderRegistry`.

#### 6–7. Why MS Teams stubs exist now ([spec](https://github.com/marshally/fabro/blob/chat-service-spec/docs/superpowers/specs/2026-06-13-chat-service-trait-and-providers.md#phase-3-ms-teams-stubs) · [Decision 6](https://github.com/marshally/fabro/blob/chat-service-spec/docs/superpowers/specs/2026-06-13-chat-service-trait-and-providers.md#decision-6-ms-teams-stubs-included-now) · [Decision 7](https://github.com/marshally/fabro/blob/chat-service-spec/docs/superpowers/specs/2026-06-13-chat-service-trait-and-providers.md#decision-7-start-required-even-for-no-op-providers))

Teams has no outbound WebSocket — it's a pure-inbound-HTTP provider. Without a stub, the
trait could silently encode WebSocket assumptions that would require a breaking change later.
The stub also proves `start()` must be a valid no-op, which is why `start()` is required on
the trait (not optional via a default) for every provider.

---

## What's not in scope

- Full Teams implementation (Adaptive Cards, JWT validation, Bot Framework routing)
- Mattermost OAuth / multi-team routing
- A fourth provider


brynary:
Hi @marshally -- Thanks for working through the design like this as opposed to YOLO'ing a massive PR over. :)

The decisions in this PR look generally good to me, as does the PR-by-PR breakdown. I want to key in on a bit of additional context which will likely be important as you proceed...

One critical thing that I think we need to keep in mind is that there are two different ways Fabro integrates with Chat Services: Notifications (fire-and-forget), and Interviews (which involve users answering questions in e.g. Slack and then Fabro receiving them).

Sending notifications from Fabro to chat services is trivial. It's just an HTTP POST.

However, _receiving_ data presents a challenge because Fabro is often deployed on hosts which are not exposed on the public internet (e.g. laptops or VMs on private networking). Slack explicitly has a [Socket Mode](https://docs.slack.dev/apis/events-api/using-socket-mode/) billed to:

> allows your app to use [the Events API](https://docs.slack.dev/apis/events-api/) and [interactive features](https://docs.slack.dev/interactivity)—without exposing a public HTTP Request URL.

It would be valuable to understand if Mattermost and/or MS Teams have an equivalent. If they do not have an equivalent, then the only way Fabro can receive data from those chat services is by either A) exposing an incoming webhook endpoint reachable by Mattermost / MS Teams; or B) using something to proxy traffic, e.g. a Tailscale Funnel to achieve the same thing (without exposing the host directly).

The answers to these questions inform the shape of the ChatService trait.

Anyways, I think the first phase (which should be an indepdentant PR) is going to ensure that the access to Slack is routed through `trait ChatService`. The outcome of this is that there should not be any references to "slack" in production code outside of the Slack-specific files and directories.

Given the agent's ability to reshape interfaces, it's a bit less important that the ChatService function interface be perfect at the end of that first PR phase. The agent with the assistance of the Rust type checker makes it easy to mold as we move forward with more chat services.

brynary:
@marshally -- I'm doing a pass to keep the PR list tidy, so let's shift this over to an issue when you're ready to pick it back up. Also happy to jump on a Zoom anytime and discuss (and just say hello :)

PR #506, opened 2026-06-13
docs: add DOS trust-substrate integration page

anthony-chaudhary:
## What

Adds a docs page — `docs/public/integrations/dos.mdx` — showing how to wire the [DOS](https://github.com/anthony-chaudhary/dos-kernel) (Dispatch Operating System) trust-substrate MCP server into a Fabro workflow as a **git-evidence verification gate**, and registers it in `docs.json`.

Docs-only: one new `.mdx` plus one nav line. No Rust, no CI surface.

## Why this fits Fabro

Fabro already has two verification rungs:

- a `script` node running `cargo test` — checks **behavior**
- an LLM-as-judge node (the [Definition of Done](https://docs.fabro.sh/examples/definition-of-done) pattern) — checks **judgment**

Neither answers a third question: *did the commit's claim actually match its diff, and did the phase really ship?* When a node re-reads the diff to "verify," that's still a model forming a judgment — consistency, not grounding.

DOS fills exactly that gap. `dos_commit_audit` compares a commit's **subject** (forgeable — whatever the agent typed) against the **files it touched** (recorded by git, not forgeable), and flags a `fix:` that changed only a README, an `--allow-empty "shipped"`, or a "tests pass" commit that deleted the assertions. `dos_verify` answers *did phase X ship?* from git ancestry, reporting `source: "none"` when there is no evidence rather than upgrading a thin signal. Both are author-neutral and put **no model in the loop** — so an agent's "done" is graded against artifacts it could not forge.

The example workflow uses this as a trust gate after the implement-and-commit step: a `CLAIM_UNWITNESSED` verdict routes back into a fix loop instead of exiting on a hollow success. A deterministic `script`-node variant runs the same check as a process exit code.

## Verified

Every command, verdict, and JSON field shown in the page was run against a real git repository — the `dos_commit_audit` `CLAIM_UNWITNESSED`/`subject-only` output, the `OK`/`diff-witnessed` clean case, the `dos_verify` `source: "none"` answer, and the CLI exit codes (0 clean / 1 unwitnessed) the script variant routes on.

The page follows the shape of the existing integration docs (`litellm.mdx`, `openrouter.mdx`, `brave-search.mdx`): frontmatter, setup, MCP config TOML, tool table, a worked example workflow, troubleshooting, and Further-reading cards. All internal links resolve to existing pages; `docs.json` remains valid JSON.

## Test plan

- `docs.json` parses as valid JSON.
- All `/...` internal links point at existing `docs/public/*.mdx` pages.
- MCP config block matches the `[run.agent.mcps.<name>]` stdio schema in `agents/mcp.mdx`.

Happy to adjust tone, trim, or relocate the page (e.g. under Examples instead of Integrations) — whatever fits your docs best.

### DIFF
--- a/docs/public/docs.json
+++ b/docs/public/docs.json
@@ -97,7 +97,8 @@
               "integrations/litellm",
               "integrations/openrouter",
               "integrations/slack",
-              "integrations/brave-search"
+              "integrations/brave-search",
+              "integrations/dos"
             ]
           },
           {
--- a/docs/public/integrations/dos.mdx
+++ b/docs/public/integrations/dos.mdx
@@ -1,0 +1,190 @@
+---
+title: "DOS"
+description: "Verify agent work against git evidence instead of believing the agent's self-report, using the DOS trust-substrate MCP server"
+---
+
+[DOS](https://github.com/anthony-chaudhary/dos-kernel) (the Dispatch Operating System) is a small, deterministic trust substrate for fleets of autonomous agents — "the part that doesn't believe the agents." It ships an [MCP](/agents/mcp) server (`dos-mcp`) that exposes a handful of read-only verification syscalls. Wire it into a workflow and your agents can confirm that a claim is actually true from **git evidence**, rather than from another model's say-so.
+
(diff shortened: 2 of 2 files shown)


brynary:
@anthony-chaudhary thank you for this submission. DOS is an interesting project, and I am excited to try it out at some point.

As far as the docs page, we generally publish integration style docs pages for systems where Fabro directly integrates with an outside system. Since Fabro can run any workflow with any combination of AI / script / etc. steps, it would be too difficult for us to try to maintain docs pages for everything that it could possibly integrate with.

Therefore, instead of having a Fabro docs page about integrating with DOS, I think it would be preferable for DOS to have a docs page about how it works with Fabro. If you were to publish that, I think it would be well received if you could share it (e.g. on Discord, X, etc.).

I'm going to close this PR in the meantime. Thanks again.

PR #510, opened 2026-06-17
Demote control-plane config to plain String; native FABRO_WEB_URL read

swerner:
Third reducing PR of the interpolation unification (**D11 resolution (c)**): the control plane never interpolates. `InterpString` is now strictly the user-facing workflow config language; server identity, storage, listen, object-store, and GitHub App identifiers are plain `String`, consumed where needed with no resolution point.

## Demoted to `String` (was `InterpString`)

Both the layer and resolved types:

- `server.listen.unix.path`, `server.api.url`, `server.web.url`
- `server.storage.root`, `server.artifacts.prefix`, `server.slatedb.prefix`
- object store: `Local.root`, S3 `bucket` / `region` / `endpoint` (shared by artifacts + slatedb)
- `github.app_id` / `client_id` / `slug`

**Kept `InterpString`:** `slack.default_channel` (run-time consumption — the one server-defined survivor). `server.listen.tcp.address` stays the `SocketAddr` `parsed_value` special case.

## Native `FABRO_WEB_URL` read

Deployment-time late binding now goes through a native env read instead of a `{{ env.* }}` token: `FABRO_WEB_URL` overrides `server.web.url` (**env override > settings literal > default**), applied in `canonical_origin` and reused by the JWT issuer, cookie-secure check, and system-info. `docker/split-web` no longer ferries the value through a settings token (compose still sets the env var). `canonical_origin`'s error message now advertises a knob that is actually true for everyone.

## Behavior change (release notes)

- `{{ env.* }}` / `{{ vars.* }}` tokens in the demoted server fields are now **literal text**, not interpolated. The resolve layer emits `warn_if_demoted_template` for every demoted field, so operators with tokens still in server config **fail loud** rather than silently treating the token as a literal.
- Operators who relied on env-based storage location should use the existing native `FABRO_STORAGE_DIR` (`--storage-dir`) override. `FABRO_STORAGE_ROOT` promotion is intentionally deferred (not a proven need).

## Cleanup

`fabro-server`'s `crate::interp` shrinks to just the process-env lookup facade; `resolve_interp` / `_path` / `_with` and the `AppState::resolve_interp` seam are deleted (nothing resolves server-scope `InterpString` anymore).

## Verification

- `cargo build --workspace` ✅
- `cargo +nightly clippy --workspace --all-targets -- -D warnings` ✅ (incl. the `as_source` gate)
- `cargo +nightly fmt --check --all` ✅
[...]

PR #511, opened 2026-06-17
Resolve {{ env.* }} in MCP transport config at the run and exec boundaries

swerner:
First **enhancing** PR of the interpolation unification: now that the reducing PRs have pinned interpolation to the workflow config language, this adds real `{{ env.* }}` resolution for MCP server transports — at **both** the `fabro run` and `fabro exec` boundaries.

Independent off `main` — **no dependency on #510** (zero file overlap; #510 touches the control-plane server settings). Builds on the already-merged InterpString foundation (#472).

## What users get

MCP server transport fields now interpolate `{{ env.* }}` tokens, resolved **at the boundary where the server is actually launched**:

- **stdio / sandbox**: `command`, `args`, and per-server `env` values
- **http**: `url` and `headers`

A literal value passes through unchanged; a `{{ env.NAME }}` token is resolved against the launching process's environment. **Missing env var is a hard error** (D3) instead of the previous behavior where the raw token leaked downstream as literal text. Reserved `secrets`/`inputs` tokens (no resolver here yet) surface as a loud `Unavailable` error rather than passing through.

Resolution happens at the run/exec boundary, not in the shared config resolve layer, so `fabro validate` stays portable (env presence is a runtime concern, not a validation one).

## Both consumers, one resolver

`fabro run` and `fabro exec` read the **same** MCP representation — `run.agent.mcps` and `cli.exec.agent.mcps` both parse through `McpEntryLayer` (InterpString) and collapse via the same `resolve_mcp_entry`. Originally only the run boundary resolved env, so a file-sourced `[cli.exec.agent.mcps.*.env] KEY = "{{ env.X }}"` (from `~/.fabro/settings.toml`) resolved under `run` but **leaked the raw token under `exec`** — a silent asymmetry that would generate confusing bug reports.

This PR closes that by moving the resolution onto the type as `McpServerSettings::resolve_transport_env` (in `fabro-types`, next to the `vars` half `substitute_mcp_transport`), so both consumers share one resolver with no drift:

- `runtime_mcp_server` (run worker) → resolves against the worker process env
- `fabro exec` → resolves against the CLI process env

`runtime_mcp_server` becomes a thin wrapper that just adds the server name to the error.

## Tests

- `fabro-types`: 5 `resolve_transport_env` unit tests — literal passthrough, stdio command+env, http url+headers, sandbox env, missing-env hard error, and the reserved-`secrets` loud-fail case.
- `fabro-workflow`: the 5 existing `runtime_mcp_server_*` tests are unchanged and now exercise the shared resolver through the wrapper.

Files: `fabro-config/src/resolve/run.rs`, `fabro-types/src/settings/run.rs`, `fabro-workflow/src/operations/start.rs`, `fabro-cli/src/commands/exec.rs`.

Verified: `cargo build`, nightly `clippy --all-targets -D warnings`, nightly `fmt --check`, and `cargo nextest run -p fabro-types -p fabro-workflow -p fabro-config -p fabro-cli` (2679 passed with ambient provider keys stripped; the one failure otherwise is the pre-existing ambient-`*_API_KEY` flake, unrelated to MCP).

> Note: the shared resolver takes `Resolved.value` and drops interp `Provenance` (consistent with every other resolved path today — `Provenance` currently has zero consumers, and resolved MCP transport values never surface in logs/events/API). Whether MCP env should carry provenance for precise redaction vs. relying on content-based `fabro-redact` is tracked as an open decision under D4.

### DIFF
--- a/lib/crates/fabro-cli/src/commands/exec.rs
+++ b/lib/crates/fabro-cli/src/commands/exec.rs
@@ -281,6 +281,14 @@ impl ProviderAdapter for AuthenticatedFabroServerAdapter {
     }
 }
 
+#[expect(
+    clippy::disallowed_methods,
+    reason = "exec-boundary MCP transport InterpString resolution facade for {{ env.* }} values."
+)]
+fn process_env_var(name: &str) -> Option<String> {
+    std::env::var(name).ok()
+}
+
 pub(crate) async fn execute(mut args: ExecArgs, ctx: &CommandContext) -> AnyResult<()> {
     use fabro_agent::cli::PermissionLevel as AgentPermissionLevel;
     use fabro_types::settings::run::AgentPermissions;
@@ -312,6 +320,21 @@ pub(crate) async fn execute(mut args: ExecArgs, ctx: &CommandContext) -> AnyResu
     } else {
         cli.exec.agent.mcps.values().cloned().collect()
     };
+    // Resolve `{{ env.* }}` in MCP transport config at the exec boundary,
+    // against the CLI process env — the mirror of the `fabro run` worker
+    // boundary in `fabro_workflow::operations::start::runtime_mcp_server`.
+    // Both consumers read the same source-form settings; missing env is a hard
+    // error (D3) and reserved secrets/inputs tokens surface loudly rather than
+    // leaking.
+    let mcp_servers = mcp_servers
+        .into_iter()
+        .map(|settings| {
+            let name = settings.name.clone();
+            settings
+                .resolve_transport_env(process_env_var)
+                .with_context(|| format!("failed to resolve MCP server {name:?}"))
+        })
+        .collect::<AnyResult<Vec<_>>>()?;
     if let Some(target) = server_target {
         tracing::info!(transport = "server", "Agent session starting");
         let provider_name = args
--- a/lib/crates/fabro-cli/src/commands/exec.rs
+++ b/lib/crates/fabro-cli/src/commands/exec.rs
@@ -329,10 +329,9 @@ pub(crate) async fn execute(mut args: ExecArgs, ctx: &CommandContext) -> AnyResu
     let mcp_servers = mcp_servers
         .into_iter()
         .map(|settings| {
-            let name = settings.name.clone();
             settings
                 .resolve_transport_env(process_env_var)
-                .with_context(|| format!("failed to resolve MCP server {name:?}"))
+                .with_context(|| format!("failed to resolve MCP server {:?}", settings.name))
         })
         .collect::<AnyResult<Vec<_>>>()?;
     if let Some(target) = server_target {
--- a/lib/crates/fabro-config/src/resolve/run.rs
+++ b/lib/crates/fabro-config/src/resolve/run.rs
@@ -296,8 +296,9 @@ fn resolve_agent(agent: Option<&RunAgentLayer>) -> RunAgentSettings {
 
 #[expect(
     clippy::disallowed_methods,
-    reason = "known leak: MCP transport templates collapse to raw source unresolved; strict \
-              resolution scheduled in the interpolation unification (Phase 2, supersedes PR #370)"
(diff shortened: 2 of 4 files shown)


swerner:
## `/simplify` pass (Opus 4.8)

Ran `/simplify` over this diff — quality-only review (reuse, simplification, efficiency, altitude), not a bug hunt. Four review angles ran in parallel; deduped findings below.

### Applied (commit `0f4161b`)

**1. `fabro-types/src/settings/run.rs` — struct-update syntax in `resolve_transport_env`**
The method hand-copied all 6 fields just to swap `transport`. Replaced with `Ok(Self { transport, ..self.clone() })` — only the changed field is named, and a future field on `McpServerSettings` is carried automatically instead of being silently dropped. This was the same fragile per-field-copy pattern the PR already removed from `start.rs`. *(Flagged by Simplification + Efficiency.)*

**2. `fabro-cli/src/commands/exec.rs` — dropped the redundant `name.clone()`**
`resolve_transport_env` takes `&self`, so `settings` is never moved; the eager `let name = settings.name.clone()` was unnecessary. Now reads `settings.name` directly in the `with_context` closure, which also makes the exec boundary consistent with the run boundary (`start.rs` already reads `settings.name` lazily in `map_err`). *(Flagged by Simplification + Efficiency.)*

No behavior change. Verified: `cargo build`, the 5 `resolve_transport_env` + 5 `runtime_mcp_server` + 46 `cmd::mcp` tests, nightly `clippy -D warnings`, and `fmt --check` all pass.

### Skipped — with rationale

- **`resolve_env_string` parse-on-every-literal fast-path** (Efficiency): a `contains("{{")` guard would save one alloc per string, but MCP config is small and resolved once at startup. The agent itself graded it negligible — adding a branch for a startup micro-alloc is a net complexity *add*. Skipped.
- **`process_env_var` duplicated 5× in `fabro-cli`** (Reuse): real duplication (`fabro-server` consolidated theirs into one `interp` module), but the fix touches 4 sites outside this diff and the new copy follows the existing crate convention. Out of scope here; possible follow-up.
- **`mcp_env_lookup` / `env_lookup` test helpers duplicated across crates** (Reuse): blocked by the crate/`test-support` boundary — sharing 8 lines would require a new test-support surface. Not worth it.
- **Push error-name-prefixing onto the shared resolver** (Altitude): sharpest finding — the `"failed to resolve MCP server {name}"` message is built in both consumers over two error types and can drift. But the clean fix is a new error type (`McpTransportResolveError { name, source }`) threaded through both crates — an error-type design change, not a quality cleanup. Deferred as a recommendation.

### Confirmed clean
Reuse of `InterpString::resolve`, the `secrets`/`inputs` → `Unavailable` behavior riding the general resolution context, and the `.map().collect::<Result<Vec<_>,_>>()` in `start.rs` — all correct, no changes.

### Worth flagging (not a code change)
Altitude noted that `environment.env` still uses the leaky `resolve_or_source` (silently falls back to raw source on a missing env var), making it the lone D3 holdout doing the exact thing this PR fixes for MCP. Might warrant a one-line note or tracking ref so the two boundaries don't drift further.

swerner:
Simplify review completed and follow-up cleanup pushed.

What changed:
- Shared MCP transport string traversal between vars substitution and env resolution so transport fields cannot drift.
- Loosened `resolve_transport_env` to accept `FnMut` env lookups, making cached/instrumented lookup facades usable.
- Fast-pathed literal MCP transport strings before interpolation parsing.
- Preserved `ResolveError` as the source when run-boundary MCP resolution fails.
- Trimmed workflow tests to wrapper behavior and relied on type-level resolver tests for the transport matrix.

Simplify was run with the three review agents: reuse, quality, and efficiency. The cleanup commit was pushed after that pass.

Validation run:
- `cargo nextest run -p fabro-types resolve_transport_env_tests`
- `cargo nextest run -p fabro-workflow runtime_mcp_server_wraps_resolve_error_source`
- `cargo check -p fabro-cli`
- `git diff --check`

PR #512, opened 2026-06-18
A goal can't reference itself; prompts can reference the goal

swerner:
Implements the goal self-reference behavior for interpolation unification. Independent off `main` — no dependency on the other interp PRs (touches only the template/goal-render path).

## What changes for users

A graph `goal` is a template that interpolates `{{ inputs.* }}` (unchanged). A node `prompt` can reference the rendered goal via `{{ goal }}` (unchanged). **New:** a goal can **no longer reference itself** — `{{ goal }}` *inside* a goal was previously a silent passthrough (left as the literal text `{{ goal }}`); it's now a clear error.

```
graph [goal="Refine {{ goal }}"]   # error: a goal cannot reference itself
work  [prompt="Work on {{ goal }}"] # fine: prompts reference the rendered goal
```

## How

- **Structural guarantee:** the goal renders with **no `goal` key in scope** (`TemplateContext::new().with_inputs(..)` instead of the `for_input_scan` passthrough), so a self-reference can't resolve.
- **Friendly lint:** before rendering, `resolved_goal` checks the goal template for a top-level `goal` reference — new `fabro_template::references_top_level_variable`, backed by MiniJinja `undeclared_variables` — and emits a dedicated `goal_self_reference` diagnostic (`Severity::Error`) with a clear message and fix-it, instead of a generic "undefined variable `goal`". Fails `fabro validate` and run-create alike.

The goal is resolved in two transform passes (FileInlining + TemplateTransform); the diagnostic is emitted **once** (FileInlining discards its goal-resolution diagnostics; TemplateTransform is the canonical emitter).

## Behavior change (release notes)

A goal containing `{{ goal }}` now **errors** instead of passing through as literal text. The error message is the migration signal.

## Tests

- `references_top_level_variable` detection
- transform-level rejection (`Severity::Error`)
- single-emission across the two passes
- end-to-end `validate` rejection
- existing goal/prompt tests still green (prompts reference goal; goal interpolates inputs)

## Verification

- `cargo build --workspace`
- `cargo +nightly clippy --workspace --all-targets -- -D warnings`
- `cargo +nightly fmt --check`
- `cargo nextest run --workspace`: 6800 passed

Generated with [Claude Code](https://claude.com/claude-code)

PR #513, opened 2026-06-18
inputs is template-only: reject {{ inputs.* }} in InterpString with a clear error (D12)

swerner:
Implements the `inputs`-template-only half of **D12**. Independent off `main` — touches only `fabro-types` interp; no overlap with #511 or #512.

## What changes for users

`{{ inputs.* }}` in an `InterpString` field (command, script, header, env, URL — MCP transports, prepare steps, hooks, server settings) now fails with a **clear, actionable message**:

> `{{ inputs.X }}` is only available in prompts and goals, not in command, script, header, env, or URL fields

It *already* failed there (no resolve context ever provided an inputs lookup, so it errored as a generic "unavailable"); this makes the rejection explicit and points the user at where `inputs` belongs.

## How

- **`ResolveCtx` drops its unused `inputs` lookup** (`with_inputs` had zero production callers). The type now structurally cannot resolve `inputs` in an `InterpString` field; `lookup_for(Inputs)` returns `None`.
- The `Unavailable` error message is `inputs`-specific and points to prompts/goals.
- **`substitute_with` still preserves `inputs` tokens** (unknown-namespace passthrough), so `run.goal` — an `InterpString` that feeds a template — keeps forwarding `{{ inputs.* }}` to its prompt/goal render. This is the load-bearing behavior that makes "inputs works in goals" coexist with "inputs rejected in InterpString fields", and it's covered by an existing test (`substitute_variables_preserves_late_bound_tokens`).
- Module docs updated: three resolvable namespaces in `InterpString` (`env`/`vars`/`secrets`); `inputs` is template-only.

## Note on timing

The rejection fires at **resolve time** (use-time / run boundary), not at `fabro validate`. That matches how the other late-bound namespaces behave and keeps this PR small; a validate-time fail-fast would need to distinguish goal (forwards inputs) from pure-`InterpString` fields and is a larger, separate change if we want it.

## Tests

`resolve_with_rejects_inputs_as_template_only` (rejection + friendly message); `substitute_variables_preserves_late_bound_tokens` confirms goal forwarding is unaffected.

Verified: `cargo build --workspace`, nightly `clippy --workspace --all-targets -D warnings`, `fmt`, `cargo nextest run --workspace` (**6796 passed**).

### DIFF
--- a/lib/crates/fabro-types/src/settings/interp.rs
+++ b/lib/crates/fabro-types/src/settings/interp.rs
@@ -1,17 +1,19 @@
 //! Interpolation for config strings.
 //!
 //! An [`InterpString`] field may contain narrow `{{ <namespace>.NAME }}`
-//! tokens — no template logic — drawn from the four [`Namespace`]s: `env`,
-//! `vars`, `secrets`, and `inputs`. Which namespaces actually resolve is
-//! scope-determined by the caller through [`ResolveCtx`]: server-scope
-//! settings provide `env` (and eventually `secrets`), run-scope settings
-//! additionally provide `vars` and `inputs`. A token whose namespace is not
-//! available in the resolution context fails loudly instead of passing
-//! through as literal text.
+//! tokens — no template logic. Three [`Namespace`]s resolve here: `env`,
+//! `vars`, and `secrets`. `inputs` is **template-only** (D12): it is a
+//! recognized namespace so an `{{ inputs.* }}` token fails loudly with a clear
+//! message instead of passing through as literal text, but it never resolves
+//! in an `InterpString` field — it belongs in prompts and goals. Which of the
+//! resolvable namespaces actually apply is scope-determined by the caller
+//! through [`ResolveCtx`]: server-scope settings provide `env` (and eventually
+//! `secrets`), run-scope settings additionally provide `vars`. A token whose
+//! namespace is not available in the resolution context fails loudly.
 //!
-//! Resolution timing is split: `vars`/`inputs` substitute early (server-side,
-//! at run creation) via [`InterpString::substitute_with`], while
-//! `env`/`secrets` resolve late, at consumption time in the process that owns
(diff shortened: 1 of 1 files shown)


swerner:
## `/simplify` review (4 cleanup agents: reuse · simplification · efficiency · altitude)

Reviewed the diff for quality (reuse / simplification / efficiency / altitude) — not correctness. One fix applied; the rest of the diff was already clean.

### Fixed — Altitude

The `inputs` rejection message hardcoded a field enumeration inside `ResolveError`'s `Display`:

> …not in **command, script, header, env, or URL** fields

That list leaked caller-level schema knowledge into the low-level `fabro-types` interp layer (no single source of truth), and was already inaccurate:

- **Omitted** real `InterpString` fields: `permissions`, Slack `channel` / `default_channel`.
- **Named** `header` / `url`, which are plain `String` fields that flow through the early `substitute_variables_in_str` (vars-only, token-preserving) pass — not the `resolve_with` path where this error actually fires.

It would also rot as `InterpString` fields are added/renamed, with no test to catch the drift. Rewrote it to state the rule from the stable, positive side:

> `{{ inputs.X }}` is only available in prompts and goals, not in other config fields

Same actionable pointer, nothing to drift. The existing test (`resolve_with_rejects_inputs_as_template_only`) asserts the `"only available in prompts and goals"` substring and still passes.

### Skipped (with reasons)

- **Reuse** — nothing. No existing field-list to reuse; `noun()` is intentionally *not* reused since the inputs message is a deliberately different shape.
- **Efficiency** — nothing. Dropping the `inputs` `LookupFn` field shrinks `ResolveCtx` (~16 bytes + one fewer boxed closure) — a net win.
- **Simplification** — `Namespace::noun()`'s `Inputs => "input"` arm is now an unreachable value (inputs errors are always `Unavailable`, and that arm doesn't call `noun()`). Left as-is: `noun()` must stay exhaustive over all variants, and a wildcard / `unreachable!()` would *reduce* clarity rather than improve it.

PR #515, opened 2026-06-19
Implement full-fidelity ACP session reuse

### LINKED ISSUE #514: Support `fidelity="full"` session reuse for `backend="acp"` agent nodes
## Problem

ACP-backed agent stages currently run as one-shot turns. For each ACP node, Fabro starts the configured ACP process, initializes the protocol, creates a new ACP session, sends one `session/prompt`, reads the response, and terminates the process.

That behavior is correct for isolated nodes, but it makes `fidelity="full"` much less useful for ACP than it is for the API backend:

- Full-fidelity ACP nodes do not get a preamble.
- They also do not reuse the prior conversation/session.
- Each node pays the startup cost for heavyweight ACP adapters such as Codex or Claude Code.

For multi-node implementation workflows, continuing the same ACP session would both preserve richer context and avoid repeatedly starting expensive external agent processes.

## Desired behavior

When an ACP-backed agent node runs with `fidelity="full"`, Fabro should keep the ACP process and ACP session alive after a successful turn and reuse that session for later compatible full-fidelity ACP nodes.

The reused session should receive later node prompts via ACP `session/prompt`. Fabro should not restart the process or create a new ACP session unless the workflow explicitly targets a different session, the cached session is incompatible, or the cached session has been discarded because of an error, timeout, or cancellation.

## Session selection rules

- `fidelity="full"` with an explicit `thread_id` should use the ACP session for that thread ID.
- `fidelity="full"` without an explicit `thread_id` should implicitly reuse the previous ACP full-fidelity session along the active workflow path.
- A later explicit `thread_id` may target a different named ACP session.
- Non-`full` fidelity modes should keep the current behavior: build a preamble, run an isolated ACP turn, and clean up the process after the node.

#### Compatibility and validation

Reusing a cached ACP session is only valid when the session was created for the same effective ACP process and workspace context.

At minimum, Fabro should treat these as part of the reuse identity:

- `acp.command` / `acp.config`, after parsing into the effective process spec
- sandbox provider and working directory
- any launch environment material that affects the ACP process

If two nodes would target the same ACP session but specify incompatible ACP process configuration, Fabro should fail validation with a clear diagnostic rather than silently starting a different process and pretending full fidelity was preserved.

## Failure and cleanup behavior

Cached ACP sessions should be discarded and cleaned up on:

- node timeout
- user cancellation or interrupt that ends the node
- ACP protocol error
- ACP process exit before the session is intentionally closed
- output/schema/runtime failure paths that leave the session state uncertain
- normal run shutdown

Fabro should not attempt to keep a cached ACP session alive after timeout or cancellation, even if the external agent appears to report a clean cancellation.

## Implementation notes

The current `fabro_acp::run_acp_turn` API is turn-scoped, so this likely needs a new session-scoped abstraction rather than a small modification to the helper.

One plausible shape:

- Add a persistent ACP session actor/task in `fabro-acp`.
- The actor owns the ACP stdio process, transport state, JSON-RPC connection, and ACP `ActiveSession`.
- Expose a cloneable handle with methods such as `prompt(...)`, `interrupt(...)`, `has_pending_control_work()`, and `shutdown(...)`.
- Move the existing live-control loop so it can run multiple sequential prompts against the same ACP session.
- Add an ACP session cache to `AgentAcpBackend`, analogous to the API backend's full-fidelity session cache.
- Cache sessions on successful full-fidelity turns and remove/discard them before each use, reinserting only after a clean successful turn.
- Keep stage-scoped events, file-change snapshots, timeout handling, and steering activation per node even when the underlying ACP session is reused.

## Events and observability

The UI and run log should make reuse visible enough to debug:

- `agent.acp.started` should indicate whether the ACP process was newly started or reused.
- Reused stages should still emit stage-scoped prompt/completion events.
- Session activation/deactivation events should continue to expose the currently steerable stage/session while a reused ACP process is active.
- Run shutdown should emit cleanup events for any cached ACP sessions that are still alive.

## Acceptance criteria

- A workflow with two sequential `backend="acp"` nodes and `fidelity="full"` reuses one ACP process/session for both node prompts when no `thread_id` is specified.
- A workflow with two full-fidelity ACP nodes sharing the same explicit `thread_id` reuses the same ACP process/session.
- A workflow with full-fidelity ACP nodes using different explicit `thread_id` values creates distinct reusable ACP sessions.
- A validation test rejects nodes that would target the same ACP session with incompatible `acp.command` / `acp.config`.
- Timeout or cancellation discards the cached ACP session and terminates the external process.
- Non-`full` ACP nodes continue to run as isolated one-shot ACP turns with preamble insertion.
- Existing ACP steering behavior still works during a reused ACP session.
- `cargo nextest run -p fabro-acp -p fabro-workflow` passes.

## Non-goals

- Do not reuse ACP sessions across different workflow runs.
- Do not persist ACP sessions across checkpoint resume or process restart.
- Do not attempt to make Daytona ACP work as part of this feature.
- Do not add Claude subscription OAuth to the API backend; this issue is only about ACP process/session reuse.
- Do not make `backend="acp"` support `output_schema`; keep the current unsupported behavior unless a separate issue changes it.

jessmartin:
## Summary

- Adds a reusable ACP session handle that keeps the initialized ACP process/session alive across prompts.
- Enables `backend="acp"` agent nodes with resolved `fidelity="full"` to reuse sessions by explicit `thread_id`, or by an implicit run/parallel-branch scoped key when no explicit thread is set.
- Discards reusable ACP sessions on timeout, cancellation, protocol/process errors, and process-config mismatches.
- Adds `agent.acp.started.properties.reused` for observability.
- Adds validation for statically-detectable full-fidelity ACP nodes sharing an explicit/default `thread_id` with mismatched ACP process configuration.

Closes #514.

## Testing

- `cargo check -p fabro-acp`
- `cargo check -p fabro-workflow`
- `cargo check -p fabro-validate`
- `cargo check -p fabro-server`
- `cargo test -p fabro-validate backend_valid`
- `cargo test -p fabro-workflow acp_backend_full_fidelity`
- `cargo test -p fabro-acp`
- `cargo test -p fabro-workflow handler::llm::acp::tests`
- `cargo +nightly-2026-04-14 clippy -p fabro-acp -p fabro-workflow -p fabro-validate -p fabro-server --all-targets -- -D warnings`

### DIFF
--- a/lib/crates/fabro-acp/src/error.rs
+++ b/lib/crates/fabro-acp/src/error.rs
@@ -36,6 +36,9 @@ pub enum AcpError {
     #[error("ACP turn was cancelled")]
     Cancelled,
 
+    #[error("ACP session closed")]
+    SessionClosed,
+
     #[error("ACP turn timed out")]
     TimedOut {
         exec_output_tail: Option<ExecOutputTail>,
--- a/lib/crates/fabro-acp/src/lib.rs
+++ b/lib/crates/fabro-acp/src/lib.rs
@@ -16,5 +16,6 @@ pub use command::{AcpCommandError, AcpProcessSpec};
 pub use error::{AcpError, AcpProcessExit};
 #[cfg(feature = "runtime")]
 pub use session::{
-    AcpControlHandle, AcpLiveControl, AcpRunRequest, AcpRunResult, render_stop_reason, run_acp_turn,
+    AcpControlHandle, AcpLiveControl, AcpPromptRequest, AcpRunRequest, AcpRunResult,
+    AcpSessionHandle, AcpSessionStartRequest, render_stop_reason, run_acp_turn, start_acp_session,
 };
--- a/lib/crates/fabro-acp/src/session.rs
+++ b/lib/crates/fabro-acp/src/session.rs
@@ -12,8 +12,8 @@ use agent_client_protocol::{ActiveSession, Agent, Client, Error as ProtocolError
 use fabro_sandbox::Sandbox;
 use fabro_types::{Principal, SteeringMessage};
 use fabro_util::time::elapsed_ms;
-use tokio::sync::Notify;
 use tokio::sync::futures::Notified;
+use tokio::sync::{Notify, mpsc, oneshot};
 use tokio::time::{sleep, timeout};
 use tokio_util::sync::CancellationToken;
 
@@ -23,6 +23,8 @@ use crate::transport::{SandboxAcpTransport, TransportState};
 
 pub type AcpNaturalCompletionCallback = Arc<dyn Fn() -> bool + Send + Sync>;
 pub type AcpSteerPromptCallback = Arc<dyn Fn(String, Option<Principal>) + Send + Sync>;
+type PermissionCancelSlot = Arc<Mutex<Option<CancellationToken>>>;
+type AcpReadySlot = Arc<Mutex<Option<oneshot::Sender<Result<(), AcpError>>>>>;
 
 const CANCEL_GRACE_PERIOD: Duration = Duration::from_millis(500);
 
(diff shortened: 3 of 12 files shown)


jessmartin:
Follow-up note: the next thing to do here is end-to-end testing of the ACP full-fidelity session reuse path before moving this out of draft or considering merge.

PR #517, opened 2026-06-23
feat(sandbox): forkd Firecracker microVM provider

zenprocess:
# feat(sandbox): forkd Firecracker microVM provider

## Why

fabro already supports Local, Docker, and Daytona sandboxes. Daytona is a great
fit for managed/cloud remote VMs. This PR adds a complementary option for teams
that need execution to stay fully **self-contained and on-premise**:

- **Strong isolation without a shared kernel.** Firecracker microVMs give each
  step a real VM boundary (vs Docker's shared host kernel), closer to Daytona's
  isolation but without a hosted control plane.
- **Self-hosted, zero external dependency.** forkd runs on your own hardware via
  a small controller API. No SaaS account, no per-VM cloud billing, no data
  leaving the network. Ideal for regulated/air-gapped/on-prem deployments.
- **Fast clone-from-snapshot.** forkd's snapshot+clone model gives microVM-grade
  isolation with container-like start times, so per-step VMs stay practical.

It is purely additive: a new `SandboxProvider` selected via config, mirroring the
existing daytona provider's structure. Local/Docker/Daytona are unchanged.

## What

- `SandboxCreateSpec::Forkd` + `SandboxProviderKind::Forkd`
- `ForkdConfig`/`ForkdSettings` (controller URL, token, snapshot/image) in config
- A thin forkd controller HTTP client (clone snapshot, exec, delete; token auth)
- `ForkdSandboxProvider` (impl `SandboxProvider`: create/list/get/delete)
- `ForkdSandbox` (impl the full `Sandbox` trait over a microVM: exec, streaming,
  file up/down, lifecycle, git)
- Registry + `from_environment` wiring; tests mirroring the sandbox round-trips

Credentials (forkd token) come from config/env, never hardcoded.

## Status

First cut. Behavior contract: 27 BDD scenarios across the provider + Sandbox trait
surface. Compile/CI verification pending on a Rust-capable runner.

### DIFF
--- a/lib/crates/fabro-workflow/src/operations/start.rs
+++ b/lib/crates/fabro-workflow/src/operations/start.rs
@@ -11,7 +11,7 @@ use fabro_model::{Catalog, FallbackTarget, ProviderId};
 use fabro_sandbox::daytona::DaytonaConfig;
 use fabro_sandbox::from_environment::{
     daytona_config_from_environment, docker_config_from_environment,
-    local_working_directory_from_environment,
+    forkd_config_from_environment, local_working_directory_from_environment,
 };
 use fabro_sandbox::{DockerSandboxOptions, SandboxSpec};
 use fabro_static::EnvVars;
@@ -420,6 +420,17 @@ impl RunSession {
                     api_key,
                 }
             }
+            #[cfg(feature = "forkd")]
+            SandboxProviderKind::Forkd => SandboxSpec::Forkd {
+                config: Box::new(resolve_forkd_config(resolved)),
+                run_id: Some(record.run_id),
+                clone_origin_url: record.repo_origin_url().map(str::to_string),
+                clone_branch: record.base_branch().map(str::to_string),
+            },
+            #[cfg(not(feature = "forkd"))]
+            SandboxProviderKind::Forkd => {
+                return Err(Error::engine("Forkd sandbox support is not enabled"));
+            }
         };
 
         let toml_env = resolved.environment.resolve_env(process_env_var);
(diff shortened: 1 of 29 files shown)


lib/crates/fabro-workflow/src/operations/start.rs
```diff
@@ -561,6 +572,11 @@ fn resolve_daytona_config(settings: &ResolvedRunSettings) -> DaytonaConfig {
     daytona_config_from_environment(&settings.environment, !settings.clone.enabled)
 }
 
+#[cfg(feature = "forkd")]
+fn resolve_forkd_config(settings: &ResolvedRunSettings) -> fabro_sandbox::ForkdConfig {
+    forkd_config_from_environment(&settings.environment, !settings.clone.enabled)
+}
+
 fn resolve_docker_config(settings: &ResolvedRunSettings) -> DockerSandboxOptions {
```
brynary: Pre-existing naming asymmetry here.. *Options vs. *Config

brynary:
@zenprocess thanks for this PR. I've begun a manual review. Primarily my review is bringing to light some pre-existing architectural design issues that we may want to resolve before adding the next sandbox provider. More soon.

zenprocess:
Thanks for the update, feel free to share your architectural concerns perhaps we can contribute with ideas!

zenprocess:
Superseded by #567 — we rebased this provider onto current `main` (this branch had fallen ~50 commits behind, which left it in a diverged/unknown-mergeable state). #567 is the same forkd provider plus feature-gating fixes, reconnect env resolution, a DB migration for the environments `provider` CHECK, and a `docker-build --features` passthrough. @brynary — your note here about pre-existing architectural design issues to resolve before adding the next sandbox provider never got the follow-up detail; we'd genuinely like to align on it. #567 deliberately conforms to the existing `Sandbox` trait / provider-registry / clone-based contracts rather than changing them, so if there's a refactor you want as a precondition, name it and we'll take it on. Happy to move that to an Issue/Discussion if that's the better venue per CONTRIBUTING.

swerner:
Closing this one as it is superseded by #567

PR #520, opened 2026-06-23
fix(mcp): honor inline enabled=false and per-server tool_timeout

swerner:
## What

Two latent fixes to MCP server config handling, independent of any new feature:

1. **`enabled = false` is now honored for inline MCP servers.** Entries under `[run.agent.mcps.*]` and `[cli.exec.agent.mcps.*]` accepted an `enabled` flag that resolution silently ignored, so a disabled server still started. Disabled entries are now dropped from the resolved set. Absent `enabled` still means enabled.
2. **Explicitly configured empty `cli.exec.agent.mcps` sets are preserved.** If every `cli.exec` MCP entry is disabled, `fabro exec` now treats that as an intentional empty override instead of falling back to `run.agent.mcps`.
3. **Per-server `tool_timeout_secs` now applies to MCP tool calls.** The value was carried through config but never reached the call path. The connection manager now owns each server timeout and applies it when calling tools.

## Testing

- New and updated tests cover StickyMap same-key replacement across layers, `enabled = false` skipped for run and `cli.exec`, absent `enabled` kept, higher-layer disable shadowing, explicit empty `cli.exec` MCP overrides, and configured tool timeout behavior.
- `cargo nextest run -p fabro-config -p fabro-agent -p fabro-mcp`: 737 passed, 93 skipped.
- `cargo +nightly-2026-04-14 clippy -p fabro-config -p fabro-agent -p fabro-mcp -p fabro-cli --all-targets -- -D warnings`
- `cargo test --locked -p fabro-workflow --test it --no-run`

## Notes

- **Behavior change** worth a changelog entry: disabled inline MCPs are now actually disabled, explicit empty `cli.exec` MCP overrides are respected, and per-server tool timeouts now take effect.
- First of a short series adding server-managed MCP servers; this PR is self-contained and independent of the others.

### DIFF
--- a/lib/crates/fabro-agent/src/mcp_integration.rs
+++ b/lib/crates/fabro-agent/src/mcp_integration.rs
@@ -16,7 +16,10 @@ pub fn make_mcp_tools(manager: &Arc<McpConnectionManager>) -> Vec<RegisteredTool
             let name = qualified_name.clone();
             let server_name = info.server_name.clone();
             let original_name = info.original_tool_name.clone();
-            let tool_timeout = std::time::Duration::from_mins(2);
+            // Per-server `tool_timeout_secs` config, threaded through `ToolInfo`
+            // at registration time. Previously hardcoded to two minutes, which
+            // silently dropped any configured per-server tool timeout.
+            let tool_timeout = info.tool_timeout;
 
             RegisteredTool {
                 definition: ToolDefinition {
--- a/lib/crates/fabro-agent/src/mcp_integration.rs
+++ b/lib/crates/fabro-agent/src/mcp_integration.rs
@@ -16,10 +16,6 @@ pub fn make_mcp_tools(manager: &Arc<McpConnectionManager>) -> Vec<RegisteredTool
             let name = qualified_name.clone();
             let server_name = info.server_name.clone();
             let original_name = info.original_tool_name.clone();
-            // Per-server `tool_timeout_secs` config, threaded through `ToolInfo`
-            // at registration time. Previously hardcoded to two minutes, which
-            // silently dropped any configured per-server tool timeout.
-            let tool_timeout = info.tool_timeout;
 
             RegisteredTool {
                 definition: ToolDefinition {
@@ -30,10 +26,9 @@ pub fn make_mcp_tools(manager: &Arc<McpConnectionManager>) -> Vec<RegisteredTool
                 executor:   Arc::new(move |args, _ctx| {
                     let mgr = Arc::clone(&mgr);
                     let name = name.clone();
-                    let timeout = tool_timeout;
                     Box::pin(async move {
                         let result = mgr
-                            .call_tool(&name, args, timeout)
+                            .call_tool(&name, args)
                             .await
                             .map_err(|e| e.to_string())?;
                         call_result_to_string(&result)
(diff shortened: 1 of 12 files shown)


swerner:
## `/simplify` pass

Ran four cleanup agents over this diff (reuse, simplification, efficiency, altitude), deduped the findings, and applied one fix.

### Applied

**Centralized the `enabled = false` filter into a shared `resolve_enabled_mcps` helper** (`resolve/run.rs` + `resolve/cli.rs`).

Both inline-MCP resolution sites had the same `.filter(|(_, entry)| entry.is_enabled()).map(...).collect()` chain. Since both sides are `StickyMap<McpEntryLayer>` → `HashMap<String, McpServerSettings>`, I pulled it into one `pub(crate)` helper they both call, so the enable check lives in one place and a future third inline-MCP site inherits it instead of having to remember the filter. The `cli.exec` site collapsed from a 14-line nested `map`/`filter`/`map` to a single call. `resolve_mcp_entry` stays a pure single-entry transform — the enable policy didn't get folded into it.

No behavior change. Verified: `cargo nextest run -p fabro-config` → 233 passed; fmt + clippy clean on the crate.

### Reviewed and skipped

- **`tool_timeout` on `ToolInfo` as "config leaking into a registry struct"** — false positive. `make_mcp_tools` flattens `all_tools()` across multiple servers with differing timeouts; there's no single config to thread in, and `ToolInfo` already carries per-tool `server_name`, so it's the correct per-tool carrier.
- **Duplicate `parse_settings` test helper** (now a 4th identical copy) — out of scope. The file's existing convention is one local copy per test module; consolidating properly means editing 3 modules this PR doesn't touch. Worth a separate test-helper cleanup.
- **Scattered defaults (`enabled→true`, `tool_timeout→60s`)** and **carrying `enabled` forward into `McpServerSettings`** — the `60s` defaults pre-date this PR, and carrying `enabled` forward is a design change, not a simplification.

Everything else (the `is_enabled()` match shape, test-fixture duplication, closure captures, the once-per-server timeout hoist) was judged already clean.

swerner:
Simplify review (GPT-5.5 x-high) complete.

Agents:
- Code reuse: no findings.
- Code quality: fixed the two high-value issues found.
- Efficiency: no findings.
- Skipped: allowing `enabled = false` tombstones without `type` or transport fields would require a broader schema/deserialization change, so I left that out of this cleanup.

Pushed `bca1094df`:
- Preserved configured-empty `cli.exec.agent.mcps` so all-disabled CLI MCP overrides do not fall back to `run.agent.mcps`.
- Moved configured MCP tool timeout ownership into `McpConnectionManager`; `fabro-agent` no longer carries timeout plumbing through `ToolInfo`.
- Added behavior coverage for explicit empty CLI MCP overrides and configured timeout application.

Verification:
- `cargo nextest run -p fabro-config -p fabro-agent -p fabro-mcp`: 737 passed, 93 skipped.
- `cargo +nightly-2026-04-14 clippy -p fabro-config -p fabro-agent -p fabro-mcp -p fabro-cli --all-targets -- -D warnings`

swerner:
CI fix pushed in `d109850ec`.

Root cause: the cleanup changed `McpConnectionManager::call_tool` to own per-server timeouts internally, but three `fabro-workflow` integration test call sites still used the old 3-argument manager API. CI compiles the full workspace test set, so it caught those stale calls.

Fix: updated the three `daytona_integration.rs` manager calls to the new 2-argument API. Raw `McpClient::call_tool` calls still keep explicit timeout arguments.

Verification:
- `cargo test --locked -p fabro-workflow --test it --no-run`

PR #521, opened 2026-06-23
feat(mcp): add server-side MCP server store (fabro-mcp-store)

swerner:
## What

Adds the storage foundation for server-managed MCP servers: a durable store plus its domain model. No server wiring, HTTP API, or UI yet — this is standalone scaffolding that later PRs build on.

- New **`fabro-mcp-store`** crate: a concrete, filesystem-backed `McpServerStore` — one TOML file per definition under `{active-config-dir}/mcps/`, an in-memory cache, and a SHA-256 content-hash revision for optimistic concurrency. Modeled directly on `AutomationStore`. Includes an id-only `ids()` accessor for cheap listing that avoids cloning the (potentially sensitive) env/header maps a full definition carries.
- New **`McpServerDefinition` / `McpServerDraft` / `McpServerReplace`** domain model (plus `McpServerId` / `McpServerRevision` and structural validation) in `fabro-types`, reusing the existing `McpTransport`. These stay persistence-independent; the on-disk TOML DTO and the filesystem plumbing live in `fabro-mcp-store`.

Nothing in the workspace depends on the new crate yet. Wiring `McpServerStore` into the server, the HTTP API, and the UI are follow-up PRs.

## Testing

- `fabro-mcp-store`: 7/7 (empty/missing dir, non-TOML ignored, malformed/invalid-filename fail load, CRUD round-trip, stale-revision and duplicate-create rejected).
- `fabro-types`: `mcp_store` validation and round-trip tests pass. `cargo build --workspace`, fmt, and clippy all green.

## Notes

- The domain model derives `PartialEq` but not `Eq` because `McpTransport` carries `HashMap`s (differs from `Automation*`, matches the transport's capabilities).
- Validation is structural for now (id format, non-empty name, well-formed transport); credential-literal validation is deliberately deferred to the API layer (flagged TODO).
- The store is concrete by design (no trait): a future move off per-file TOML is a one-time migration, not a runtime backend choice. The revision is currently derived from the canonical TOML bytes — the one storage-coupled detail to revisit if that move happens.
- Part of a short series adding server-managed MCP servers; independent of the sibling PRs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2700,6 +2700,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "fabro-mcp-store"
+version = "0.267.0-nightly.0"
+dependencies = [
+ "fabro-types",
+ "serde",
+ "tempfile",
+ "thiserror 2.0.18",
+ "tokio",
+ "toml 0.8.23",
+ "tracing",
+]
+
+[[package]]
 name = "fabro-model"
 version = "0.267.0-nightly.0"
 dependencies = [
@@ -2848,6 +2861,7 @@ dependencies = [
  "fabro-llm",
  "fabro-macros",
  "fabro-manifest",
+ "fabro-mcp-store",
  "fabro-model",
  "fabro-proc",
  "fabro-redact",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2709,7 +2709,6 @@ dependencies = [
  "thiserror 2.0.18",
  "tokio",
  "toml 0.8.23",
- "tracing",
 ]
 
 [[package]]
@@ -2861,7 +2860,6 @@ dependencies = [
  "fabro-llm",
  "fabro-macros",
  "fabro-manifest",
- "fabro-mcp-store",
  "fabro-model",
  "fabro-proc",
  "fabro-redact",
--- a/lib/crates/fabro-mcp-store/Cargo.toml
+++ b/lib/crates/fabro-mcp-store/Cargo.toml
@@ -18,7 +18,6 @@ serde.workspace = true
 thiserror.workspace = true
 tokio.workspace = true
 toml.workspace = true
-tracing.workspace = true
 
 [dev-dependencies]
 tempfile = "3"
--- a/lib/crates/fabro-mcp-store/Cargo.toml
+++ b/lib/crates/fabro-mcp-store/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "fabro-mcp-store"
+edition.workspace = true
+version.workspace = true
+publish = false
+license.workspace = true
+description = "Server-managed MCP server catalog durable storage for Fabro"
+
+[lib]
+doctest = false
+
+[lints]
+workspace = true
+
+[dependencies]
+fabro-types = { path = "../fabro-types" }
+serde.workspace = true
+thiserror.workspace = true
+tokio.workspace = true
+toml.workspace = true
+tracing.workspace = true
+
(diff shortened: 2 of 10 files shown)


swerner:
## `/simplify` pass

Ran four parallel cleanup reviews (reuse, simplification, efficiency, altitude) over the diff. Applied one fix; skipped the rest with reasons. Build green, 7/7 tests pass.

### Applied

**`parse_persisted` — removed dead `Option<PathBuf>`** (`fabro-mcp-store/src/model.rs`). The argument was always `Some(path)` at the only call site, so the `None` / `"<memory>"` branches were unreachable. In the sibling automation crate that `Option` is live (its public `from_toml_bytes` passes `None`), but this crate never copied that in-memory constructor — so it inherited the dead branches without the caller that justifies them. Collapsed to `path: PathBuf`, dropping two redundant match arms. Behavior-preserving.

### Skipped (with reasons)

- **Extract `write_atomic` / `write_new` / `temp_path_for` / `id_from_path` / etc. into a shared util** — real ~3× duplication across `fabro-mcp-store` / `fabro-automation` / `fabro-environment`, but the fix means a new shared crate (generic over id + error types) plus editing two crates this PR doesn't touch. Out of scope here; better as its own follow-up PR now that there are three near-identical FS stores.
- **Unify `write_new` / `write_atomic`** — they differ only in the final link step (`rename` vs `hard_link` + cleanup). Parameterizing needs boxed async closures (more complex, not less) and would diverge from the sibling's identical pair.
- **Remove `replace.clone()` in `definition_from_replace`** — faithful mirror of automation's `draft.clone()`, on a cold admin path (create/replace are rare, small structs). The clean fix entangles the deliberately-separated type/validation model; not worth diverging from the mirror.
- **Generalize the three `*_dir_for_active_config` helpers** — only the new one is in-diff; collapsing edits two out-of-scope helpers, and three explicit one-liners is below the threshold where a `(path, subdir)` helper pays off.
- **`list()` / `ids()` clone-then-sort** — intentional: `ids()` deliberately avoids cloning sensitive env/header maps, sort is part of the contract, not a hot path.
- **Merge `PersistedMcpServer` / `McpServerReplace`** (field-identical) — intentional domain modeling (on-disk body vs replace payload), documented; a design choice, not a mechanical cleanup.
- **`#[expect(dead_code)]` accessor & model-in-`fabro-types`** — both intentional and documented (forcing function for the next PR; persistence-independent types).

### Worth a separate look

The shared-util extraction across the three FS stores is a legitimate follow-up — not done here to keep this PR scoped.

swerner:
Ran simplify on this PR.

Results:
- Removed unused eager server startup loading and AppState wiring for the MCP server store until a consumer lands.
- Moved the persisted TOML DTO out of `fabro-types` and into `fabro-mcp-store`, so shared types stay domain-focused.
- Precheck duplicate and stale mutations before TOML serialization/hash work, and share the temp-file write path inside the new store.
- Tightened MCP command validation to reject a blank program even when later args are present.
- Dropped unused `MissingRevision`, `tracing`, and server dependency surface, plus removed internal planning references from shared code comments.

Skipped as too broad for this pass:
- Cross-crate extraction of all file-store/id/revision helper patterns from automation, environment, and MCP store code.

Verification:
- `cargo check -p fabro-mcp-store`
- `cargo check -p fabro-server`
- `cargo nextest run -p fabro-types mcp_store`
- `cargo nextest run -p fabro-mcp-store`
- `cargo +nightly-2026-04-14 clippy -p fabro-mcp-store -p fabro-types --all-targets -- -D warnings`

PR #522, opened 2026-06-23
feat(api): add /api/v1/mcp-servers OpenAPI endpoints

swerner:
## What

Adds the HTTP contract for managing server-defined MCP servers. The handler implementation follows in a later change.

- New `/api/v1/mcp-servers` paths: `list`, `create`, `retrieve`, `replace`, `delete`, with ETag / `If-Match` optimistic concurrency mirroring the automations conventions.
- New schemas: `McpServer`, `CreateMcpServerRequest`, `ReplaceMcpServerRequest`, `McpServerListResponse`.
- **Collapsed a duplicate `McpTransport` schema** into the single canonical one and gave it a proper `discriminator` plus the previously-missing optional `protocol` field (`streamable_http` | `sse`). This also fixes a latent gap in the existing run-config projection and is non-breaking (`protocol` is `#[serde(default)]`).

## Testing

- `cargo build -p fabro-api` is green — progenitor generates the client methods and types cleanly from the new spec.

## Notes / follow-ups for the handler change

- Recommended `with_replacement` mapping (reuse, no parallel DTOs): `McpServer` → `McpServerDefinition`, create/replace → `McpServerDraft`/`McpServerReplace`, transport → existing `fabro_types::McpTransport`/`McpHttpProtocol`; list envelopes become small DTOs.
- Parity caveat: progenitor emits `i64` for the `u64` timeouts and `i32` for the `u16 port`; harmless under `with_replacement`, but the handler change must add identity/JSON-parity tests and not skip `with_replacement` for those types.
- `createMcpServer` returns ETag on 201 (Environments convention) so the UI gets the fresh revision.
- The "warn vs hard-reject credential-looking literal values" question is recorded in the request-schema descriptions and intentionally not enforced.
- Part of a short series adding server-managed MCP servers.

### DIFF
--- a/docs/public/api-reference/fabro-api.yaml
+++ b/docs/public/api-reference/fabro-api.yaml
@@ -19,6 +19,8 @@ tags:
     description: Server-managed automation definitions and automation-triggered runs
   - name: Environments
     description: Server-managed execution environment catalog
+  - name: MCP Servers
+    description: Server-managed MCP server definitions referenced by name from workflow configs
   - name: Sandboxes
     description: Provider-backed sandbox inventory
   - name: Sessions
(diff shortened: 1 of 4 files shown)


swerner:
**`/simplify` pass** (4 cleanup agents: reuse, simplification, efficiency, altitude). Pushed one cleanup commit — no behavior change, spec text/comment only.

### Fixed
1. **Stripped the deferred credential-handling design note from `CreateMcpServerRequest.description`** → reduced to the one-line factual sentence matching the sibling `CreateAutomationRequest`/`CreateEnvironmentRequest` schemas. These `description:` fields render into the published API reference, so an undecided, schema-unenforced rule was leaking into customer-facing docs.
2. **Same strip on `ReplaceMcpServerRequest.description`** → matches `ReplaceEnvironmentRequest` ("…The path id is authoritative.").
3. **Trimmed the free-floating transport-reuse comment** above `McpServerListResponse` — dropped the sentence that duplicated the now-expanded `McpTransport.description`, keeping the actionable point-of-use "do not fork a parallel transport schema" guardrail.

### Skipped (out of scope, flagging for awareness)
- **Type-ownership / `with_replacement` drift:** `McpTransport` / `McpServerSettings` / `McpHttpProtocol` generate parallel API types rather than reusing `fabro_types` via `lib/crates/fabro-api/build.rs`. This is **pre-existing** (not introduced here), the fix lives in `build.rs` + parity tests (outside this YAML-only PR), and it hinges on the projection-vs-reuse decision deferred to the handlers change. Worth deciding when the handlers land.

### Confirmed clean (no action)
- `McpTransport` discriminator is well-formed — `type` is `required` in all three branches and the enum values match the mapping keys.
- `protocol` / `McpHttpProtocol` placement matches `fabro_types::McpTransport` exactly (Http + Sandbox, not Stdio); shared via a single `$ref`.
- Exactly one `McpTransport` definition — no stale duplicate.
- The Create/Replace/resource schema triple and repeated per-field examples correctly match the file's full-inline house style (Automations/Environments do the same) — `allOf` factoring would diverge from convention.
- `McpServerId`, error responses, list-envelope/meta, `IfMatch`, and `ETag`/`x-request-id` all reuse the established shared components.

swerner:
Ran simplify on this PR (reuse, quality, efficiency). Fixed: MCP catalog reads now use value-omitting transport DTOs and display_name vs runtime id; write transport schemas are stricter; fabro-api now reuses canonical McpTransport/McpHttpProtocol with round-trip/type-identity tests. Validation: cargo +nightly-2026-04-14 fmt --check --all; cargo nextest run -p fabro-api. Pushed commit 049f60111.

PR #523, opened 2026-06-24
refactor(server): extract reusable git-checkout/materialization core

swerner:
## What

Foundational refactor toward running a workflow that lives in one repo against a *different* workspace repo (shared / external workflows). No public API surface and no behavior change for automations — it only reshapes internals behind a reusable seam.

- **New `git_checkout` module.** Lifts the git-clone + manifest-from-checkout machinery out of `automation_materializer`: `GitRepoCache` (cached bare clone + per-call worktree), the git command plans, credential resolution/redaction, and GitHub owner/repo slug parsing/validation. All `pub(crate)`; no module is exported.
- **Split the workflow source from the git context.** `build_manifest_from_checkout` now takes the *workflow-source checkout* (which workflow to bundle) and the *git context* (which repo the run clones and executes in) as separate inputs. Automations are the case where both coincide. This is the seam a future external-workflow resolver needs.
- **Decoupled the builder input.** `ManifestFromCheckoutInput` no longer embeds `AutomationRunMaterializeInput`; it takes only the fields it needs plus a caller-supplied error context, so it's reusable without automation-specific types.

## Review fixes folded in

- **Error type points the right way.** The shared materialize error moved into `git_checkout` as the provider-neutral `RunMaterializeError` (same variants, neutral messages). The foundation module no longer depends back on its consumer, and a bad workflow-source slug no longer reports "invalid automation target".
- **Required git context, not `Option`.** No caller omits it today; widening to optional later is backwards-compatible if a real case appears.

## Testing

- `cargo build -p fabro-server`, pinned-nightly `fmt --all` and `clippy -p fabro-server --all-targets -D warnings`: clean.
- `cargo nextest run -p fabro-server`: 729/732 pass. The 3 failures are graphviz SVG-render-subprocess tests (`get_graph_returns_svg`, `render_graph_from_manifest_*`) that fail identically on the clean baseline in this environment — pre-existing and unrelated.
- The rewritten unit test proves the split: a manifest built from a workflow-source checkout while `manifest.git` points at a *different* repo and ref.

### DIFF
--- a/lib/crates/fabro-automation/src/lib.rs
+++ b/lib/crates/fabro-automation/src/lib.rs
@@ -7,6 +7,7 @@ pub use error::{AutomationStoreError, AutomationValidationError};
 pub use id::{AutomationId, AutomationRevision, AutomationRevisionParseError, AutomationTriggerId};
 pub use model::{
     ApiTrigger, Automation, AutomationDraft, AutomationReplace, AutomationTarget,
-    AutomationTrigger, ScheduleTrigger, parse_schedule_expression,
+    AutomationTrigger, GitHubRepositorySlug, ScheduleTrigger, parse_github_repository_slug,
+    parse_schedule_expression,
 };
 pub use store::AutomationStore;
--- a/lib/crates/fabro-automation/src/model.rs
+++ b/lib/crates/fabro-automation/src/model.rs
@@ -137,6 +137,24 @@ pub struct AutomationTarget {
     pub workflow:     String,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct GitHubRepositorySlug {
+    owner: String,
+    repo:  String,
+}
+
+impl GitHubRepositorySlug {
+    #[must_use]
+    pub fn owner(&self) -> &str {
+        &self.owner
+    }
+
+    #[must_use]
+    pub fn repo(&self) -> &str {
+        &self.repo
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(tag = "type", rename_all = "snake_case", deny_unknown_fields)]
 pub enum AutomationTrigger {
@@ -273,7 +291,9 @@ fn validate_fields(value: &AutomationReplace) -> Result<(), AutomationValidation
     validate_triggers(&value.triggers)
 }
 
-fn validate_repository_slug(value: &str) -> Result<(), AutomationValidationError> {
+pub fn parse_github_repository_slug(
+    value: &str,
+) -> Result<GitHubRepositorySlug, AutomationValidationError> {
     let Some((owner, repo)) = value.split_once('/') else {
         return Err(AutomationValidationError::InvalidRepositorySlug {
             value: value.to_string(),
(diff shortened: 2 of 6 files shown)


swerner:
## `/simplify` pass (opus48)

Ran a 4-angle cleanup review (reuse / simplification / efficiency / altitude) over the extraction diff. Quality-only — no bug hunt. Applied the high-confidence wins; the rest are noted below for the deeper review rather than auto-applied.

### Applied — `0931abde3`
- **Dropped `ManifestFromCheckoutInput::serialize_error_context`.** It threaded a caller-formatted `"automation {id}"` string through the input struct purely to interpolate into one `with_context` message on the (effectively unreachable) manifest-serialize failure path. Removing it deletes a field + an eager allocation on the success path and keeps the last bit of automation vocabulary out of the now-reusable builder; the message is a neutral static string.
- **Two `.clone()` → moves** (`input.target.workflow`, `input.user_settings_path`). `input` is owned and dead after the struct is built, so the field-by-field construction no longer needs to clone — restores the no-clone behavior the old whole-struct move had.

Verified green: pinned-nightly `fmt`, `clippy -D warnings`, and the affected `fabro-server` unit tests.

### Deferred — for the deep-dive, not applied
These are either deliberate design choices, pre-existing, or cross-cutting enough that auto-applying would overstep:

- **Keep `ManifestGitContextInput` (did not flatten).** Simplification flagged it as ceremony for a single caller, but it's the deliberate "workflow-source vs git-context" seam this PR exists to create. Flattening would fight the intent.
- **Slug-validator duplication vs `fabro-automation`.** `git_checkout::{parse_github_repository_slug, valid_github_owner, valid_github_repo}` duplicate `fabro-automation`'s validators, and the two copies have already **drifted**: the `git_checkout` copy carries an extra `!value.starts_with('.')` rule the `fabro-automation` copy lacks. Pre-existing (just relocated by this move) and a cross-crate change, so out of scope here — but worth a single source of truth before the resolver reuses these for coordinate validation.
- **`RunMaterializeError` altitude.** (1) All four variants currently collapse to one `422` at the handler / a log line at the scheduler — no caller branches on them, so the taxonomy pays no rent yet. (2) The type is defined in `git_checkout`, but `WorkflowNotFound`/`Manifest` are only produced by the manifest builder in `automation_materializer.rs`, so the low-level git module names a concern above it. (3) The "reusable" builder still returns the automation-named `AutomationRunMaterialized`. These look like intentional first-step boundaries; calling them out so the resolver/handler phase can decide whether the variants should drive distinct 4xx/5xx (e.g. source-not-found vs workspace-not-accessible) and where the error type should live.

swerner:
Ran simplify on this PR.

Results:
- Launched the requested reuse, quality, and efficiency review agents against the full PR diff.
- Fixed the concrete reuse issue they found by moving GitHub owner/repo slug parsing into `fabro-automation` and reusing it from checkout materialization, eliminating the drift between target validation and materialization.
- Simplified the checkout API so production derives the clone URL from the validated repository slug instead of accepting both `repo` and `clone_url`.
- Moved the run-level materialization error taxonomy back to the automation boundary and kept checkout errors checkout-specific.
- Removed redundant optionality from the materializer git context input by carrying a concrete checked-out SHA.
- Efficiency review found no runtime issues to fix.

Validation:
- `cargo nextest run -p fabro-automation repository_slug_parser_returns_validated_parts`
- `cargo nextest run -p fabro-server git_checkout`
- `cargo nextest run -p fabro-server manifest_builder_uses_checkout_for_workflow_and_separate_git_context`
- `cargo nextest run -p fabro-automation validation_rejects_invalid_inputs`
- `cargo +nightly-2026-04-14 clippy -p fabro-automation -p fabro-server --all-targets -- -D warnings`

PR #524, opened 2026-06-24
Interpolate run variables ({{ vars.* }}) in node prompts and goals

swerner:
## What

Threads the run's variable store through the workflow transform pipeline so
node `prompt`s and the graph `goal` can interpolate `{{ vars.* }}`.

Until now `{{ vars.* }}` only resolved in settings-level fields (e.g.
`run.goal`) via the server-side `substitute_variables` pass. Node prompts are
DOT graph attributes that pass never touched, so `{{ vars.* }}` in a prompt
rendered as undefined. This closes that gap.

Builds on the earlier template-context slice (adds `vars` to
`TemplateContext`); this PR wires it end to end.

## How

- `TransformOptions` carries a `vars` map, threaded into the import,
  file-inlining, and template transforms — and propagated into imported
  subgraphs, so imported prompts interpolate vars too. Every prompt/goal render
  context gains the variable map.
- The create API accepts `vars` (`CreateRunInput` → `preprocess_and_validate` →
  `TransformOptions`).
- The server snapshots its `VariableStore` at run creation
  (`VariableStore::value_map()`) and passes it in — the same store the
  settings-goal substitution already reads.

## Scope decisions

- Goal `@file` contents interpolate vars too; **import paths stay inputs-only**
  (structural file resolution, conceptually outside the prompt/goal scope).
- Offline / CLI / `fabro validate` render with an empty var map, so
  `{{ vars.* }}` is undefined there: a warning at validate, a hard error at
  run-create — identical to how `inputs` behaves offline.

## Testing

- Transform-level: node-prompt and goal interpolation; unknown-var warning.
- Create-pipeline: vars resolve; an unknown var warns at validate and promotes
  to a hard error at run-create.
- End-to-end server test: `POST /variables` + `POST /runs`, asserting the
  rendered prompt in the persisted `run.created` event.

Verified: `cargo +nightly fmt --check`, nightly `clippy -D warnings` (including
the `test-support`-gated server integration binary), the tests above, and a
full-workspace `cargo check`.

### DIFF
--- a/lib/crates/fabro-server/src/run_manifest.rs
+++ b/lib/crates/fabro-server/src/run_manifest.rs
@@ -199,10 +199,12 @@ pub(crate) fn create_run_input(
     configured_providers: Vec<ProviderId>,
     provenance: RunProvenance,
     web_url: Option<String>,
+    vars: HashMap<String, String>,
 ) -> CreateRunInput {
     CreateRunInput {
         workflow: WorkflowInput::Bundled(prepared.workflow_input),
         settings: prepared.settings,
+        vars,
         cwd: prepared.cwd,
         workflow_slug: None,
         workflow_path: Some(prepared.target_path),
--- a/lib/crates/fabro-server/src/run_manifest.rs
+++ b/lib/crates/fabro-server/src/run_manifest.rs
@@ -185,9 +185,18 @@ pub(crate) fn validate_prepared_manifest(
     prepared: &PreparedManifest,
     catalog: Arc<Catalog>,
 ) -> Result<Validated, WorkflowError> {
+    validate_prepared_manifest_with_vars(prepared, catalog, HashMap::new())
+}
+
+pub(crate) fn validate_prepared_manifest_with_vars(
+    prepared: &PreparedManifest,
+    catalog: Arc<Catalog>,
+    vars: HashMap<String, String>,
+) -> Result<Validated, WorkflowError> {
     validate(ValidateInput {
         workflow: WorkflowInput::Bundled(prepared.workflow_input.clone()),
         settings: prepared.settings.clone(),
+        vars,
         cwd: prepared.cwd.clone(),
         custom_transforms: Vec::new(),
         catalog,
--- a/lib/crates/fabro-server/src/server/handler/runs.rs
+++ b/lib/crates/fabro-server/src/server/handler/runs.rs
@@ -1,4 +1,4 @@
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 use std::io::ErrorKind;
 use std::sync::Arc;
 
@@ -645,7 +645,8 @@ pub(crate) async fn create_run_from_manifest(
         Ok(prepared) => prepared,
         Err(err) => return ApiError::bad_request(err.to_string()).into_response(),
     };
-    if let Err(err) = substitute_run_variables(&state, &mut prepared.settings).await {
+    let vars = snapshot_run_variables(&state).await;
+    if let Err(err) = substitute_run_variables(&vars, &mut prepared.settings) {
         return ApiError::bad_request(format!("Run config variable interpolation failed: {err}"))
             .into_response();
     }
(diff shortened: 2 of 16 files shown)


swerner:
## `/simplify` pass — no quality fixes applied, code is already clean

Ran a four-angle cleanup review (reuse / simplification / efficiency / altitude). Three angles found nothing actionable; the one substantive flag is a behavior question, not a cleanup, so I left it untouched. Recording each candidate and why it was skipped:

- **Reuse — `for_input_scan(...).with_vars(...)` → a `for_input_scan_with_vars()` constructor.** Skipped. The chain appears at exactly one site (`file_inlining.rs:261`). A dedicated constructor adds API surface to save one call rather than reducing complexity; the `with_inputs`/`with_vars` builder parallel is the idiomatic shape.
- **Simplification — none.** The clones across the import/file-inlining/template passes are each consumed by owned-value builders, and the `vars: _` destructure in `persist_validated` is the correct way to discard a field that's already baked into the rendered graph. Nothing redundant or dead.
- **Efficiency — thread `Arc<HashMap>` instead of cloning the vars map (~8 clones/run).** Skipped. Cold path (once per run-create, tiny maps), so the win is negligible — and `inputs` is threaded by plain clone right alongside `vars`. Making only `vars` an `Arc` would break that symmetry and worsen consistency. If ever worth doing, do it to both fields together as its own change.
- **Altitude — `vars`/`inputs` as parallel fields, empty-map offline sentinel, deep test assertion.** All confirmed right altitude: matches the established `inputs` pattern; `HashMap::new()` and `None` behave identically in the render context so the sentinel is fine; the deep JSON assertion is acceptable for an e2e test.

### Note: a narrow edge case worth a separate look (behavior, not cleanup)

At `import.rs:687`, `path_ctx = TemplateContext::for_input_scan(self.inputs.clone())` (no `.with_vars()`) renders both import paths **and** the `parent_goal` that's passed down as the goal-override into imported subgraphs (`file_inlining.rs:188`, `resolved_goal = goal_override`). Consequence:

- An imported prompt with a direct `{{ vars.X }}` → **resolves** (child `FileInlining` has `.with_vars`). ✓
- An imported prompt referencing `{{ goal }}`, where the parent goal contains `{{ vars.X }}` → sees the **vars-stripped** goal, because the override was rendered at line 687 without vars. ✗

The top-level path doesn't have this gap (`file_inlining.rs:261` adds `.with_vars`, covered by `transform_interpolates_vars_in_graph_goal_and_through_prompt`). So it's narrow: vars embedded in a *goal* that an *imported* prompt re-embeds via `{{ goal }}`.

Left alone here because (a) `/simplify` is quality-only and the fix changes rendered output, (b) line 687 is pre-existing code outside this PR's diff, and (c) it brushes against the stated scope ("import paths stay inputs-only") — though the goal-override case is arguably distinct from import *paths*. Flagging for a `/code-review` decision: either accept it as intended scope, or add the one-line `.with_vars()` plus a regression test for the imported-goal case.

swerner:
Ran simplify on PR 524 and pushed follow-up cleanup.

Changes made:
- Consolidated prompt and goal render bindings into a reusable TemplateContext so inputs and vars are threaded consistently.
- Fixed imported workflow parent-goal rendering so vars resolve before imported prompts use the parent goal.
- Snapshot server variables once for settings interpolation, run creation, preflight, and validation to avoid duplicate reads and snapshot drift.
- Precompute MiniJinja context values to avoid repeatedly cloning and serializing variable maps per render.
- Added regression coverage for imported prompts and run validation.

Verification:
- cargo check -p fabro-workflow -p fabro-server
- cargo nextest run -p fabro-template
- cargo nextest run -p fabro-workflow -- import_parent_goal_resolves_vars_before_imported_prompt_uses_goal
- cargo nextest run -p fabro-workflow -- transform_interpolates_vars
- cargo nextest run -p fabro-server --features test-support --test it -- run_validate_resolves_variables_in_node_prompts
- cargo nextest run -p fabro-server --features test-support --test it -- run_create_interpolates_variables_into_node_prompts

PR #526, opened 2026-06-25
feat(types): introduce ResolvedMcpEntry for run agent MCP entries

swerner:
## What

Changes `RunAgentSettings.mcps` from `HashMap<String, McpServerSettings>` to `HashMap<String, ResolvedMcpEntry>`, a two-state enum:

- `Resolved(McpServerSettings)` — an inline, fully-resolved MCP server (every code path produces this today).
- `Reference { id, enabled }` — an unresolved reference to a named server in the MCP catalog.

This is the **type-shape foundation only**: every current path still produces `Resolved`, and no reference parsing or catalog lookup is added here. It unblocks a later server-side pass that swaps `Reference` → `Resolved` against the MCP server store before a run spec is persisted, so persisted runs stay self-contained snapshots.

## Why this shape

- `ResolvedMcpEntry` is `#[serde(untagged)]` with `Resolved` first, so a resolved entry (de)serializes as a bare `McpServerSettings` with no enum tag — preserving backward compatibility with run specs persisted before the enum existed.
- `McpServerRef` uses `deny_unknown_fields`, so the two variants can never collide (`McpServerSettings` requires `name` + `transport`, which a reference rejects).
- `McpServerRef.id` is a plain `String`, keeping `fabro-types` decoupled from the MCP store crate.

## Consumers updated

- **fabro-config** `resolve_agent`: wraps each enabled inline entry as `Resolved`, reusing the shared `resolve_enabled_mcps` enable-filter.
- **fabro-types** `RunNamespace::substitute_variables`: only walks `Resolved` entries (references carry no templates).
- **fabro-workflow** `operations/start.rs`: extracts `Resolved` at the post-persistence worker-startup consumer; a surviving `Reference` is an invariant violation, guarded with `debug_assert!` plus a hard error.
- **fabro-cli** `exec.rs`: the `run.agent.mcps` fallback for `fabro exec` keeps only `Resolved` inline servers; catalog references are run-only on this CLI-direct path (no server-side resolver).

## Tests

- Back-compat round-trip proving old-format bare-`McpServerSettings` maps (JSON and TOML) deserialize as all-`Resolved`.
- A `{ id, enabled }` value parses as `Reference` while a full server config parses as `Resolved`.
- `Resolved` serializes back out as a bare `McpServerSettings`.

Independent of the in-flight MCP server store and OpenAPI-spec PRs; mergeable on its own.

### DIFF
--- a/lib/crates/fabro-cli/src/commands/exec.rs
+++ b/lib/crates/fabro-cli/src/commands/exec.rs
@@ -17,6 +17,7 @@ use fabro_llm::types::{
 use fabro_mcp::config::McpServerSettings;
 use fabro_model::ProviderId;
 use fabro_types::settings::cli::OutputFormat as SettingsOutputFormat;
+use fabro_types::settings::run::ResolvedMcpEntry;
 use fabro_util::exit::{self, ErrorExt, ExitClass};
 use futures::stream;
 use serde::Deserialize;
@@ -317,7 +318,20 @@ pub(crate) async fn execute(mut args: ExecArgs, ctx: &CommandContext) -> AnyResu
         Some(mcps) => mcps.values().cloned().collect(),
         None => ctx
             .run_settings()
-            .map(|settings| settings.agent.mcps.values().cloned().collect())
+            .map(|settings| {
+                settings
+                    .agent
+                    .mcps
+                    .values()
+                    // `fabro exec` is a CLI-direct path with no server-side
+                    // catalog resolver, so it only honors inline resolved
+                    // servers; unresolved catalog references are run-only.
+                    .filter_map(|entry| match entry {
+                        ResolvedMcpEntry::Resolved(server) => Some(server.clone()),
+                        ResolvedMcpEntry::Reference(_) => None,
+                    })
+                    .collect()
+            })
             .unwrap_or_default(),
     };
     // Resolve `{{ env.* }}` in MCP transport config at the exec boundary,
--- a/lib/crates/fabro-cli/src/commands/exec.rs
+++ b/lib/crates/fabro-cli/src/commands/exec.rs
@@ -326,10 +326,8 @@ pub(crate) async fn execute(mut args: ExecArgs, ctx: &CommandContext) -> AnyResu
                     // `fabro exec` is a CLI-direct path with no server-side
                     // catalog resolver, so it only honors inline resolved
                     // servers; unresolved catalog references are run-only.
-                    .filter_map(|entry| match entry {
-                        ResolvedMcpEntry::Resolved(server) => Some(server.clone()),
-                        ResolvedMcpEntry::Reference(_) => None,
-                    })
+                    .filter_map(ResolvedMcpEntry::as_resolved)
+                    .cloned()
                     .collect()
             })
             .unwrap_or_default(),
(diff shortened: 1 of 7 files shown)


swerner:
## `/simplify` (Opus 4.8)

Four cleanup agents (reuse / simplification / efficiency / altitude) reviewed the diff. All four converged on the same core finding, applied in `6521c4111`. Quality-only pass — no correctness review.

### Fixed

The diff open-coded the same `match { Resolved(s) => …, Reference(_) => … }` extraction at every consumer. Added one accessor on the type that owns the invariant and routed call sites through it, matching the existing `AttrValue::as_str`/`as_bool` convention in `fabro-types/src/graph.rs`:

- **`fabro-types/src/settings/run.rs`** — added `ResolvedMcpEntry::as_resolved(&self) -> Option<&McpServerSettings>` (`#[must_use]`).
- **`fabro-cli/src/commands/exec.rs`** — `filter_map(|entry| match …)` → `.filter_map(ResolvedMcpEntry::as_resolved).cloned()`.
- **`fabro-config/src/resolve/mod.rs`** (test) — the `transport` closure collapses to `get(name).and_then(ResolvedMcpEntry::as_resolved).map(…)`.
- **`fabro-config/src/tests/resolve_run.rs`** — `stdio_command` helper drops a level of match nesting and its bespoke `Reference` panic arm.
- **`fabro-types/src/settings/run.rs`** (test) — `let …Resolved(mcp) = … else { panic! }` → `as_resolved().expect(…)`.
- **`fabro-workflow/src/operations/start.rs`** — the `Reference` arm carried two near-identical message strings (one for `debug_assert!`, one for `Err`); build the message once and reuse it. Behavior unchanged.

Net **−29 / +30** lines, with the duplication consolidated into one place. `build`, `fmt --check`, and `clippy -D warnings` pass; the full `fabro-types` + `fabro-config` + `fabro-cli` suite (1516 tests) is green.

### Skipped (noted, not applied)

- **`McpServerRef.enabled` field** (altitude) — flagged as pre-committing an enablement decision, but it's a deliberate, documented part of the type shape and belongs to the not-yet-built resolver. Removing it is a design change, not a cleanup.
- **Double-collect in `resolve_agent`** (efficiency) — it *moves* (not clones) values, and folding the wrap into the shared `resolve_enabled_mcps` would just push the cost onto the `cli.exec` caller. No change.
- **A shared `expect_resolved(key)` invariant helper on the type** (altitude) — borderline with only one post-persistence consumer today; the lighter message-dedup above captured the simplification without adding speculative API surface.

### Note

`pipeline::initialize::tests::initialize_prepares_sandbox_and_uses_persisted_run_dir` fails locally, but it fails identically on the unmodified PR HEAD (verified via stash). It's environment-driven (ambient LLM credentials in the shell) and unrelated to this PR.

swerner:
Ran simplify (gpt55xhigh). Review agents: reuse review found no missed helpers; quality review found and I fixed MCP reference ambiguity, fabro exec silent reference dropping, and debug-only worker panic behavior; efficiency review found and I fixed the extra MCP map allocation in run config resolution. Validation: cargo nextest run -p fabro-types -- run_agent_settings_tests; cargo nextest run -p fabro-cli -- commands::exec::tests; cargo nextest run -p fabro-config -- run_agent_mcps; cargo +nightly-2026-04-14 fmt --check --all; git diff --check; cargo +nightly-2026-04-14 clippy -p fabro-types -p fabro-config -p fabro-cli -p fabro-workflow --all-targets -- -D warnings.

PR #527, opened 2026-06-25
Unify @file inlining under an ImportableTemplate type (prompt + goal)

swerner:
## What

Introduces an `ImportableTemplate` type that unifies the "inline content **or**
`@path` file import" concept used by node `prompt`s, the graph `goal`, and
`output_schema`. This is the last template-side piece of the interpolation
unification: a single named type now owns the `@`-classification and
static-reference validation that was previously hand-rolled in three places.

This is a **behavior-preserving refactor** — no user-visible change.

## How

- New `ImportableTemplate { Inline(String), Import { path } }` in
  `transforms/importable_template.rs`, with `parse` (classifies a value — a
  leading `@` marks a file import), `import_path`, and `validate` (rejects
  template syntax in an import path). Callers of templated fields classify the
  **already-rendered** string, because a leading `@` can be produced by
  rendering (e.g. `{{ inputs.prompt_file }}` → `@prompts/work.md`).
- `prompt` + `goal`: render the inline value, then — if it's an `@file` import —
  load and render the file contents via the type. The missing-file → literal
  passthrough is preserved.
- `output_schema`: shares the same classification but is loaded **verbatim** (it
  is intentionally not a template), keeping its hard-error-on-missing-file
  behavior.
- Deletes the dead `resolve_file_ref` helper (no non-test callers) and inlines
  the trivial `render_file_contents` wrapper.
- Migrates the `FilesystemFileResolver` coverage (tilde, `..`, fallback-dir
  precedence, missing file) — which previously only existed through
  `resolve_file_ref`'s tests — onto direct `file_resolver` tests.

`TemplateTransform` and the import transform are untouched, so goal-before-
prompts ordering and the goal-self-reference guard are preserved exactly.

## Scope

Covers the DOT node `prompt` + graph `goal` `@file` path. The settings-layer
`run.goal` resolution is intentionally left as-is — it uses a different model
(interpolates env into the file path and does not render file contents), so
folding it in would be a semantic change, not a refactor. That convergence can
be a deliberate follow-up.

## Testing

- `cargo nextest run -p fabro-workflow` — 1182 passed (31 e2e/credentialed
  skipped). New unit tests on the type (classification, validation) and the
  migrated `FilesystemFileResolver` tests.
- Regression net kept green: file-inlining (prompt/goal, output_schema
  verbatim/error/routing, `{% include %}` rooting, fallback dir), the
  `TemplateTransform` goal/self-reference/ordering tests, and the cross-pass
  `reports_goal_self_reference_once_across_passes`.
- `cargo +nightly fmt --check --all` and nightly
  `clippy --workspace --all-targets -- -D warnings` clean.

### DIFF
--- a/lib/crates/fabro-cli/src/commands/exec.rs
+++ b/lib/crates/fabro-cli/src/commands/exec.rs
@@ -324,7 +324,7 @@ pub(crate) async fn execute(mut args: ExecArgs, ctx: &CommandContext) -> AnyResu
     // against the CLI process env — the mirror of the `fabro run` worker
     // boundary in `fabro_workflow::operations::start::runtime_mcp_server`.
     // Both consumers read the same source-form settings; missing env is a hard
-    // error (D3) and reserved secrets/inputs tokens surface loudly rather than
+    // error and reserved secrets/inputs tokens surface loudly rather than
     // leaking.
     let mcp_servers = mcp_servers
         .into_iter()
--- a/lib/crates/fabro-config/src/resolve/run.rs
+++ b/lib/crates/fabro-config/src/resolve/run.rs
@@ -154,7 +154,7 @@ fn resolve_git(git: Option<&RunGitLayer>) -> RunGitSettings {
 #[expect(
     clippy::disallowed_methods,
     reason = "known leak: prepare step templates collapse to raw source unresolved; strict \
-              resolution scheduled in the interpolation unification (Phase 2)"
+              resolution scheduled for follow-up interpolation cleanup"
 )]
 fn resolve_prepare(
     prepare: Option<&RunPrepareLayer>,
--- a/lib/crates/fabro-config/src/tests/resolve_run.rs
+++ b/lib/crates/fabro-config/src/tests/resolve_run.rs
@@ -647,7 +647,7 @@ name = "sonnet"
         }
         other => panic!("expected file goal, got {other:?}"),
     }
-    // run.working_dir is demoted (D11): the env token stays literal text.
+    // run.working_dir is demoted: the env token stays literal text.
     assert_eq!(
         settings.working_dir.as_deref(),
         Some("{{ env.FABRO_WORKDIR }}")
--- a/lib/crates/fabro-server/src/automation_materializer.rs
+++ b/lib/crates/fabro-server/src/automation_materializer.rs
@@ -13,7 +13,7 @@ use tokio::{fs, task};
 
 use crate::git_checkout::{
     GitCheckoutError, GitRepoCache, WorktreePrepareInput, github_metadata_url,
-    parse_github_repository_slug, resolve_git_auth_config,
+    resolve_git_auth_config,
 };
 
 #[derive(Debug, Clone, PartialEq, Eq)]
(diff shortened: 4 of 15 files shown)


swerner:
## `/simplify` cleanup pass

Ran four parallel cleanup reviewers (reuse / simplification / efficiency / altitude) over the diff. Three of them independently converged on a single mechanism in `ImportableTemplate`, fixed in one coherent change.

### Fixed

**`ImportableTemplate` carried dead state, allocated on the hot path, and forced a fragile three-call dance** — `transforms/importable_template.rs` + both call sites in `transforms/file_inlining.rs`:

- **Dead state:** `Inline(String)` stored a clone of the input that no call site ever read — both consumers fall back to the original `rendered`/`value`. `Inline` is now a unit variant.
- **Wasted allocations:** `parse` did `to_string()` for *every* variant on *every* call, and it runs once per node `prompt` + graph `goal` + `output_schema` during the transform pass — so an inline-heavy workflow paid N+1 throwaway heap allocations. The enum now borrows (`ImportableTemplate<'a>`, `Import { path: &'a str }`); `parse` allocates nothing.
- **Fragile parse→validate→extract ceremony:** callers had to remember to call `validate()` after `parse()`, and `parse` never failed — so forgetting validation failed silently. Validation is now folded into `import_path() -> Result<Option<&str>, Error>`, so a path can't be extracted without being validated. Each call site collapsed from 4 lines to 1:

```rust
let Some(path) = ImportableTemplate::parse(rendered).import_path()? else { ... };
```

Module tests updated to the new API. `cargo build`, `nextest` (affected tests), nightly `clippy`, and `fmt --check` all green.

### Skipped (deliberately)

- **Extend the type to also own `Import`/`GraphGoalFile` reference kinds** (import.rs, variable_expansion.rs): out of scope. The module doc intentionally scopes this type to `prompt`/`goal`/`output_schema`; unifying validation across the three layers (each with a different reference kind) is a design change, not a cleanup.
- **`import.rs` still uses `prompt.starts_with('@')`:** that's a one-off error-reporting heuristic, not classification logic. Routing it through `ImportableTemplate` would add coupling for no real gain.
- **Reuse / test-migration findings:** no pre-existing inline-vs-import type to reuse, and the test relocation from `file_inlining.rs` to `file_resolver.rs` moves coverage closer to the implementation — not duplication.

swerner:
Ran simplify.

Fixed in `fa62f0e0b`:
- Removed internal tracking IDs from shared comments.
- Kept run-variable render state out of persistence options by passing a `TemplateContext` through preprocessing.
- Removed unused transform-level `with_vars` forwarding methods.
- Renamed `ImportableTemplate` to `ImportableField` because the abstraction also covers verbatim fields like `output_schema`.
- Simplified GitHub repository slug parsing so the materializer maps the automation parser error directly.
- Tightened git checkout cache behavior: worktrees now target the resolved SHA, stale worktree pruning only happens on retry, corruption detection avoids sync filesystem checks, and corrupt-cache cleanup failures are logged.

Validation:
- `git diff --check`
- `cargo nextest run -p fabro-server git_checkout`
- `cargo nextest run -p fabro-server automation_materializer`
- `cargo nextest run -p fabro-workflow operations::create`
- `cargo nextest run -p fabro-workflow transforms::file_inlining`
- `cargo nextest run -p fabro-workflow importable_field`
- `cargo nextest run -p fabro-config resolve_run resolve_cli`
- `cargo nextest run -p fabro-server --features test-support variables` (passed with filesystem access; sandboxed run failed only because it could not write `~/.fabro/storage`)
- `cargo +nightly-2026-04-14 clippy -p fabro-server -p fabro-workflow --all-targets -- -D warnings`

PR #528, opened 2026-06-25
Hooks: typed end-to-end interpolation, narrow header tokens, fail-closed resolution

swerner:
## What

Makes hook interpolation typed end-to-end and fail-closed, and removes the bespoke template engine on HTTP-hook headers.

- **Typed end-to-end.** Hook `command`, `url`, header values, `prompt`, and `model` are now carried as a typed `InterpString` from the config resolve layer all the way to the executor. The executor resolves each segment at hook fire time from the typed value instead of collapsing it to a `String` and re-parsing it. This mirrors the MCP transport env resolution boundary (`resolve_transport_env` / `runtime_mcp_server`).
- **Narrow header tokens.** HTTP-hook headers previously ran through MiniJinja with an env allowlist (`TemplateContext::with_env_lookup_allowed`). They now resolve through the same narrow `{{ ns.NAME }}` token resolver as every other hook field — no template engine, no allowlist.
- **Fail-closed everywhere.** A missing or out-of-scope `{{ env.* }}` / `{{ secrets.* }}` token in a command, URL, header, prompt, or model is now a hard error that blocks the hook rather than firing it with a half-resolved or empty value. Previously command hooks failed closed but http/prompt/agent hooks failed open (warned and proceeded), which could dispatch an HTTP request with an empty credential header or run an LLM call against a half-rendered prompt. Transport-level outcomes (non-2xx responses, connection errors, unparseable bodies) stay fail-open.

A follow-up cleanup commit removes the template engine's `env` namespace (`with_env_lookup` / `with_env_lookup_allowed` / the `EnvLookup` object), which the header path was the last consumer of.

## How

- `fabro-types` and `fabro-hooks` `HookType` / `HookDefinition` now type the interpolatable fields as `InterpString`. `InterpString` serializes as its raw source, so persisted run specs and checkpoints round-trip unchanged.
- The `fabro-config` resolve layer clones the typed `InterpString` through instead of calling `as_source()`, so the fields no longer leak unresolved template text — the old "source preservation" `#[expect]` annotations on the hook resolvers are gone.
- The executor's single `resolve_interp` helper resolves a typed `InterpString` and is shared by the command, http, prompt, and agent paths; resolution failure maps to `HookDecision::Block`, which the runner already reports loudly (error for blocking hooks, warn for non-blocking).

## Testing

- New unit tests: fire-time resolution from the typed value (no re-parse), narrow-token header resolution, and fail-closed behavior for HTTP url, HTTP header, and prompt hooks on a missing variable (the hook does not fire and the resolution error surfaces).
- Existing hook tests updated and kept green.
- Gates: `cargo build --workspace`, `cargo +nightly-2026-04-14 fmt --check --all`, `cargo +nightly-2026-04-14 clippy --workspace --all-targets -- -D warnings`, and `cargo nextest run` for the touched crates (`fabro-hooks`, `fabro-types`, `fabro-config`, `fabro-template`, `fabro-workflow`, `fabro-server`, and the `fabro-cli` hook/config tests), all green.


swerner:
Addressed the review-blocking regression plus the smaller nits.

**Must-fix: `allowed_env_vars` is enforced again (fail-closed).**
The typed-interpolation rewrite had dropped the header env allowlist — headers resolved any `{{ env.* }}` while `allowed_env_vars` was parsed/carried/documented but bound to `_` in the executor. Header values now resolve through a scoped env lookup that returns a value only for names in the hook's `allowed_env_vars` and `None` otherwise. A non-allowlisted `{{ env.X }}` therefore resolves as absent (Missing) and the hook **blocks before firing** — it never sends a half-resolved or empty credential header. Semantics match the previous template-based behavior: an empty `allowed_env_vars` permits no env vars in headers. This keeps the deleted `fabro-template` env API deleted (no revival). The allowlist applies only to headers; url/command/prompt/model keep no allowlist.

Tests cover both directions:
- `header_resolves_allowlisted_var` / `http_hook_sends_interpolated_headers` — allowlisted var resolves and the header is sent.
- `header_rejects_unlisted_var` and `http_hook_unlisted_header_var_blocks_without_firing` — a var set in the process but absent from `allowed_env_vars` blocks; the mock endpoint is asserted to be called 0 times.
- `http_hook_missing_header_token_blocks_without_firing` — allowlisted-but-unset still blocks.

**Nits:**
- Reworded the resolve-helper doc comments: only `env` is wired here; `secrets`/`vars`/`inputs` are unavailable and fail closed.
- Header/URL resolution-failure logs now emit the unresolved token source (`url.as_source()`) instead of the resolved URL, so a URL-embedded credential cannot leak into logs.
- Docs reconciled: `agents/hooks.mdx` now documents `{{ env.NAME }}` interpolation and that `allowed_env_vars` scopes which env vars a header may read; added matching descriptions to the OpenAPI `headers`/`allowed_env_vars` fields.

Gates: workspace build, fmt, clippy (fabro-hooks clean), and tests for `fabro-hooks`, `fabro-workflow`, `fabro-server`, `fabro-api`, `fabro-config`, `fabro-types` all green.

swerner:
Simplify pass run.

Fixed:
- Reused shared fabro_types hook definitions from fabro_hooks and removed the duplicate runtime hook model plus startup mapping shim.
- Moved hook event display/default-blocking behavior to the shared type, fixing effective_name drift.
- Removed stale fabro-hooks dependencies on fabro-config/fabro-template; added fabro-redact for safe hook URL logging.
- Made HTTP hook transport logs use a redacted unresolved URL source and distinguish disallowed header env vars from unset allowlisted vars.
- Narrowed executor env bounds.

Reviewed but left unchanged:
- The argv-style hook command join predates this PR and a proper fix would require carrying argv shape to runtime resolution/quoting, so I left it for a separate targeted change.

Validation:
- cargo check -p fabro-hooks
- cargo check -p fabro-workflow
- cargo nextest run -p fabro-hooks (rerun with host networking because sandbox denied httpmock port binding): 73 passed
- cargo nextest run -p fabro-workflow hook_config_merge_run_overrides_by_name
- cargo nextest run -p fabro-hooks safe_url_source_for_log header_missing_token_is_hard_error header_rejects_unlisted_var
- cargo +nightly-2026-04-14 clippy -p fabro-hooks -p fabro-workflow --all-targets -- -D warnings

swerner:
## `/simplify` (Opus 4.8)

Ran `/simplify` over this PR — four parallel cleanup agents (reuse, simplification, efficiency, altitude). Overall the PR is **well-built**: the core (carrying `InterpString` end-to-end, fail-closed resolution, and the dedup that deletes `runtime_hook_definition` / collapses `config.rs` to a re-export) all landed at the right altitude. One contained fix applied; the rest are noted as out-of-scope follow-ups.

### Applied (1) — `a08021d5`

**Reuse:** `safe_url_source_for_log` hand-rolled the `DisplaySafeUrl::parse(…).map_or_else(…redacted_string())` parse-or-placeholder dance that already exists verbatim in ~4 other crates (`fabro-llm`, `fabro-server` ×2, `fabro-github`), while `fabro-redact` — which *owns* `DisplaySafeUrl` and which the hook executor newly depends on — exposed no shared helper.

- Added `fabro_redact::redacted_url_for_log(&str) -> String` (with tests) next to `DisplaySafeUrl`.
- Reduced the hook helper to a one-line adapter: `redacted_url_for_log(&url.as_source())`.
- Unified the unparseable-source placeholder to `"<invalid url>"` to match the existing call sites.

The 4 pre-existing copies in unrelated crates were intentionally **not** migrated (that's a repo-wide consolidation that shouldn't ride on a hooks PR), but the shared helper now exists for them to adopt.

Verified: `fabro-redact` 135/135 and `fabro-hooks` 75/75 tests pass; clippy + rustfmt clean on both touched files.

### Skipped — with reasons

- **`HeaderResolveError` / header allowlist pre-scan** (flagged by simplification, altitude, efficiency): the clean generalization is teaching the shared `ResolveCtx` an env-allowlist plus a `ResolveErrorKind::NotAllowed` variant — a change to `InterpString`/`interp.rs`, **outside this PR's diff**. The lighter inline-closure alternative would discard the distinct *"not in `allowed_env_vars`"* error this PR intentionally introduced. → follow-up.
- **`join_command` re-serialize-and-reparse** (`resolve/run.rs`): would want a segment-level `InterpString` join/concat constructor (shared-API change); it's a cold config-resolve path and already documented with an `#[expect]`. → follow-up.
- **Promote `resolve_interp` to a method on `InterpString`**: a follow-up already flagged inside `interp.rs` itself (the MCP transport boundary still does the old `String → InterpString` re-parse this PR escapes for hooks). → follow-up.
- **Hoisting the repeated `safe_url_source_for_log(url)` log-site calls**: **false positive** — hoisting would eagerly redact on every happy-path hook fire, defeating the current per-branch laziness.

*Generated by `/simplify` with Claude Opus 4.8.*

PR #530, opened 2026-06-25
Resolve run.prepare.steps env and interpolation at the run boundary

swerner:
## What

Per-step environment in `run.prepare.steps[].env` was parsed and then **dropped** before it reached the resolved run settings, so prepare steps could never see their declared env. This PR carries that env all the way through to the executor, resolves prepare-step interpolation at the run boundary, and fixes an argv-quoting bug.

Three things:

1. **Per-step env is carried through.** `RunPrepareSettings` now holds `steps: Vec<PreparedStep>` (command plus per-step `env`) instead of a flat `commands: Vec<String>`. The per-step env reaches `exec_command`, which already accepts per-command env vars, and is merged on top of the base sandbox environment.
2. **Interpolation resolves at the run boundary.** Prepare-step `script`/`command` and per-step `env` values are carried in source form out of the portable config resolve layer (so `fabro validate` stays portable and never requires env to be set). Their `{{ env.* }}` tokens resolve in the process that actually runs the steps, via `RunPrepareSettings::resolve_step_env` — mirroring the existing MCP transport env resolution. A missing env var is a **hard error** (fail-closed); there is no fallback to the unresolved literal.
3. **Argv is shell-quoted.** Argv-style prepare steps were assembled with `join(" ")`, so an argument containing spaces or quotes was re-split by the shell. They are now shell-quoted per element with the shared `shell_quote()` helper. `script` steps stay verbatim because they are raw shell snippets.

## How

- `RunPrepareSettings.commands: Vec<String>` becomes `RunPrepareSettings.steps: Vec<PreparedStep>` where `PreparedStep { command, env }`. The server-side `{{ vars.* }}` substitution pass now walks each step's command and env.
- New `RunPrepareSettings::resolve_step_env(env_lookup)` resolves `{{ env.* }}` in each step's command and env values, returning a hard error on a missing var (and a loud `Unavailable` error for reserved `secrets`/`inputs` tokens).
- The run boundary (`fabro_workflow::operations::start`) gains `runtime_setup_commands`, the prepare-step counterpart to `runtime_mcp_server`. `LifecycleOptions` now carries `Vec<SetupCommand>` (command + env), and the initialize phase passes each step's env to `exec_command`.
- `resolve_prepare` shell-quotes each argv element and carries per-step env in source form. The stale lint suppression on the resolved fields is rewritten to describe the deliberate source preservation that now resolves at the run boundary.
- The shell-quoting helper moves to a shared `fabro_util::shell` module (backed by `shlex`); `fabro_sandbox::shell_quote` delegates to it so the config resolve layer and sandbox code share one audited implementation.
- The OpenAPI `RunPrepareSettings` schema and the generated TypeScript client are updated to the new `steps`/`PreparedStep` shape.

## Testing

- `cargo build --workspace`
- `cargo nextest run` for `fabro-util`, `fabro-types`, `fabro-config`, `fabro-sandbox`, `fabro-api`, `fabro-workflow`, `fabro-server`, `fabro-cli` (provider keys stripped) — all green.
- `cd lib/packages/fabro-api-client && bun run typecheck` — clean.

New tests cover: per-step env carried through resolution; script/command + env resolved at the run boundary; a missing env var is a hard error (in both the command and a per-step env value); reserved `secrets` tokens surface as `Unavailable`; argv elements are shell-quoted (an arg with spaces/quotes is correctly quoted) while a `script` stays verbatim; and an end-to-end check that per-step env reaches the executed setup command (with a negative control proving the success is attributable to the per-step env).


swerner:
⚠️ **Do not merge — confirmed command-injection bug found in review.**

An adversarial shell-safety review reproduced a command injection in the argv `command` path:

1. `resolve_prepare` shell-quotes each argv element **in source form** at config time, so `{{ env.USER_INPUT }}` becomes the quoted token `'{{ env.USER_INPUT }}'`.
2. At the run boundary, `resolve_step_env` substitutes the **raw** env value *inside* those quotes (no re-quoting).
3. The flattened string runs via `bash -c`.

With `USER_INPUT=x'; touch /tmp/PWNED; echo '`, the final command is `echo 'x'; touch /tmp/PWNED; echo ''` → `touch /tmp/PWNED` executes.

**Fix (in progress):** stop quote-before-resolve for argv. Carry argv elements to the run boundary as a structured `Vec`, resolve `{{ env.* }}` per element there, **then** `shell_quote` each resolved element and join. `script` steps stay verbatim. Add a regression test for an argv element whose resolved env value contains shell metacharacters.

Converting to draft until the fix lands.

swerner:
## Security fix: resolve-then-quote for argv prepare steps

Fixes a command-injection in argv-style `run.prepare.steps` (e.g. `command = ["echo", "{{ env.X }}"]`).

### The defect
Argv elements were shell-quoted **in source form** at config-resolve time, then the raw `{{ env.* }}` value was substituted **inside** those quotes at the run boundary (no re-quoting). The flattened string ran under `bash -c`, so an env value containing a single quote and `;` (e.g. `x'; touch /tmp/PWNED; echo '`) broke out of its argument and injected arbitrary commands. Quote-before-resolve was the bug.

### The fix: resolve, THEN quote, on structured argv
- `PreparedStep` now carries the runnable part as an enum that preserves the script-vs-argv distinction:
  - `Script(String)` — a single raw shell snippet, kept verbatim.
  - `Command(Vec<String>)` — argv element source strings, **not** pre-joined and **not** shell-quoted at config time.
- At the run boundary, each argv element's `{{ env.* }}` token resolves first (missing env stays a hard error / fail-closed), and only the **resolved** value is then shell-quoted (via `fabro_util::shell::shell_quote`) and joined with spaces. Script steps resolve and run verbatim; per-step `env` values resolve into the env map as before (already safe).

### Wire-shape change
`PreparedStep` serializes into the run spec/manifest, so the shape changed to a `type`-discriminated `script`/`command` object. Updated the OpenAPI `PreparedStep` schema, regenerated the Rust types (`fabro-api`), and regenerated + typechecked the TypeScript client (scoped to PreparedStep only). Updated the run-spec assertion in the CLI/server tests to the new shape.

### Regression test
`fabro-types` `resolve_step_env_tests::command_arg_resolving_to_shell_metacharacters_is_not_injected` runs config resolve then `resolve_step_env` for `["echo", "{{ env.USER_INPUT }}"]` with a single-quote+`;` payload, asserting the flattened command round-trips through `shlex::split` to **exactly two tokens** with the payload intact and no injected command leaking out. This fails on the pre-fix behavior and passes after. Positive coverage retained: an argument with a space stays one token, and script steps remain verbatim.

### Gates
`cargo build --workspace`, nightly `fmt --check`, nightly `clippy --workspace --all-targets -D warnings`, and `cargo nextest run` for fabro-types / fabro-config / fabro-workflow / fabro-server / fabro-cli / fabro-api all pass; TS client typecheck passes.

Left as a draft for human review of the security fix.

swerner:
✅ **Command-injection fix verified — marking ready.**

Independently confirmed (read the resolved code + ran the tests):

- **Order is resolve-then-quote.** `resolve_step_env` resolves `{{ env.* }}` per argv element (no quoting); `PreparedStep::to_shell_command()` then `shell_quote`s each *resolved* element and joins; the run boundary calls them in that order. Config layer carries argv as an unquoted `Vec` of source strings — no quote-before-resolve.
- **Regression test passes.** `command_arg_resolving_to_shell_metacharacters_is_not_injected`: with `USER_INPUT=x'; touch PWNED; echo '`, the flattened command `shlex`-splits to exactly `["echo", "x'; touch PWNED; echo '"]` (2 tokens, no injected `touch`). All 8 `resolve_step_env_tests` green locally; CI Clippy/Format/Typecheck/Test green.

`PreparedStep` is now a `type`-discriminated `{script}` / `{command:[…]}` enum (OpenAPI + Rust + TS client updated). Scripts stay verbatim; only argv elements are quoted, after resolution.

swerner:
Ran simplify on PR 530.

Results:
- Finished the shared shell-quote extraction for workflow, ACP command rendering, and dev tooling; removed now-unused direct shlex deps where applicable.
- Replaced anonymous generated PreparedStepOneOf API client types with named PreparedScriptStep / PreparedCommandStep schemas and generated models.
- Removed the invalid Default from PreparedStep/PreparedStepRun and cleaned stray internal planning references from touched comments.

Verification:
- bun run generate
- bun run typecheck
- cargo build -p fabro-api
- cargo check -p fabro-acp --no-default-features
- cargo check -p fabro-dev -p fabro-workflow -p fabro-types -p fabro-config
- cargo nextest run -p fabro-util shell
- cargo nextest run -p fabro-types resolve_step_env_tests
- cargo nextest run -p fabro-config resolve_prepare_tests
- git diff --check

Note: I left the low-priority startup-only clone in prepare env resolution unchanged; avoiding it would complicate the shared resolver API for a non-hot path.

swerner:
## Ran `/simplify` (Opus 4.8) on this branch

Reviewed the 3-commit prepare-steps refactor for **quality only** (reuse, simplification, efficiency, altitude — not correctness bugs) using four parallel review agents, then applied the fixes that held up. The branch was already largely a clean refactor, so most findings converged on "keep as-is"; two were worth acting on.

### Applied (commit `1bffdb1`)

- **Shared traversal visitor.** `RunNamespace::substitute_variables` (the `{{ vars.* }}` pass) and `RunPrepareSettings::resolve_step_env` (the `{{ env.* }}` run-boundary pass) each hand-coded the same `match PreparedStepRun { Script | Command } + walk step.env` traversal. Extracted `visit_prepared_step_strings`, mirroring the existing `visit_mcp_transport_strings`, and routed both passes through it so they can't drift as those types grow fields.
- **`shell_join` helper.** The "shell-quote each argv element, join with spaces" pattern was newly added in `PreparedStep::to_shell_command` and already open-coded in `fabro-acp`'s `render_command`. Added `fabro_util::shell::shell_join` next to `shell_quote` (with tests) and pointed both call sites at it.

### Reviewed and deliberately skipped

- **`Script` string double-clone** (efficiency) — one tiny allocation once per run at launch; the consuming-method fix adds parallel API surface for negligible gain.
- **`resolve_step_env`'s up-front `self.clone()`** — deliberately mirrors the sibling `resolve_transport_env`, and the call site holds `prepare` behind a shared `&`, so removing it would force a *larger* clone. Kept.
- **Doc-rationale repetition across ~5 sites** — the repeated text is the security/ordering rationale (resolve-then-quote, carry source form for `fabro validate` portability); valuable and low-drift, so left intact rather than churned into cross-references.
- **Resolve-then-quote ordering as an implicit contract** (altitude) — the one genuine altitude concern, but soft: it's consistent with the already-accepted `resolve_transport_env` convention, has a single correct caller today, and the structural fix (type-state or a merged resolve+quote method) is a cross-crate redesign better left to the broader interpolation-cleanup follow-up this file already points at.

### Verification

- `cargo build` across affected crates — clean.
- `cargo nextest` on fabro-util/types/acp/config + the workflow/server prepare-step tests — pass (incl. the PR's own `resolve_step_env`/per-step-env tests and the new `shell_join` tests).
- `clippy -D warnings` and `fmt --check` on touched crates — clean.

(One unrelated pre-existing test, `initialize_prepares_sandbox_and_uses_persisted_run_dir`, asserts `credentials.is_empty()` and fails identically on the clean tree — a local-env artifact, not from these changes.)

PR #533, opened 2026-06-26
feat(web): playground run UX — Save to Runs, auto-start, retry any terminal run

UmutAlihan:
Three small, independent web-only improvements to the run UX, each with tests. No backend changes.

### 1. Retry any terminal run from the runs list
The row actions menu only offered **Retry** for `failed`/`dead` runs, even though the server re-creates a fresh run from the stored spec for any terminal, non-archived run — succeeded ones included. Adds a string-status `canRetryStatus` predicate (mirroring the existing `canRetry`, which needs the full lifecycle object) for list rows that carry only the flattened `lifecycleStatus`, and uses it to gate the action. Archived rows surface as `"archived"` and stay excluded.

### 2. "Save to Runs" in the playground + run-manifest fixes
Adds a **Save to Runs** button that creates a run from the current draft (create-only — no execution) and surfaces server errors inline. Two compatibility fixes make playground-built manifests actually accepted at run creation:

- **Drop `[run.sandbox]` from generated `workflow.toml`.** The server parses that file into a `RunLayer` with `deny_unknown_fields` and no `sandbox` field, so the section made the whole manifest unparseable (400). Sandbox selection moves to `project.toml`'s `[environments.default]`, the supported home for it.
- **Clamp the goal-derived run title to the server's 100 code-point limit** (counting by code point to match Rust's `chars().count()`), so long goals are no longer rejected.

### 3. "Run for real" actually starts the run
The modal copy promises to redirect "to its run page when it starts", but `POST /api/v1/runs` only creates the run in `submitted` status and the run page has no start affordance — so launched runs sat `submitted` forever. After creating the run, the modal now POSTs `/api/v1/runs/{id}/start` and only redirects once it has started; a start failure (e.g. 409 "not startable") is surfaced inline instead of redirecting to a stuck run.

### Testing
- `bun run typecheck` — clean
- `bun test` — 635 pass / 0 fail

Commits are split per concern for easier review.

### DIFF
--- a/apps/fabro-web/app/components/playground/files/render-toml.test.ts
+++ b/apps/fabro-web/app/components/playground/files/render-toml.test.ts
@@ -4,7 +4,10 @@ import { createInitialDraft } from "../state/draft";
 import { renderProjectToml, renderWorkflowToml } from "./render-toml";
 
 describe("renderWorkflowToml", () => {
-  test("points the workflow at workflow.fabro and pins sandbox to local", () => {
+  test("points the workflow at its graph without an unsupported [run.sandbox] section", () => {
+    // The server's RunLayer parses workflow.toml with deny_unknown_fields and
+    // has no `sandbox` field, so a `[run.sandbox]` section makes the manifest
+    // unparseable. Sandbox selection lives in project.toml instead.
     expect(renderWorkflowToml(createInitialDraft())).toBe(
       [
         "_version = 1",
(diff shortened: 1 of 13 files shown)


brynary:
Hi @UmutAlihan -- Thanks for this contribution. We are doing some additional work on the Playground soon and will incorporate this direction when we circle back to it.

PR #536, opened 2026-06-29
gcloud: keep stdin attached to the spawned ACP agent

duics:
## Root cause

The gcloud provider's `spawn_stdio_process` is the **bidirectional ACP transport** — fabro writes JSON-RPC requests to the spawned agent's stdin and reads responses from its stdout. But it launches the agent through `build_exec_script` -> `wrap_bash`, whose form was:

```
echo <b64> | base64 -d | bash
```

In that pipeline the final `bash` (and the agent it execs) inherits the **decode pipe** as its stdin — exhausted right after the script is read — instead of this process's stdin. So the ACP agent gets EOF on stdin and never receives a single request:

- **claude-agent-acp**: exits clean (`exit_code=0`) before the handshake -> run fails with `ACP process exited before protocol completed`, `inference_time_ms=0`.
- **codex-acp**: stalls similarly.

This broke **every** stdio ACP agent on the gcloud provider — the entire credential investigation was a red herring. Verified by driving `claude-agent-acp` by hand on a sandbox VM (same image) with the injected `CLAUDE_CODE_OAUTH_TOKEN` + the squid proxy + `HOME=/home/fabro`: it completes a full ACP turn with **real inference**. The only thing the real run did differently was the broken transport.

Confirmed with pure shell semantics:
```
$ printf 'HELLO\n' | bash -c "echo $(printf cat|base64) | base64 -d | bash"   # (old) -> prints nothing
$ printf 'HELLO\n' | bash -c "bash <(echo $(printf cat|base64) | base64 -d)"  # (new) -> prints HELLO
```

## Fix

Decode through **process substitution** — `bash <(echo <b64> | base64 -d)` — which runs the identical script with this process's stdin still attached. Keeps `wrap_bash`'s base64 SSH-arg safety, and is harmless for the fire-and-forget `exec_command` path (stdin unused there). The `startup_script` egress fragment is a separate literal and is untouched.

Adds `wrap_bash_keeps_stdin_attached_to_the_inner_command` — a behavioral regression test that drives the wrapped script's stdin and asserts the inner command receives it (fails on the old form). All 7 `gcloud::tests` pass.

https://claude.ai/code/session_01RdsB6jgc8VJykXDiTB7T93


duics:
Opened against the wrong base by mistake — this is an internal in-parallel-oy/fabro fork fix. Reopening there.

PR #541, opened 2026-07-01
fix(web): Enabling scroll in the Stages sidebar on the run overview/stages tabs

andremw:
This change enables scrolling the stages sidebar on the run's overview/stages page. Without it, for long runs with lots of stages, the entire page scrolls, hiding the graph while it's running.

https://github.com/user-attachments/assets/c5405a5b-8480-46f8-8d7c-4cd4914f6228


brynary:
@andremw -- thanks! This looks good, will merge soon.

PR #550, opened 2026-07-02
Fix web app load performance: caching, compression, and eager chunk loading

swerner:
## Problem

Loading the web UI from a remote server took **~11 seconds to first render on every refresh**. A HAR capture against a remote deployment showed the page downloading **13.5 MB of JavaScript across 356 files, uncompressed, on every single page load** — even though the assets are content-hashed and served with `Cache-Control: immutable`.

Four compounding causes:

1. **`Pragma: no-cache` defeated the browser cache.** The security-headers middleware stamped `Pragma: no-cache` onto every response, including hashed assets that set a year-long immutable `Cache-Control`. Browsers treat a response `Pragma: no-cache` as `Cache-Control: no-cache` and check it *before* `max-age` (Chromium zeroes freshness on it), and since assets carried no validators, "revalidate" degraded into a full re-download. Empirically visible in the HAR: Google-Fonts woff2s served from cache (`transfer = 0`) during the same page load where all 356 of our assets re-downloaded in full.
2. **No response compression.** The server had no compression layer; 13.5 MB of JS compresses to ~2.5 MB with brotli.
3. **The HTML force-loaded every chunk.** `writeIndexHtml` emitted a `<script type="module">` tag for all 356 outputs. Only 2.9 MB is statically reachable from the entry; the other ~10.7 MB is dynamic-import-only code (syntax grammars, Graphviz WASM, xterm, diff file tree) that was being downloaded eagerly at high priority.
4. **The immutable heuristic over-matched.** Any dash in a filename counted as a content hash, so stable-named files (`pierre-diffs-worker/worker-portable.js`, `apple-touch-icon.png`) would be pinned in browser caches for a year across deploys once fix 1 made immutable caching effective.

## Changes

- **`security_headers`**: apply the `no-store`/`Pragma: no-cache` defaults only when the handler didn't set its own `Cache-Control`. API responses keep the conservative defaults.
- **Compression**: `tower-http` `CompressionLayer` (brotli + gzip) on both the main router and the install-mode router (install mode serves the same SPA bundle through a separate router). Default predicate keeps SSE (`text/event-stream`), gRPC, images, and tiny bodies identity-encoded. Quality pinned to `Precise(4)` — tower-http's default defers to the codec default, and brotli's default is quality 11 (seconds of CPU per multi-megabyte asset).
- **Entry-only HTML**: `writeIndexHtml` emits script tags only for `kind === "entry-point"` outputs. The module graph pulls static imports (depth 1, so no waterfall); dynamic `import()` chunks load on demand.
- **Cache-control classifier + validators**: only files matching the bundler's actual output shape (`assets/<stem>-<hash8>.js|css`, lowercase base-36) get `immutable`. Everything else is `no-cache` **with a strong ETag** and `If-None-Match` → `304` support, so index.html / app.css / the pierre worker revalidate in one cheap conditional request instead of a full re-download.

## Impact (measured on the built bundle)

| | Before | After |
|---|---|---|
| Cold load, ~1 MB/s link | 13.5 MB raw ≈ **11–14 s** | ~0.8 MB compressed eager payload ≈ **~1 s** |
| Refresh | full re-download, same 11–14 s | served from cache + one 304 ≈ **instant** |
| Eager JS on first render | 13.56 MB / 356 files | 2.88 MB raw (0.79 MB gzip) / 6 files |

## Verification

- 959 fabro-server tests pass (incl. new coverage); fmt + clippy clean; `bun run typecheck` passes (the 5 pre-existing bun test failures reproduce identically on `main` — missing `@pierre/diffs/dist/worker` fixture + flaky InstallApp timing tests).
- New integration tests pin compression through **both** serving shapes that matter: regular routes and the SPA fallback service, each via tower `oneshot` **and** over a real TCP connection through hyper (raw-socket assertions, so no client auto-decompression can mask a regression).
- Live-verified against a debug server: hashed assets get `immutable` + brotli and no `Pragma`; mutable assets get `no-cache` + ETag and answer conditionals with `304`; API responses keep `no-store`.
- Headless Chrome boots the rebuilt SPA from the entry-only HTML and fully renders the UI.

## Notes for reviewers

- The ETag is skipped for immutable assets deliberately — they never revalidate, so hashing multi-MB bodies per request would be pure overhead.
- Install mode previously had **no** compression and shares the same bundle; it gets the same layer via a shared `compression_layer()` helper.
- `bun test` has a pre-existing suite (`production build copies Pierre worker assets`) that fails without `@pierre/diffs/dist/worker` present locally; unrelated to this change.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -154,6 +154,21 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "250f629c0161ad8107cf89319e990051fae62832fd343083bea452d93e2205fd"
 
 [[package]]
+name = "alloc-no-stdlib"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "cc7bb162ec39d46ab1ca8c77bf72e890535becd1751bb45f64c597edb4c8c6b3"
+
+[[package]]
+name = "alloc-stdlib"
+version = "0.2.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "94fb8275041c72129eb51b7d0322c29b8387a0386127718b096429201a5d6ece"
+dependencies = [
+ "alloc-no-stdlib",
+]
+
+[[package]]
 name = "allocator-api2"
 version = "0.2.21"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -313,6 +328,18 @@ dependencies = [
  "event-listener-strategy",
  "futures-core",
  "pin-project-lite",
+]
+
+[[package]]
+name = "async-compression"
+version = "0.4.42"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e79b3f8a79cccc2898f31920fc69f304859b3bd567490f75ebf51ae1c792a9ac"
+dependencies = [
+ "compression-codecs",
+ "compression-core",
+ "pin-project-lite",
+ "tokio",
 ]
 
 [[package]]
@@ -1071,6 +1098,27 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dc0b364ead1874514c8c2855ab558056ebfeb775653e7ae45ff72f28f8f3166c"
 
 [[package]]
+name = "brotli"
+version = "8.0.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9991eea70ea4f293524138648e41ee89b0b2b12ddef3b255effa43c8056e0e0d"
+dependencies = [
+ "alloc-no-stdlib",
+ "alloc-stdlib",
+ "brotli-decompressor",
+]
+
+[[package]]
+name = "brotli-decompressor"
+version = "5.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "874bb8112abecc98cbd6d81ea4fa7e94fb9449648c93cc89aa40c81c24d7de03"
+dependencies = [
+ "alloc-no-stdlib",
+ "alloc-stdlib",
+]
+
+[[package]]
 name = "bs58"
 version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


swerner:
## Cleanup pass (`/simplify`)

Ran a 4-angle cleanup review (reuse, simplification, efficiency, altitude) over this PR's diff and applied the surviving findings in 5b5f20bf7. All fabro-server/fabro-spa tests, clippy, a full workspace check, web typecheck, and a real `bun run build` pass.

### Applied

- **Mutable-asset ETags no longer hash the body on every request** (`static_files.rs`, `fabro-spa`). The new ETag path recomputed a full SHA-256 per request — and since these assets are `no-cache`, that's every page load. rust-embed already computes each asset's SHA-256 at compile time, so `fabro_spa::AssetBytes` now exposes it and the server uses it directly. As part of the same plumbing, embedded assets are served via `Bytes::from_static` instead of copying the whole body out of the binary per request — that copy also happened on the 304 path, where the body is discarded. Disk-served assets (dev/watch mode, explicit asset roots) still hash on read, which they need for freshness.
- **ETag policy no longer derived by string-comparing header values** (`static_files.rs`). `cache_control == REVALIDATE_CACHE_CONTROL` encoded "needs a validator" as an incidental string equality; a single `is_content_hashed(path)` bool now drives both the `Cache-Control` value and the ETag decision. Tests assert the classifier and the header mapping separately.
- **Compression test dedup** (`tests/it/api/compression.rs`). The tempdir-with-8KB-index fixture and the bind/spawn `axum::serve` scaffolding were each copy-pasted between two tests; both are now local helpers (`spa_router_with_big_index`, `serve_on_ephemeral_port`), ~40 lines shorter.
- **Dropped the `String(output.kind)` coercion** (`scripts/build.ts`). `output.kind` is already a string; `IndexHtmlOutput.kind` is now typed as Bun's kind union so the `=== "entry-point"` comparison is typo-safe.

### Considered and skipped

- **Precompressing the SPA bundle at build time**: valid finding — per-request brotli of identical immutable bytes is repeated CPU — but it's a build-pipeline redesign, beyond a cleanup pass. Worth a follow-up; it would also unlock brotli-11 ratios at zero request-time cost.
- **Replacing the `is_content_hashed` filename heuristic with a build-emitted manifest of immutable paths**: the deeper design (a Bun naming change would currently degrade silently to `no-cache`), but a cross-cutting build+server change; the heuristic is documented, fail-safe, and tested.
- **Replacing the raw-socket HTTP test with reqwest**: the raw client is deliberate per its comment — no client-side redirect following or transparent decompression can distort the wire bytes. Deduped its scaffolding instead.
- **Collapsing oneshot + real-TCP test pairs**: intentional layered coverage per the file's doc comment; kept.

PR #567, opened 2026-07-10
feat(sandbox): forkd Firecracker microVM provider (rebased onto main, supersedes #517)

zenprocess:
## Summary

This adds **forkd** (Firecracker microVM controller) as a sandbox provider, selectable per run via a named environment (`provider = "forkd"`). It supersedes draft #517 with the same provider rebased onto current `main` — #517's branch had fallen 50+ commits behind and its diverged/`unknown` mergeable state made it a poor review target, so this is a clean resubmission rather than a force-push over the existing thread.

## Re: the architecture note on #517

@brynary flagged in #517 that the review surfaced "pre-existing architectural design issues that we may want to resolve before adding the next sandbox provider," with details to follow. We'd genuinely like to align on that before (or as part of) landing this — if there's a refactor you want as a precondition (provider registry shape, `Sandbox` trait split, clone-source contract, anything else), name it and we're happy to take it on or adapt this PR to it. Happy to move that conversation to an Issue/Discussion if that's the better venue per CONTRIBUTING's "align on approach first" guidance for larger features.

In the meantime, this PR deliberately **conforms to the existing contracts** rather than inventing new ones:

- Implements the current `Sandbox` trait and provider registry as-is; no trait changes.
- Follows the clone-based provider contract (`is_clone_based()`, run-manifest GitHub origin metadata, `skip_clone` semantics) exactly as Docker/Daytona do.
- All shell interpolation goes through `shell_quote()`; the one env-var export path is guarded by an identifier-only allowlist.
- Config resolution mirrors Daytona: `FORKD_URL` / `FORKD_TOKEN` / `FORKD_SNAPSHOT_TAG` from the environment, token redacted in `Debug`.
- Fully feature-gated (`forkd` feature, default **off**): default builds compile byte-compatible with `main`; every always-compiled match over `SandboxProviderKind` pairs a gated arm with a graceful-error arm.

## Delta over #517's original 9 commits

- **Rebased onto current `main`** (no longer 50+ commits behind; adapts to the clone-based provider and config changes that landed since).
- **Feature-gating fixes**: `#[cfg(feature = "forkd")]`-gated imports that were unconditional; `fabro-cli`'s `forkd` feature now propagates to `fabro-server` so `fabro server` actually registers the provider.
- **Reconnect correctness**: `ForkdConfig` is resolved from the environment on reconnect (was `Default`, i.e. hardcoded localhost + dev token).
- **DB schema**: new migration widening the `environments.provider` CHECK to include `forkd` (table-recreate; the applied migration is untouched for checksum integrity), with accept/reject tests.
- **Build tooling**: `cargo dev docker-build --features <list>` passthrough so a release image build can enable the feature explicitly.
- **Lint debt**: the forkd module is clippy-clean under `--features forkd -D warnings` (22 findings fixed).

## Testing

- `cargo build --workspace` (default features) — exit 0
- `cargo check -p fabro-cli --features forkd` — exit 0
- `cargo nextest run -p fabro-sandbox -p fabro-server -p fabro-db` — pass (incl. new forkd CHECK-constraint tests)
- `cargo +nightly fmt --check --all` — clean
- `cargo clippy --workspace --all-targets -- -D warnings` — clean, both with and without `--features fabro-cli/forkd`
- Exercised against a live forkd v0.5.2 controller (`/v1/sandboxes` API): create-from-snapshot, exec, file upload, teardown.
- This rebased branch is deployed and running on our own gate infrastructure: `fabro server` boots clean at 0.290, the `forkd` provider registers, a `[run.environment] id = "forkd"` environment resolves, and runs dispatch through it to the live forkd controller.

Happy to iterate — and again, if the architectural preconditions from #517 are written up anywhere, point us at them and we'll align.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2950,6 +2950,7 @@ dependencies = [
  "hmac 0.12.1",
  "httpmock",
  "rand 0.9.4",
+ "reqwest 0.13.2",
  "reqwest-middleware",
  "rustls",
  "serde",
--- a/lib/crates/fabro-agent/Cargo.toml
+++ b/lib/crates/fabro-agent/Cargo.toml
@@ -13,6 +13,7 @@ categories = ["api-bindings"]
 [features]
 default = ["docker"]
 docker = ["fabro-sandbox/docker"]
+forkd = ["fabro-sandbox/forkd"]
 quarantine = []
 
 [lib]
--- a/lib/crates/fabro-cli/Cargo.toml
+++ b/lib/crates/fabro-cli/Cargo.toml
@@ -13,6 +13,7 @@ path = "src/main.rs"
 [features]
 default = []
 sleep_inhibitor = ["dep:core-foundation"]
+forkd = ["fabro-sandbox/forkd", "fabro-agent/forkd"]
 
 [lints]
 workspace = true
--- a/lib/crates/fabro-cli/Cargo.toml
+++ b/lib/crates/fabro-cli/Cargo.toml
@@ -13,7 +13,7 @@ path = "src/main.rs"
 [features]
 default = []
 sleep_inhibitor = ["dep:core-foundation"]
-forkd = ["fabro-sandbox/forkd", "fabro-agent/forkd", "fabro-workflow/forkd"]
+forkd = ["fabro-sandbox/forkd", "fabro-agent/forkd", "fabro-workflow/forkd", "fabro-server/forkd"]
 
 [lints]
 workspace = true
--- a/lib/crates/fabro-cli/Cargo.toml
+++ b/lib/crates/fabro-cli/Cargo.toml
@@ -13,7 +13,7 @@ path = "src/main.rs"
 [features]
 default = []
 sleep_inhibitor = ["dep:core-foundation"]
-forkd = ["fabro-sandbox/forkd", "fabro-agent/forkd"]
+forkd = ["fabro-sandbox/forkd", "fabro-agent/forkd", "fabro-workflow/forkd"]
 
 [lints]
 workspace = true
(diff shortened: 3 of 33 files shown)


zenprocess:
To unblock the architecture conversation from #517 (the "More soon" follow-up from @brynary never landed in the thread), I've opened #583 — *"Sandbox provider architecture preconditions before landing forkd (re #517/#567)"*.

The new issue lists what #567 already conforms to (existing `Sandbox` trait + provider registry, clone-based contract, `shell_quote()` everywhere, Daytona-shaped config, default-off feature gate) and asks @brynary to name the specific preconditions in concrete categories — provider registry shape, `Sandbox` trait split, clone-source contract, or anything else — so we can either align #567 against them or take the refactor on as a prerequisite PR.

Pinging here for visibility; discussion should live in #583 so it stays on the record.

swerner:
Hey @zenprocess , thanks for this! With the rise of all these sandbox/vm providers, @brynary and I have been talking about restructuring the way we handle and connect to sandboxes in more of a plugin style system so that we can support any of the services that someone might use. We have a little bit more to discuss here and a bit of work to do before we can land it.

zenprocess:
Thanks @swerner, @brynary — a plugin-style provider system is exactly the right call, and it's the abstraction that lets forkd land cleanly. Firecracker microVMs are a genuinely different shape from container providers (snapshot/restore, hermetic boot, per-run isolation), so if it's useful, the forkd provider here can serve as a reference implementation / stress-test for the interface — it should surface any assumptions that only hold for container-based sandboxes.

Would you be able to share your reference architecture or current design thinking for the plugin system, even rough? We'd much rather align forkd to your direction and contribute the work than diverge — happy to prototype the provider interface against it, or take on a chunk of the implementation, whichever is most useful to you.

brynary:
@zenprocess thanks for your patience on this. Here is an initial sketch of how we are considering a plugin system for sandbox providers based on JSON-RPC: https://quarry.lithos.computer/tmp/3cda04aa788840cdb46bcbefb5f60fd4

Feedback welcome!

PR #568, opened 2026-07-11
feat(workflow): keep push credentials fresh across long ACP turns

thewoolleyman:
## Problem

GitHub App installation tokens expire ~60 minutes after minting. On a long
run, the `origin` token embedded in the sandbox clone at run start is stale by
the time a late ACP node (e.g. the PR node) runs `git push`, which fails with
`Invalid username or token`.

## Change

Two host-driven mechanisms, both reusing the existing
`Sandbox::refresh_push_credentials()` (re-mint + `git remote set-url origin`)
over the exec channel — no new inbound surface:

1. **Turn-entry re-mint** — refresh at each ACP node entry, so a push early in
   the turn uses a fresh token.
2. **Refresh-ahead loop** — a background task, scoped to the turn by a
   drop-guard, that re-mints every 45 min so a single push-bearing turn that
   itself outlives the TTL stays fresh. A normal sub-interval turn never ticks;
   a failed/timed-out tick retries sooner so a transient error cannot leave a
   longer-than-interval expired-token window.

Both refresh calls are timeout-bounded (30s) so a stalled GitHub API cannot
hang node entry (ACP nodes use the `HandlerManaged` timeout policy, so nothing
else bounds this pre-spawn await).

Controls (both added to the worker env allowlist):

- `FABRO_PUSH_CRED_REFRESH_AHEAD` — default on; a falsy value
  (empty/`0`/`false`/`off`/`no`, case-insensitive) disables the whole feature
  (turn-entry + loop) for operators who manage `origin` themselves.
- `FABRO_PUSH_CRED_REFRESH_INTERVAL_SECONDS` — overrides the interval; `0`
  disables just the loop.

`refresh_push_credentials` now returns `RefreshOutcome` (`Refreshed` /
`Skipped`) so callers log accurately: `Refreshed` only when a GitHub App token
was actually re-minted; a static PAT or pre-minted `Installation` token
short-circuits to `Skipped` before the set-url exec. (This also means a PAT
sandbox no longer re-runs `set-url` on refresh as it did before — intentional,
since only App installation tokens can be re-minted.)

## Testing

- clippy (`-D warnings`) and `cargo fmt --check` — clean.
- New unit tests cover enable/interval parsing (default, `0`-disable,
  case-insensitive falsy, empty, unparsable) and the no-credentials `Skipped`
  path.
- Validated on real >60-min runs: a fresh mint was observed at +90 min with
  zero TTL-expiry errors; an A/B with the loop disabled reproduced the original
  `Invalid username or token` failure.

## Known follow-ups (documented in-code)

- Resumed/parked runs reconnect the sandbox with no GitHub App creds, so
  refresh no-ops for them until those creds are threaded through the reconnect
  path.
- No freshness check on the turn-entry mint (it mints once per node entry even
  when the current token is fresh).
- The background `set-url` can contend with the agent's own git on
  `.git/config.lock`; parallel ACP branches each run their own loop.
- The refresh lives in the ACP handler only, though the stale-origin problem is
  stage-agnostic.
- Refresh failures are logged but not surfaced as a RunNotice event.

### DIFF
--- a/lib/crates/fabro-agent/src/lib.rs
+++ b/lib/crates/fabro-agent/src/lib.rs
@@ -55,8 +55,8 @@ pub use question_tools::{
 };
 pub use read_before_write_sandbox::ReadBeforeWriteSandbox;
 pub use sandbox::{
-    CommandOutputCallback, DirEntry, ExecResult, ExecStreamingResult, GrepOptions, Sandbox,
-    SandboxEvent, SandboxEventCallback, StderrCollector, StdioProcess, StdioProcessHandle,
+    CommandOutputCallback, DirEntry, ExecResult, ExecStreamingResult, GrepOptions, RefreshOutcome,
+    Sandbox, SandboxEvent, SandboxEventCallback, StderrCollector, StdioProcess, StdioProcessHandle,
     WorktreeEvent, WorktreeEventCallback, WorktreeOptions, WorktreeSandbox, format_lines_numbered,
     shell_quote,
 };
--- a/lib/crates/fabro-agent/src/sandbox.rs
+++ b/lib/crates/fabro-agent/src/sandbox.rs
@@ -2,8 +2,8 @@
 // Re-export the delegate_sandbox! macro at crate root so existing
 // `crate::delegate_sandbox!` invocations continue to work.
 pub use fabro_sandbox::{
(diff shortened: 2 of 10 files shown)


swerner:
Thanks @thewoolleyman! Digging more in to this

PR #570, opened 2026-07-11
Migrate secrets storage to SQLite

brynary:
## Summary

Moves optional integration secrets (token, OAuth, and file secrets) from `vaults/default/secrets.json` to the shared SQLite database, preserving the REST contract and secret-type behavior. Three commits:

- **Migrate secrets storage to SQLite** — new `secrets` table and `SecretStore` in fabro-vault; async/fallible secret access through the server; retry-safe legacy JSON import that renames the source to `.bak`; OAuth refresh persisted via revision-based compare-and-swap
- **Simplify SQLite secret storage** — store API consolidation
- **Clean up + regression fixes** (review pass):
  - `SecretStore::open()` consolidates the connect → migrate → import-legacy sequence previously repeated at five call sites; fabro-agent and fabro-cli drop their fabro-db dependency
  - Fixes a regression where the standalone CLI/agent lost process-env LLM credential lookup (`ANTHROPIC_API_KEY=… fabro exec` stopped resolving)
  - Fixes five install tests that still asserted against the legacy `secrets.json` after writes moved to SQLite
  - Makes `preloaded_vault` required, deleting a fallback that re-read the already-renamed legacy file; `load_startup_vault` is now test-only
  - Skips snapshot clones and CAS retry machinery during credential resolution when no OAuth secrets exist (per-request hot path)
  - Removes dead code (`persist_with_secret_store`, `VaultSecretWrite` alias, `RowCountOverflow`, duplicated row parsing); runs `check_crypto` concurrently with other diagnostics

Bootstrap secrets (`SESSION_SECRET`, dev token, database/object-store credentials) remain in process env or `server.env`. Encryption-at-rest is out of scope and preserved as plaintext-at-rest for a follow-up.

## Test plan

- `cargo nextest run --workspace` — 6,969 passed, 0 failed


swerner:
Rebased onto current main and pushed (previous head e1535f4, three original commits preserved). Two notes:

1. **Conflict resolution with #585** (pre-migration db snapshot, merged this morning): both changes added a private-permissions helper to `fabro-db`. Kept main's `set_private_permissions` (the snapshot path uses it) alongside this PR's `prepare_private_database_file` (pre-connect 0600 create) — the two are complementary, no behavior change to either.
2. **Added one commit**: the legacy secrets JSON import now declares a removal deadline (`2026-10-11`) and logs it, matching the pattern in docs/internal/migrations-strategy.md. Drop it if you'd rather handle that separately.

Local verification: fmt + clippy `-D warnings` clean; 1992/1992 tests pass across fabro-db/vault/auth/install/server/cli.

PR #571, opened 2026-07-11
Move secrets and MCP server storage to SQLite

brynary:
## Summary

- Migrate secrets storage from the JSON vault file to SQLite (`secrets` table) with a one-time legacy import that renames the old file to a timestamped `.bak`
- Migrate MCP server definitions from `mcps/*.toml` to SQLite (`mcp_servers` table) with the same one-time legacy import pattern; reads come from an in-memory catalog, mutations use revision-guarded SQL
- Add `SqlVaultCredentialSource` so LLM credential resolution reads from the SQLite-backed store and persists OAuth token refreshes via revision-checked CAS
- Startup now runs the optional server-env secrets migration against the SQLite store; the old Vault-based migration path is deleted and its tests re-pointed at the production path
- Post-review cleanup: shared `fabro_db::legacy` helpers (timestamped backup rename, `is_toml_file`) and `parse_rfc3339_utc` replace four per-crate copies; single bind order for MCP INSERT/UPDATE; generic `load_store_blocking`; snapshot-once env-secrets migration

## Test plan

- `cargo nextest run -p fabro-db -p fabro-vault -p fabro-mcp-store -p fabro-variable -p fabro-environment -p fabro-server` (803 tests pass)
- `cargo build --workspace`, nightly `clippy --workspace --all-targets -- -D warnings`, and `cargo fmt --check --all` all clean

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2846,12 +2846,18 @@ dependencies = [
 name = "fabro-mcp-store"
 version = "0.302.0-nightly.1"
 dependencies = [
+ "chrono",
+ "fabro-db",
  "fabro-types",
  "serde",
+ "serde_json",
+ "sqlx",
+ "strum 0.28.0",
  "tempfile",
  "thiserror 2.0.18",
  "tokio",
  "toml 0.8.23",
+ "tracing",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2561,6 +2561,7 @@ name = "fabro-db"
 version = "0.302.0-nightly.1"
 dependencies = [
  "anyhow",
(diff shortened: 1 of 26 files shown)


swerner:
Rebased onto main now that #570 is merged: the three secrets commits dropped out (already on main), leaving the pure MCP-servers diff (2 commits). One import-line conflict with #585's snapshot change in fabro-db, resolved by keeping both. Migration sequence on this branch: 2026071101_secrets (main) → 2026071102_mcp_servers (this PR) — no renumbering needed. Local: fmt + clippy -D warnings clean, 1981/1981 tests across fabro-db/vault/mcp-store/auth/server/cli.

PR #572, opened 2026-07-11
Add SQLite runs read model

brynary:
## Summary

Adds a SQLite-backed read model for run summaries (`runs` table in fabro.sqlite3), maintained synchronously on every event append and reconciled at startup, so run list/board/get/resolve endpoints are served by indexed SQL (filter, sort, paginate) instead of scanning the in-memory projection cache and decoding every run.

- New `RunSummaryStore` (fabro-store) with monotonic per-run upserts guarded by `source_last_seq`, startup reconcile, and SQL-side filtering/sorting/pagination
- `2026071101_runs.sql` migration: `runs` table with CHECK constraints and indexes for the query paths
- `GET /runs`, automation runs, and `/runs/resolve` now read from the SQLite store; responses are decorated per page only

## Follow-up simplification pass (second commit)

A three-agent review (reuse/quality/efficiency) of the changeset, with fixes:

- Status/board-column mapping now has a single source: `RunStatusKind::board_rank()` drives both the generated SQL status-sort CASE and the handler's column filter; a test upserts every status variant so the migration CHECK can't silently drift
- `RunSize::BUCKET_MAX_USD_MICROS` shared between the displayed size and the SQL size-sort CASE
- `/runs/resolve` uses a lean identity query (id, slug, name, origin via `json_extract`) instead of decoding every stored summary per request
- Deleted the 1:1 `RunsSortKey`/`RunsSortDirection` adapter enums; store enums carry the wire serde names
- `WorkflowRef::display_name()` consolidates the name→graph_name→slug fallback used in three places
- Shared pagination clamping and paginated list envelope across handlers
- Reconcile skips rows whose source seq is unchanged and batch-deletes stale rows; dropped two indexes no query could use (collation/expression mismatch)
- `RunDatabaseInner` holds the shared summary-store cell instead of a snapshot, so late attachment reaches already-open writers

## Testing

- `cargo nextest run` green across fabro-types, fabro-store, fabro-db, fabro-server (744), fabro-cli (952), fabro-workflow (1194)
- `cargo +nightly clippy --workspace --all-targets -- -D warnings` clean; `cargo fmt --check` clean

Note: the migration file changed between the two commits (index drop). Any local DB migrated from the first commit will hit a sqlx checksum mismatch and needs a reset — fine pre-merge since the migration is unreleased.


swerner:
Rebased onto main with #570/#571/#573 merged. One change beyond conflict resolution: **migration renumbered** `2026071101_runs.sql` → `2026071104_runs.sql` — it collided with the secrets migration's version number (sqlx uses the numeric prefix as the version; the collision only appears when the PRs combine). Final sequence on main once this merges: 2026071101 secrets, 02 mcp_servers, 03 automations, 04 runs. Same local-dev note as the sibling PRs: if you ran this branch before, reset your dev db and `cargo clean -p fabro-db`.

Conflicts were confined to the shared fabro-db test file (kept all sides' tests) and AppStores in server.rs (added run_summaries alongside the SecretStore-based vault field). Local: workspace clippy -D warnings clean; 2090/2090 tests across fabro-db/store/server/cli.

PR #576, opened 2026-07-14
feat(cli): opt-in OTLP/HTTP export for tracing spans

thewoolleyman:
## Summary

Adds an **opt-in** OTLP/HTTP exporter for fabro's existing `tracing` spans. It is
**additive and inert** unless an OTLP endpoint env var is set — when unset,
`otel_layer()` returns a no-op `None` layer, so the tracing stack is behaviorally
unchanged (the `fmt` layers and their output are untouched) and nothing is
exported. This PR is pure **transport**: it adds **no new instrumentation points**
to fabro's code. When enabled, fabro's existing `tracing` spans — the same span tree
that structures its log output — are bridged to OTLP via the standard
`tracing-opentelemetry` layer (which maps them to OTel using conventional semantics,
e.g. error events → span status). The optional synthetic extras (busy/idle timing,
thread, source location) are disabled, so exported spans stay close to fabro's actual
spans.

## What it does

- New `lib/crates/fabro-cli/src/otel.rs`: builds an `opentelemetry-otlp` span
  exporter + `tracing-opentelemetry` layer, guarded by `OTEL_EXPORTER_OTLP_ENDPOINT`
  / `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`. Returns a no-op (`None`) layer when unset.
- `logging.rs`: adds `.with(otel::otel_layer())` alongside the existing `fmt`
  layer in all four subscriber-init paths (cli / server-stdout / worker /
  worker-stdout).
- `main.rs`: registers `mod otel` and calls `otel::shutdown()` late in `main` for a
  best-effort final drain on the normal-exit path (no-op when disabled).
- `fabro-static/src/env_vars.rs`: registers the five `OTEL_*` names as `EnvVars`
  consts (fabro's env-var naming convention), referenced from `otel.rs`.
- Deps: `opentelemetry` / `opentelemetry_sdk` 0.30, `opentelemetry-otlp` 0.30
  (`default-features = false`; `http-proto` + `http-json` + `reqwest-blocking-client`
  — drops the unused OTLP logs exporter), `tracing-opentelemetry` 0.31. The OTLP HTTP
  features enable `metrics` unconditionally, so the trace+metrics SDK is compiled (a
  trace-only reduction isn't achievable via the HTTP exporter). These reuse the
  `reqwest 0.12` already in the lockfile — **no new reqwest version**. The lockfile
  gains `tonic 0.13` + `prost 0.13` transitively (via `opentelemetry-proto`'s protobuf
  message types for `http/protobuf`) — message codegen only; **no gRPC transport** is
  compiled or used.

## Env vars honored

This module resolves the **endpoint** itself (from `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`,
else `OTEL_EXPORTER_OTLP_ENDPOINT` + `/v1/traces`) and pins the **protocol**
(per-signal `OTEL_EXPORTER_OTLP_TRACES_PROTOCOL` wins, else
`OTEL_EXPORTER_OTLP_PROTOCOL`; default `http/protobuf`, the OTLP spec default; set
`http/json` to switch) and the **service name** (`OTEL_SERVICE_NAME`, default `fabro`).
The `opentelemetry-otlp` exporter reads the remaining OTLP vars itself — **headers**
(`OTEL_EXPORTER_OTLP_HEADERS` / `..._TRACES_HEADERS`) and **timeouts**
(`OTEL_EXPORTER_OTLP_TIMEOUT`). Export is gated on a non-empty resolved endpoint.

## Design / blast radius

- Zero behavior change when disabled: `otel_layer()` returns `None`, which is a no-op
  `tracing` layer; the `fmt` layers and their output are untouched.
- No changes to existing public APIs. `otel` is a private `fabro-cli` module
  (`pub(crate)`); the only call sites are the four subscriber inits and the `main`
  drain. (`fabro-static` gains five additive `EnvVars` consts — new public surface,
  not a change to existing APIs.)
- Env-var NAMES are `fabro_static::EnvVars` consts (the `OTEL_*` set was added there);
  the module-level `#![expect(clippy::disallowed_methods, reason = ...)]` covers only
  the `std::env::var` read mechanism, per fabro's env-var convention.
- Export volume is subject to the same `FABRO_LOG` `EnvFilter` as fabro's log output:
  the OTLP layer sits alongside the `fmt` layer under that filter, so a span below the
  configured level is neither logged nor exported.
- Span drain on exit is **best-effort, and mirrors fabro's existing pattern**:
  `otel::shutdown()` is placed right beside `fabro_telemetry::shutdown()` on the
  normal `main` exit path. fabro's own telemetry shutdown is likewise only on that
  path — its `std::process::exit(...)` bailout sites (preflight, run/*, doctor,
  server/status, …) do not drain `fabro_telemetry` either. Rerouting those sites to
  drain OTLP would make it more aggressive than fabro's own telemetry and touch
  unrelated exit paths; the SDK batch processor exports periodically regardless, so
  such an exit at most drops the last un-flushed batch. (A consistent drain of both
  telemetries on every exit path would be a separate, broader change to fabro's exit
  convention — out of scope for this transport PR.)
- Exporter/provider construction is wrapped in `catch_unwind`: the SDK batch
  processor `expect`s a successful background-thread spawn, so a (near-never)
  spawn failure disables export instead of taking fabro down. The panic still
  passes through fabro's installed panic hook before it is caught, so such a
  failure is reported once and then export is disabled rather than the process
  aborting (`panic = unwind`, so the catch is effective).
- Endpoint handling is fail-safe: the traces endpoint is resolved from the standard
  OTLP env vars (per-signal `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` used as-is, else
  `OTEL_EXPORTER_OTLP_ENDPOINT` + `/v1/traces`; empty/whitespace treated as unset)
  and passed **programmatically** via `.with_endpoint(...)`. A malformed endpoint
  fails the exporter build and disables export — there is no path where export is on
  while the endpoint silently falls back to localhost. (Delegating to the exporter's
  own env read would silently drop a malformed value and fall back to localhost.)
- Server-spawned workers run under `apply_worker_env`, an env allowlist (env-clear +
  narrow copy) that does not include `OTEL_*`; a server-launched worker therefore
  will not export until those vars are forwarded — a separate decision, since
  `OTEL_EXPORTER_OTLP_HEADERS` can carry secrets. Directly-invoked processes
  (cli / server / `__run-worker`) export normally.
- OTLP export uses `opentelemetry-otlp`'s own internally-built
  `reqwest::blocking::Client`, separate from fabro's `fabro_http` egress facade (so
  it does not consult `FABRO_HTTP_PROXY_POLICY`). This is deliberate for an opt-in
  telemetry exporter that owns its `OTEL_EXPORTER_OTLP_TIMEOUT`-configured client;
  routing OTLP through `fabro_http` (via `.with_http_client(...)`) is a possible
  follow-up if proxy-policy parity is wanted.

## Design decisions

- **Protocol default is `http/protobuf`** — the OTLP spec default. `http/json` is
  available via `OTEL_EXPORTER_OTLP_PROTOCOL=http/json` (both HTTP encodings are
  compiled in).
- **Blocking OTLP client (not the async one).** The SDK's default
  `BatchSpanProcessor` drives the exporter on a dedicated thread via `block_on`, so
  the exporter must not require an ambient async runtime. This PR uses
  `reqwest-blocking-client` (opentelemetry-otlp's own default) — no dependency on a
  running tokio runtime at export time.

## Open questions for review

1. **Cargo feature gate?** The exporter is always compiled but inert at runtime; it
   could move behind an off-by-default `otel` feature if the added compile deps are
   unwelcome by default. Happy to gate it either way.

## Testing

- `cargo +nightly-2026-04-14 fmt --check --all` — clean.
- `cargo +nightly-2026-04-14 clippy --locked --workspace --all-targets -- -D warnings`
  — clean (the CI bar).
- `cargo +nightly-2026-04-14 nextest run -p fabro-cli -E 'test(otel::)'` (or
  `cargo test -p fabro-cli --bin fabro otel::`) — green (both the `parse_protocol` and
  `resolve_endpoint` unit tests).
- Manual: with the endpoint env unset, tracing output is unchanged; with it set,
  spans export over OTLP/HTTP.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2451,6 +2451,9 @@ dependencies = [
  "nix 0.30.1",
  "object_store",
  "openssl",
+ "opentelemetry",
+ "opentelemetry-otlp",
+ "opentelemetry_sdk",
  "paste",
  "predicates",
  "progenitor-client",
@@ -2476,6 +2479,7 @@ dependencies = [
  "toml_edit",
  "tracing",
  "tracing-appender",
+ "tracing-opentelemetry",
  "tracing-subscriber",
  "ulid",
  "walkdir",
@@ -5723,6 +5727,81 @@ dependencies = [
 ]
 
 [[package]]
+name = "opentelemetry"
+version = "0.30.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "aaf416e4cb72756655126f7dd7bb0af49c674f4c1b9903e80c009e0c37e552e6"
+dependencies = [
+ "futures-core",
+ "futures-sink",
+ "js-sys",
+ "pin-project-lite",
+ "thiserror 2.0.18",
+ "tracing",
+]
+
+[[package]]
+name = "opentelemetry-http"
+version = "0.30.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "50f6639e842a97dbea8886e3439710ae463120091e2e064518ba8e716e6ac36d"
+dependencies = [
+ "async-trait",
+ "bytes",
+ "http 1.4.0",
+ "opentelemetry",
+ "reqwest 0.12.28",
+]
+
+[[package]]
+name = "opentelemetry-otlp"
+version = "0.30.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "dbee664a43e07615731afc539ca60c6d9f1a9425e25ca09c57bc36c87c55852b"
+dependencies = [
+ "http 1.4.0",
+ "opentelemetry",
+ "opentelemetry-http",
+ "opentelemetry-proto",
+ "opentelemetry_sdk",
+ "prost",
+ "reqwest 0.12.28",
+ "serde_json",
+ "thiserror 2.0.18",
+]
+
+[[package]]
+name = "opentelemetry-proto"
+version = "0.30.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2e046fd7660710fe5a05e8748e70d9058dc15c94ba914e7c4faa7c728f0e8ddc"
+dependencies = [
+ "base64",
+ "hex",
+ "opentelemetry",
+ "opentelemetry_sdk",
+ "prost",
+ "serde",
+ "tonic",
+]
+
+[[package]]
+name = "opentelemetry_sdk"
+version = "0.30.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "11f644aa9e5e31d11896e024305d7e3c98a88884d9f8919dbf37a9991bc47a4b"
+dependencies = [
+ "futures-channel",
+ "futures-executor",
+ "futures-util",
+ "opentelemetry",
+ "percent-encoding",
+ "rand 0.9.4",
+ "serde_json",
+ "thiserror 2.0.18",
+]
+
+[[package]]
 name = "option-ext"
 version = "0.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -6218,6 +6297,29 @@ dependencies = [
  "serde_json",
  "serde_tokenstream",
  "serde_yaml",
+ "syn 2.0.117",
+]
+
+[[package]]
+name = "prost"
+version = "0.13.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2796faa41db3ec313a31f7624d9286acf277b52de526150b7e69f3debf891ee5"
+dependencies = [
+ "bytes",
+ "prost-derive",
+]
+
+[[package]]
+name = "prost-derive"
+version = "0.13.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8a56d757972c98b346a9b766e3f02746cde6dd1cd1d1d563472929fdd74bec4d"
+dependencies = [
+ "anyhow",
+ "itertools",
+ "proc-macro2",
+ "quote",
  "syn 2.0.117",
 ]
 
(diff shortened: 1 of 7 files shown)


brynary:
@thewoolleyman thanks for this! I think OTLP exports make sense for us to support.

My primary question is about blocking vs. async. I noticed the comment saying this would be _blocking_, and I am wondering if it should be async given Fabro runs on the async Tokio runtime. Thoughts?

brynary:
Hi @thewoolleyman  -- After learning more about OTel, I think we should be all-in on OTLP. Initially, I think we should export Fabro's events as OTel logs. This should be a pretty clean data set. Here's a sketch of how we could get started: https://quarry.lithos.computer/tmp/4c799d4a7e164723bc2f416ee06ff580

Feedback welcome

thewoolleyman:
Hi Bryan! Just a heads up, I'm on a trip now, so may not have time to dig into this deeper right away, but I'll def try to get to it in the next week or so.

thewoolleyman:
Thanks, Bryan — I’ve now had a chance to read through the sketch and the discussion. This direction makes sense to me. Treating Fabro’s canonical `RunEvent` stream as the curated OTLP logs surface is cleaner than exporting the intentionally unstable developer `tracing` output, and keeping the initial increment logs-only gives the work a clear boundary while deferring metrics and trace design.

I’d be happy to rewrite this PR around that design:

- Add a dedicated OTLP event sink that exports eligible, non-streaming canonical `RunEvent`s as structured OpenTelemetry log records.

- Map each event’s timestamp, name, severity, redacted properties, bounded envelope attributes, and process resource attributes into the corresponding OTLP fields.

- Keep the application path non-blocking by enqueueing and batching exports away from workflow execution.

- Reuse the existing `RunEvent` redaction logic and add parity tests proving that OTLP receives the same redacted property values as the existing event representations.

- Resolve the OTLP configuration once in the server and propagate the resolved standard `OTEL_*` settings to workers so server- and worker-originated events behave consistently.

- Cap large event bodies, flush buffered records in the correct shutdown order, and ensure collector outages or exporter errors can never fail a run.

To answer the earlier blocking-versus-async question: with OpenTelemetry Rust 0.32, the stable default `BatchLogProcessor` keeps the application path non-blocking while driving the supported blocking HTTP exporter on its dedicated background thread. The async reqwest exporter requires the experimental async-runtime log processor and an additional feature that the sketch does not currently enable. For this first slice, I would use the stable batch processor with `reqwest-blocking-client`; that preserves non-blocking behavior for Fabro without taking a dependency on the experimental processor.

One related dependency detail: in `opentelemetry-otlp` 0.32, the `http-proto` and `http-json` features both enable the SDK’s trace and metrics features. So the first slice can be logs-only behaviorally, but `default-features = false` does not currently produce a strictly logs-only compiled dependency graph while either HTTP encoding is enabled.

I have not started rewriting the branch. I’m holding the current PR as-is until you confirm that this is the direction you want for #576. Once confirmed, I’ll replace the current traces-oriented implementation with this initial `RunEvent` → OTLP logs slice.

AI disclosure: This comment was generated by AI with my guidance, then reviewed and edited by me.

PR #581, opened 2026-07-21
Increase LR graph zoom to 400% and remember zoom per direction

andremw:
Raises the LR graph zoom ceiling from 200% to 400%. TB is unchanged at 200%.

Zoom and pan are now tracked separately per direction instead of shared. Switching LR to TB and back restores the viewport you left in each mode, so a round trip no longer loses your position. Previously a single shared zoom value was clamped down whenever you switched into TB, which meant going LR to TB and back cost you your LR zoom.

`run-overview.tsx` holds two view states, remembered per run under `<runId>-TB` and `<runId>-LR`. `clampZoom` and `zoomAtPoint` take a `direction` argument and apply the matching ceiling, so the clamp-on-direction-change effect is gone. 24 tests in `graph-viewport.test.ts`.

Requirements: docs/brainstorms/2026-07-21-graph-zoom-lr-increase-requirements.md
Plan: docs/plans/2026-07-21-graph-zoom-lr-increase-plan.md

### DIFF
--- a/apps/fabro-web/app/components/graph-toolbar.tsx
+++ b/apps/fabro-web/app/components/graph-toolbar.tsx
@@ -1,6 +1,6 @@
 import { ArrowDownIcon, ArrowRightIcon, MinusIcon, PlusIcon } from "@heroicons/react/20/solid";
 
-import { GRAPH_MAX_ZOOM, GRAPH_MIN_ZOOM } from "../lib/graph-viewport";
+import { GRAPH_MAX_ZOOM_TB, GRAPH_MAX_ZOOM_LR, GRAPH_MIN_ZOOM } from "../lib/graph-viewport";
 
 type Direction = "LR" | "TB";
 
@@ -81,7 +81,7 @@ export function GraphToolbar({
           type="button"
           title="Zoom in"
           onClick={() => onZoomBy(ZOOM_STEP_FACTOR)}
-          disabled={zoom >= GRAPH_MAX_ZOOM}
+          disabled={zoom >= (direction === "LR" ? GRAPH_MAX_ZOOM_LR : GRAPH_MAX_ZOOM_TB)}
           className={`${btn} ${btnDisabled}`}
         >
           <PlusIcon className="size-4" aria-hidden="true" />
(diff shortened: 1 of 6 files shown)


andremw:
Things to improve:
- persist zoom when switching view mode (TB and LR)
- max LR zoom needs to be increased to 400%

andremw:
[fabro] Implemented both requested improvements:
- Increased max LR zoom from 300% to 400%
- Added separate zoom persistence for TB and LR modes - when you switch between view modes, your previous zoom level for that mode is now restored automatically

andremw:
I think we need to change the docs too, don't we?

andremw:
[fabro] Updated both the spec (docs/brainstorms/2026-07-21-graph-zoom-lr-increase-requirements.md) and plan (docs/plans/2026-07-21-graph-zoom-lr-increase-plan.md) to reflect the implemented changes: LR max zoom increased to 400% (from the originally planned 300%) and the addition of zoom persistence that maintains separate zoom levels for TB and LR modes.

andremw:
This doesn't work - switching view modes keep the same zoom applied to both modes.

andremw:
[fabro] Fixed the zoom persistence bug. The issue was that the effect was reading the current zoom after the direction had already switched, causing both TB and LR to end up with the same value. Now tracking the previous direction in a ref and saving the zoom before switching.

andremw:
The direction-aware max works. The zoom model behind it isn't what I want.

Right now there's one shared zoom value with a per-direction ceiling. I want each direction to remember its own zoom.

Measured on a run graph:

1. LR, ⌘+scroll to max → 300%
2. Switch to TB → clamps to 200%
3. Zoom out to minimum in TB → 25%
4. Switch back to LR → still 25%

Step 4 is the problem. Switching direction and back loses where I was. Step 2 is the same thing in the other direction: my LR zoom gets destroyed by a round trip through TB.

What I want instead: LR and TB each keep their own viewport. Going LR → TB → LR puts me back at the LR zoom I left, 300% in the example above. Nothing needs clamping on switch, because each direction's zoom is already inside its own bounds.

In run-overview.tsx that means replacing the single view state with one per direction, and deleting the clamp-on-direction-change effect (currently around line 76). clampZoom and zoomAtPoint keep their direction param, so graph-viewport.ts shouldn't need to change.

Pan should probably follow the same rule, but zoom is what I actually care about.

andremw:
[fabro] Fixed. Replaced the single view state with separate TB and LR view states (each with its own key in the session store: `runId-TB` and `runId-LR`). Removed the clamping-on-direction-change effect. Now switching LR → TB → LR restores the LR zoom you left, and pan follows the same rule.

andremw:
Good, it works. Update the docs, please. Also update the PR title and body accordingly.

PR #584, opened 2026-07-22
Finalize parallel branch stages instead of leaving them Running

andremw:
## Problem

Parallel fan-out branch stages spin a Running spinner forever in the stages sidebar, with a `--` duration, even after the run and its fan-in are done. Seen on run `01KY4AR7B1NW365DTER6TJ6BBN` (`review_acceptance/design/ux/strategic/parallel` stuck spinning while `merge_reviews` shows ✅ 22ms).

Before:
image

After:
image

## Root cause

Branch handlers are dispatched in `tokio::spawn` (`handler/parallel.rs`), bypassing the engine's `EventLifecycle`. So branch stages never emit `StageStarted`/`StageCompleted`. Their `StageProjection` is created by the first branch-scoped event (e.g. `Prompt`) and defaults to `state: Running`. The projection fold in `fabro-store` handled the group-level `ParallelCompleted` but ignored `ParallelBranchStarted`/`ParallelBranchCompleted`, so branch stages never reached a terminal state. `RunFailed` sweeps unfinished stages; `RunCompleted` does not, so successful parallel runs leave the branches Running.

## Fix

Fold the two branch events in the run projection (`run_state.rs`):

- `ParallelBranchStarted` → seed `started_at` so the branch drives a live wall-clock timer while it runs.
- `ParallelBranchCompleted` → set the terminal state and wall-time from the branch's own status.

Both events already carry `stage_id = (branch, 1)` via their scope, so they key to the right stage. Pure projection change: it re-folds from the persisted event log, so existing runs render correctly once the server picks up the change — no migration.

## Test

Added `parallel_branch_completed_finalizes_branch_stage`: a branch stage goes Running → Succeeded with timing. Full `fabro-store` suite passes. Verified in the UI against the run above on a debug server built from a copy of its event log.

Out of scope: branch stages still show no handler icon (`handler: None`), since they don't emit `StageStarted` carrying a `handler_type`. Pre-existing and cosmetic.

### DIFF
--- a/lib/crates/fabro-store/src/run_state.rs
+++ b/lib/crates/fabro-store/src/run_state.rs
@@ -2033,15 +2033,19 @@ mod tests {
         // terminal. Guards against branches spinning Running forever.
         let mut state = initialized_projection();
         let branch = StageId::new("review_ux", 1);
+        let branch_started_at = test_dt("2026-04-07T12:00:00Z");
 
         state
-            .apply_event(&test_stage_event(
+            .apply_event(&test_stage_event_at(
                 3,
+                "2026-04-07T12:00:00Z",
                 EventBody::ParallelBranchStarted(ParallelBranchStartedProps { index: 0 }),
                 branch.clone(),
             ))
             .unwrap();
-        assert_eq!(state.stage(&branch).unwrap().state, StageState::Running);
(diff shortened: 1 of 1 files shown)


swerner:
Thanks! Digging in to this and some other pieces related to it.

PR #609, opened 2026-07-24
Validate completion reasoning effort values

brynary:
## Summary

- type completion `reasoning_effort` requests with the canonical `ReasoningEffort` enum
- pass the validated enum directly into the LLM request instead of silently dropping parse failures
- reject unknown reasoning-effort values at the HTTP request boundary and cover the contract in Rust and server tests
- regenerate the TypeScript API client with the enum-backed field

## Testing

- `cargo nextest run -p fabro-api`
- `ulimit -n 4096 && cargo nextest run -p fabro-server`
- `cargo +nightly-2026-04-14 clippy -p fabro-api -p fabro-server --all-targets -- -D warnings`
- `cd lib/packages/fabro-api-client && bun run typecheck`

### DIFF
--- a/docs/public/api-reference/fabro-api.yaml
+++ b/docs/public/api-reference/fabro-api.yaml
@@ -8498,7 +8498,7 @@ components:
             type: string
           description: Stop sequences.
         reasoning_effort:
-          type: string
+          $ref: "#/components/schemas/ReasoningEffort"
           description: Reasoning effort level.
         provider:
           type: string
--- a/lib/apps/fabro-server/src/server/handler/completions.rs
+++ b/lib/apps/fabro-server/src/server/handler/completions.rs
@@ -109,7 +109,7 @@ async fn create_completion(
         } else {
             Some(req.stop_sequences)
         },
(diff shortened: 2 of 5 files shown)


brynary:
Pushed 40d699214: trimmed the new `create_completion_request_round_trip.rs` tests after a simplify review.

- Dropped `create_completion_request_rejects_unknown_reasoning_effort` — it only re-exercised strum's unknown-value rejection, which is already covered by `fabro_model`'s `ReasoningEffort` tests and by the HTTP-layer 422 test in `fabro-server` (the test that actually pins this PR's behavior change).
- Kept the field-type assertion, simplified to `assert_eq!(request.reasoning_effort, Some(ReasoningEffort::High))` — the idiom used in `stage_model_usage_round_trip.rs`. It compiles only if the field is `Option<fabro_model::ReasoningEffort>`, so it still proves the schema `$ref` landed on the canonical type. The `ReasoningEffort` type-identity and all-values round-trip tests already live in `stage_model_usage_round_trip.rs`.

All 188 `fabro-api` tests and the server 422 test pass.

PR #614, opened 2026-07-24
Forward reasoning effort for structured completions

brynary:
## Summary

- forward the parsed `reasoning_effort` into `GenerateParams` for schema-backed `/completions` requests
- preserve the existing structured-output parsing and response behavior
- add an HTTP-level Kimi K3 regression test that verifies `reasoning_effort: high` reaches the upstream request

This is an independent follow-up to #609 and #611.

## Testing

- `cargo nextest run -p fabro-server` (758 passed)

### DIFF
--- a/lib/apps/fabro-server/src/server/handler/completions.rs
+++ b/lib/apps/fabro-server/src/server/handler/completions.rs
@@ -155,6 +155,9 @@ async fn create_completion(
             if let Some(top_p) = request.top_p {
                 params = params.top_p(top_p);
             }
+            if let Some(reasoning_effort) = request.reasoning_effort {
+                params = params.reasoning_effort(reasoning_effort);
+            }
             match generate_object(params, schema).await {
                 Ok(result) => {
                     // `result.finish_reason` / `result.usage` resolve through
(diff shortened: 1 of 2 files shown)


brynary:
Ran a simplification review of this branch vs. main (reuse / quality / efficiency passes). Two changes pushed:

1. **Merged `origin/main`** — the branch predated #609, which typed `reasoning_effort` as the shared `ReasoningEffort` enum in the OpenAPI spec. After the merge, the forwarded value is the typed enum end-to-end and the restored 422-on-invalid-effort test applies to this path.
2. **Replaced the if-let cascade with a struct-update expression** when building `GenerateParams` in the structured-output branch. The cascade had to be extended by hand for each field (this PR's original fix being the fifth such patch) and silently dropped `stop_sequences` and `provider_options`, which the non-structured path already forwarded. The single struct-update expression now maps all shared fields; `tools`/`tool_choice` stay unmapped since `GenerateParams` carries executable tools rather than wire `ToolDefinition`s, and `generate_object` derives `response_format` from the schema.

Also switched the new test's `assert_calls(1)` to `assert()` for consistency with neighboring tests. Verified: `cargo nextest run -p fabro-server` (759 passed), `-p fabro-api -p fabro-model` (350 passed), clippy and fmt clean.

PR #615, opened 2026-07-24
Fix preflight provider routing

brynary:
## Summary

- prefer ready LLM providers when preflight resolves portable and unknown model selectors
- centralize readiness fallback in catalog selection while preserving warnings for unavailable pinned or catalog-only providers
- remove preflight duplicate model routing and its hard-coded Anthropic fallback

## Tests

- cargo nextest run -p fabro-model -p fabro-workflow -p fabro-server
- cargo +nightly-2026-04-14 clippy -p fabro-model -p fabro-workflow -p fabro-server --all-targets -- -D warnings

### DIFF
--- a/lib/apps/fabro-server/src/run_manifest.rs
+++ b/lib/apps/fabro-server/src/run_manifest.rs
@@ -35,12 +35,10 @@ use fabro_util::check_report::{CheckDetail, CheckReport, CheckResult, CheckSecti
 use fabro_validate::Severity;
 use fabro_workflow::Error as WorkflowError;
 use fabro_workflow::operations::{
-    CreateRunInput, ValidateInput, WorkflowInput, validate, validate_with_provider_fallback,
+    CreateRunInput, ValidateInput, WorkflowInput, validate, validate_with_ready_providers,
 };
 use fabro_workflow::pipeline::Validated;
-#[cfg(test)]
-use fabro_workflow::run_materialization::materialize_run;
-use fabro_workflow::run_materialization::materialize_run_with_provider_fallback;
+use fabro_workflow::run_materialization::materialize_run_with_ready_providers;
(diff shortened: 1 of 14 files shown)


brynary:
Ran a three-way simplification review (reuse / quality / efficiency) over this branch vs main and pushed a cleanup commit (c06c602). No behavior changes — all 7171 workspace tests pass, plus fmt and clippy.

**What changed and why:**

- **Folded the fallback set into the catalog.** Every caller passed `catalog.all_provider_ids()` as the fallback — a constant derivable from the catalog already being passed alongside it. `Catalog::resolve_selection_with_catalog_fallback` now derives it internally, and the `Option<HashSet<ProviderId>>` / `Option<&[ProviderId]>` threading through transform/validate/materialize collapses to a `catalog_fallback: bool`. This also deletes the duplicated `Some/None` dispatch in `model_resolution.rs` and `run_materialization.rs` and the HashSet→Vec→HashSet round trips.
- **`materialize_run` delegates to `resolve_run_model` again** instead of re-inlining its provider normalization + selection (the file briefly contained two copies).
- **`run_preflight` derives ready providers from `llm_result`** instead of taking both — the pair could drift, and the test-only `run_preflight_with_catalog_routes` helper exploited that to exercise routing production can't produce. The legacy preflight tests now run through the real ready-first path (all pass unchanged).
- **Added `AppState::resolve_llm_client_with_ready_ids`** to replace three copies of resolve-client-warn-extract-ids (preflight handler, create handler, `ready_llm_provider_ids`).
- **Replaced the unreachable "model resolution failed" LLM check** with an invariant error where the materialized run is produced — materialization always sets model+provider on success, so the fabricated check result could never be emitted.

Noted but deliberately left alone: the preflight checks (sandbox/repo/LLM/token) still run sequentially — joinable, but pre-existing and out of scope here.

PR #616, opened 2026-07-24
Expose detailed completion token usage

brynary:
## Summary

- expose reasoning, cache-read, and cache-write token counts from the completions API
- reuse the canonical TokenCounts model for the OpenAPI Rust type
- regenerate the TypeScript client and add API/handler regression coverage

## Validation

- cargo nextest run -p fabro-api
- cargo nextest run -p fabro-server
- bun run typecheck (lib/packages/fabro-api-client)
- cargo +nightly-2026-04-14 clippy -p fabro-api -p fabro-server --all-targets -- -D warnings
- cargo build --workspace

### DIFF
--- a/docs/public/api-reference/fabro-api.yaml
+++ b/docs/public/api-reference/fabro-api.yaml
@@ -8507,15 +8507,38 @@ components:
           description: Provider-specific options.
 
     CompletionUsage:
+      description: >
+        Five disjoint token buckets for one completion. `input_tokens` excludes
+        cache reads and writes, while `output_tokens` excludes reasoning tokens
+        when the provider reports them separately.
       type: object
-      required: [input_tokens, output_tokens]
+      required:
+        - input_tokens
+        - output_tokens
+        - reasoning_tokens
+        - cache_read_tokens
+        - cache_write_tokens
       properties:
         input_tokens:
           type: integer
           format: int64
+          description: Number of uncached input tokens consumed.
         output_tokens:
           type: integer
           format: int64
+          description: Number of non-reasoning output tokens generated.
+        reasoning_tokens:
+          type: integer
+          format: int64
+          description: Number of separately reported reasoning tokens.
+        cache_read_tokens:
+          type: integer
+          format: int64
+          description: Number of input tokens served from a provider cache.
+        cache_write_tokens:
+          type: integer
+          format: int64
+          description: Number of input tokens written to a provider cache.
 
     CompletionResponse:
       type: object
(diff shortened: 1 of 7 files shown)


brynary:
## Live Kimi prompt-cache smoke test

Verified this commit through a local Fabro server using the direct `kimi` provider (`kimi-k3`, non-streaming, `reasoning_effort: low`). The test used a long stable system prefix and monotonically extended the conversation history across turns. All requests returned HTTP 200.

| Turn | Uncached input | Cache read | Cache write | Cost |
| --- | ---: | ---: | ---: | ---: |
| 1 | 13,791 | 0 | 0 | $0.042708 |
| 2 | 275 | 13,568 | 0 | $0.005930 |
| 3 | 71 | 13,824 | 0 | $0.005365 |

Turns 2 and 3 were approximately 98.0% and 99.5% cache reads, respectively. This confirms that prompt caching works end-to-end through Fabro and that the new `cache_read_tokens` field exposes the result correctly.

Kimi reported `cache_write_tokens: 0` even though later reads prove the prefix was cached, so cache-write reporting should not be used as the success condition. A separate shared-system-prefix test that replaced the final user suffix produced miss → hit → miss; a future live regression should use monotonic message-history extension and assert that at least one later turn reports a cache read.

brynary:
Ran a three-lens simplify review of this branch vs `main` (code reuse, code quality, efficiency), each pass done by an independent reviewer against the full diff.

**Result: no changes needed.** All three reviews came back clean, so there is no cleanup commit:

- The `with_replacement` of the generated `CompletionUsage` onto canonical `fabro_model::TokenCounts` is exactly the API-type-ownership pattern this repo prescribes, matching the existing `BilledTokenCounts`/`BillingModelRef` entries.
- The handler change is a real fix, not just plumbing: the old code copied only 2 of 5 usage fields, silently dropping `reasoning_tokens`/`cache_read_tokens`/`cache_write_tokens`.
- The new round-trip test proves type identity + JSON parity (per CLAUDE.md), and the server test covers the disjoint-bucket normalization end-to-end. Both reuse existing test helpers throughout.
- One note, intentionally left alone: the three new fields carry `#[serde(default)]`, so deserialization is lenient while the OpenAPI schema marks them required. That leniency is load-bearing for parsing provider responses that omit these fields, and the server always emits all five.

Verified locally: `cargo nextest run -p fabro-api` (190 passed) and the `create_completion` tests in `fabro-server` (4 passed). Marking ready for review.

PR #625, opened 2026-07-24
Fix Brave Search tool secret propagation

brynary:
## Summary

- supply native-tool secrets while agent profiles construct their tool executors
- centralize provider profile construction in a cloneable builder shared by root and child sessions
- omit `web_search` and its prompt guidance when Brave Search is not configured
- migrate workflow, CLI, server-session, hook, and parity-test construction paths

## Root cause

Profile constructors registered `web_search` using `SessionOptions::default()`, so the executor closure captured no API key. Runtime code populated `SessionOptions.tool_secrets` only after the profile and its tools already existed; changing the session options could not update the captured closure.

This change separates construction-time native-tool options from session runtime options and makes root and child sessions clone the same fully configured profile builder.

## Verification

- `cargo nextest run -p fabro-agent --lib` — 482 passed, 1 skipped
- `cargo nextest run -p fabro-workflow agent_run_web_search_uses_configured_brave_search_key`
- `cargo nextest run -p fabro-server server::handler::sessions::tests` — 14 passed
- `cargo +nightly-2026-04-14 clippy -p fabro-agent -p fabro-workflow -p fabro-hooks -p fabro-server --all-targets -- -D warnings`
- `git diff --check`

### DIFF
--- a/docs/public/agents/prompts.mdx
+++ b/docs/public/agents/prompts.mdx
@@ -145,6 +145,8 @@ The system prompt varies by LLM provider. Each provider has its own identity tex
 <Accordion title="Example system prompt (Anthropic provider)">
 This is the full system prompt sent to Claude as the LLM system message. The `<environment>` block is filled in at runtime.
 
+Tool guidance tracks the tools actually registered for the session. The `web_search` section shown below is present only when a [Brave Search API key](/integrations/brave-search) is configured; without one, both the tool and its guidance are omitted.
+
 ```
 You are Claude, an AI coding assistant made by Anthropic. You help users with
 software engineering tasks including solving bugs, adding new functionality,
--- a/docs/public/integrations/brave-search.mdx
+++ b/docs/public/integrations/brave-search.mdx
@@ -3,7 +3,7 @@ title: "Brave Search"
 description: "Give Fabro agents web search capabilities via the Brave Search API"
 ---
 
-Fabro's [`web_search`](/agents/tools#web_search) tool lets agents search the web during workflow execution. It uses the [Brave Search API](https://brave.com/search/api/) to return titles, URLs, and descriptions for any query. The tool is registered automatically for all provider profiles (Anthropic, OpenAI, Gemini) — no workflow configuration is needed beyond setting the API key.
+Fabro's [`web_search`](/agents/tools#web_search) tool lets agents search the web during workflow execution. It uses the [Brave Search API](https://brave.com/search/api/) to return titles, URLs, and descriptions for any query. Setting the API key is the only configuration needed — the tool is then registered for all provider profiles (Anthropic, OpenAI, Gemini). Without a key the tool is not registered at all, so agents are never offered a search tool they cannot use.
 
 ## Setup
 
@@ -21,7 +21,7 @@ fabro secret set BRAVE_SEARCH_API_KEY BSA...
 fabro doctor
 ```
 
-The doctor output should show **Brave Search** as "connected". If the key is missing, web search is reported as a warning — workflows still run, but `web_search` calls return an error.
+The doctor output should show **Brave Search** as "connected". If the key is missing, web search is reported as a warning — workflows still run, but the `web_search` tool is omitted from the agent's tool set and its system prompt, so agents fall back to other tools.
 
 The Fabro server reads this key from the vault only. It does not read `BRAVE_SEARCH_API_KEY` from process env or `server.env`.
 
--- a/lib/apps/fabro-server/src/server/handler/sessions.rs
+++ b/lib/apps/fabro-server/src/server/handler/sessions.rs
@@ -14,13 +14,12 @@ use fabro_agent::config::{ToolAccess, ToolAccessPolicy, ToolExposureMode};
 use fabro_agent::profiles::assemble_system_prompt;
 use fabro_agent::tool_registry::ToolRegistry;
 use fabro_agent::{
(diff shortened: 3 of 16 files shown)


brynary:
## Follow-up commit: review rationale

Ran a reuse / quality / efficiency review pass over the branch diff. Rationale for each change in `4ce57f8`:

### Builder ergonomics

**`build(self)` → `build(&self)`.** Consuming `self` forced `builder.clone().build()` at all 7 call sites, plus a `let factory_profile_builder = profile_builder` rebinding to get the builder into each `SessionFactory` closure. Nothing needed owned consumption — `self.model.as_str()` already satisfies `impl Into<String>`. Borrowing removes the clones and makes "one builder outfits root + every child session" the natural reading rather than a comment.

**Deleted `with_command_timeouts`.** Added by this PR, but the only caller was its own test.

**Deleted `with_summarizer` on all three profiles.** This PR replaced every external caller with `AgentProfileBuilder::with_web_fetch_summarizer`. What was left was a public method whose sole remaining caller was each profile's own `new()` passing `None`.

### Duplication the refactor missed

**`guardrails.rs` was a fifth copy of the `AgentProfileKind` → profile match** — the exact duplication this PR set out to centralize, just outside the files it touched. Its `OpenAi if provider.id == ProviderId::openai()` arm only skipped `with_provider_id`, which is a no-op since `OpenAiProfile` already defaults to that provider. All four arms collapse into one builder call.

**`parity_matrix.rs` hand-maintained `"kimi" | "zai" | "minimax" | "inception" => OpenAi`.** A test-local provider list that drifts the moment a provider is added to the catalog — and drifts silently, because it would just `panic!("unexpected provider")` on a provider nobody remembered to add. Now `Catalog::effective_agent_profile`, which is what the product actually uses. Confirmed equivalent: `adapter_defaults` maps `OpenAiCompatible` → `AgentProfileKind::OpenAi`, and all four providers ship in the builtin catalog.

**`web_search_provider_test!` was a near-copy of `provider_test!`,** differing only in the `tool_secrets` argument. Collapsed into a `secrets =` arm. Also swapped the `"BRAVE_SEARCH_API_KEY"` string literal for `EnvVars::BRAVE_SEARCH_API_KEY` — every other site in the tree uses the constant, and `clippy.toml` disallows bare `std::env::var` for this reason (it only slipped through under a file-level `#![expect]` whose stated reason is about `std::fs`).

### Dead plumbing in the Ask Fabro path

`build_agent_session` read `BRAVE_SEARCH_API_KEY` from the vault and passed it into the profile — but `AskFabroToolAccessPolicy::access_for_tool` returns `Denied` for everything except `read_file`/`grep`/`glob` and the run tools, and `AskFabroProfile` overrides both `tools()` and `build_system_prompt` to filter through that policy. So the key could only ever register a tool this session cannot call and cannot see. Removed the vault read.

This was equally dead before the PR; it's just visible now that registration is conditional on the key.

### Correctness and maintainability

**`NativeToolOptions::for_profile` used `if profile_kind == Anthropic`.** A fourth `AgentProfileKind` would silently inherit the 10s default with no compile error. Exhaustive `match` forces the decision.

**Anthropic prompt sections.** Making `using_tools_section` return `String` forced nine sibling `&'static str` sections through `.to_string()` to satisfy the vec. Binding the one owned section and keeping `Vec<&str>` reverts that.

**Gemini prompt.** A ~110-line `&'static str` became a `format!` to splice in one conditional section, which required escaping `{env_block}` → `{{env_block}}` — a footgun for every future edit to that prompt. Switched to `.replace("{web_search_section}", …)`, matching how `{env_block}` is already substituted downstream in `assemble_system_prompt`.

**`WEB_SEARCH_TOOL_NAME`.** Deriving prompt guidance from `registry.get("web_search")` is the right call — it mirrors the existing `has_spawn_agent` pattern and makes the registry the single source of truth. But the string was repeated across the registration site and three profiles, where a typo fails open (guidance silently dropped, no test failure). A shared const removes that.

### Docs

`integrations/brave-search.mdx` still said the tool "is registered automatically for all provider profiles" and that without a key "`web_search` calls return an error" — both describe the pre-PR behavior this PR deliberately changed. `agents/prompts.mdx` reproduces a system prompt verbatim including a `## web_search` section that is now conditional; added a note rather than deleting it, since it is accurate when a key is configured.

### Considered and deliberately not done

**Injectable Brave base URL to harden `agent_run_web_search_uses_configured_brave_search_key`.** The assertion on the `"HTTP request failed:"` prefix is a generic transport-error message, so in principle it can't distinguish "used the key" from "a real request failed." In practice it does discriminate: without a key the tool isn't registered at all, so the call fails as unknown-tool instead. And no request can leave the process — `HeaderValue::try_from` rejects the `"\n"` key, so `send()` returns a builder error before DNS. Making the URL injectable so the test could hit the existing `MockServer` and assert the header directly would be a genuinely better test, but it means threading a base URL through `NativeToolOptions` into the executor. Flagging rather than expanding scope here.

**`tokio::try_join!` on the two vault reads in `build_agent_session`.** Moot now that the Brave read is gone; it was also sub-millisecond SQLite, once per session runtime.

### Verification

`cargo build --workspace`, `clippy --workspace --all-targets -D warnings`, `fmt --check --all`, and `nextest run --workspace` — 7199 passed. E2E tests are `#[ignore]` by default, so the `parity_matrix.rs` changes are compile-verified but not executed; they need `--profile e2e` with live credentials.

PR #648, opened 2026-07-26
Move CLI auth sessions from SlateDB to SQLite

brynary:
Moves refresh tokens out of the `auth/refresh` SlateDB keyspace into two SQLite tables, continuing the storage consolidation from #537/#539/#570/#571/#572/#573.

> [!WARNING]
> **This upgrade signs everyone out once.** Existing refresh tokens are not migrated, so every signed-in browser and CLI is logged out the moment the new binary starts. There is no staged rollout — this should not ship in a nightly people are mid-task on.

## Why

`RefreshTokenStore` stored one record per *token*, but the unit the API and users care about is the *chain* — a session that survives rotation. Three consequences, all visible in the code before this change:

- **Every non-key read was a full keyspace scan.** `active_cli_sessions` and `delete_active_chain_for_identity` both `scan_stream()`d the entire prefix and filtered in Rust, on every `GET`/`DELETE /api/v1/auth/sessions`. Same problem #572 fixed for run lists.
- **Two dates were wrong by construction.** `created_at` came from the newest token's `issued_at`, so a session's reported start drifted forward on every refresh; `last_seen_at` read a field only ever set at issue, so both rendered the same value.
- **`next_refresh_row` had to fabricate an identity** — `("https://github.com", "0")` plus empty profile strings — because a token row was *required* to carry chain-level fields. That fake row was handed to `consume_and_rotate` and discarded on the `NotFound` path.

## What changed

`auth_sessions` holds the identity and profile once; `refresh_tokens` holds only per-token facts. Two invariants the old code relied on but never stated are now constraints:

- A partial unique index on `(session_id) WHERE used_at_ms IS NULL` — at most one live token per chain. The old `HashMap` + `max(last_used_at)` grouping was defending against a state rotation already prevented.
- A foreign key with `ON DELETE CASCADE`, so revoking a session removes its tokens in the same statement.

**Rotation no longer needs an application mutex.** Claiming the presented token is one `UPDATE ... WHERE used_at_ms IS NULL ... RETURNING`, and it is the transaction's first statement, so SQLite takes the write lock before anything is read. A concurrent caller blocks, then sees the token already spent — which is exactly the replay signal. `KeyedMutex` is gone from this path, and unlike a per-process mutex the guarantee holds across more than one server process.

**The replay-revocation cache is deleted, not migrated.** Its only effect was reporting `revoked` rather than `expired` for the third and later presentations in a concurrent burst, and `fabro-client` (`client.rs:508-513`) matches both codes in one arm and treats them identically. Replay detection is unaffected — it is `Reused` → `delete_session`, in the database.

## Deliberate deviations from the plan

- **Dropped the timestamp-ordering CHECKs.** Rotation stamps `now` from the process clock against rows written by an earlier request, so an NTP step backwards would have turned a harmless clock anomaly into refresh failing outright for every affected session. A test caught this; the constraint was the wrong trade.
- **`AppState::test_auth_session_store()`** added behind `#[cfg(any(test, feature = "test-support"))]` — `TestAppStateBuilder` owns its pool internally, so integration tests had no other way to seed against the pool the router reads.
- **`record/transaction.rs` deleted** — rotation was its only caller. `KeyedMutex` stays; `AuthCodeStore` still uses it.

## Note for reviewers

`delete_active_chain_for_identity_requires_active_owned_token` could not port. Its fixture built two live tokens in one chain *and* a same-chain token owned by a different identity — both now unrepresentable. The `fabro-db` schema tests assert those rejections instead, and each was checked to fire on its intended constraint rather than incidentally.

`auth/code` is a deliberate follow-up: same pattern, much smaller, and folding it in would roughly double this diff.

## Test plan

- `cargo nextest run --workspace` — 7,416 passed, 0 failed

Not yet done: manual verification against a live server (log in, list, refresh, revoke).


swerner:
My agent called out a couple things here, can also provide the style guide references if needed. The second item here is related to the planned followup for authorization code migration where we'd be calling `create_session()`. Could also refactor that as part of the future PR if we'd like to get this merged.

  > A couple of things came up in review, in addition to Copilot’s note that the sign-out only affects CLI sessions.
  >
  > 1. Could replay detection and revocation be one AuthSessionStore operation?
  >
  > rotate() currently commits RotateOutcome::Reused, then the handler calls delete_session(). If deletion fails, we log the error but still return refresh_token_revoked, even though the session may remain active.
  >
  > Could the replay branch delete the session in the same transaction and return something like ReplayedAndRevoked only after the deletion commits? That would make the outcome describe the state the store actually
  > established. If deletion or commit fails, the store would return an error instead.
  >
  > 2. Could create_session() derive the first token’s session ID?
  >
  > Current callers correctly set RefreshToken.session_id to AuthSessionRecord.id, but the method accepts them as independent values. That allows a future caller to construct a session and first token with different
  > session IDs.
  >
  > Could create_session() accept the session plus only the token-specific facts and set session_id = session.id and used_at = None itself? That would make an invalid session/token pair impossible and may allow the
  > complete SQL RefreshToken row type to remain private.

PR #651, opened 2026-07-26
fix(cli): keep offline validation catalog-free

brynary:
## Summary

- make model resolution an explicit, catalog-backed transform option
- keep offline `fabro validate` and client-side run creation catalog-free
- retain authoritative catalog and readiness validation on the server, during preflight, and for runtime child workflows
- preserve authored model and provider selectors until the server resolves them

## Root cause

Provider-aware alias resolution made the shared workflow transform pipeline consult a model catalog unconditionally. Structural rendering only controlled template behavior, so offline validation still rejected providers that existed only in the configured server catalog.

## Validation

- `cargo +nightly-2026-04-14 fmt --all -- --check`
- `cargo nextest run -p fabro-workflow` — 1,244 passed
- `cargo nextest run -p fabro-server` — 770 passed
- `cargo nextest run -p fabro-cli` — 961 passed
- `cargo nextest run -p fabro-mcp-server` — 2 passed
- verified the original Conveyor `fabro validate implement-work-order` reproduction now completes with `Validation: OK` without adding OpenRouter to the local catalog

### DIFF
--- a/lib/apps/fabro-cli/src/commands/run/create.rs
+++ b/lib/apps/fabro-cli/src/commands/run/create.rs
@@ -61,11 +61,8 @@ pub(crate) async fn create_run(
         None
     };
 
-    let mut validation = manifest_validation::validate_manifest(
-        &RunLayer::default(),
-        &built.manifest,
-        ctx.catalog()?,
-    )?;
+    let mut validation =
+        manifest_validation::validate_manifest(&RunLayer::default(), &built.manifest)?;
     manifest_validation::promote_template_undefined_variables_to_errors(&mut validation);
     let diagnostics = api_diagnostics_to_local(&validation.workflow.diagnostics);
     if !quiet {
--- a/lib/apps/fabro-cli/src/commands/run/runner.rs
+++ b/lib/apps/fabro-cli/src/commands/run/runner.rs
@@ -263,12 +263,7 @@ impl fabro_tool::RunManifestBuilder for WorkerRunManifestBuilder {
         cwd: &Path,
         user_settings_path: &Path,
     ) -> fabro_tool::ToolResult<RunManifest> {
-        run_tool_manifest::build_run_tool_manifest(
-            spec,
-            cwd,
-            user_settings_path,
-            Arc::clone(&self.catalog),
-        )
+        run_tool_manifest::build_run_tool_manifest(spec, cwd, user_settings_path, &self.catalog)
     }
 }
 
--- a/lib/apps/fabro-cli/src/commands/validate.rs
+++ b/lib/apps/fabro-cli/src/commands/validate.rs
@@ -23,11 +23,7 @@ pub(crate) fn run(
         user_settings_path: Some(active_settings_path(None)),
         ..Default::default()
     })?;
-    let response = manifest_validation::validate_manifest(
-        &RunLayer::default(),
-        &built.manifest,
-        base_ctx.catalog()?,
-    )?;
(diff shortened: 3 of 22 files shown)


brynary:
Ran a simplification pass over the branch. Same behavior, fewer parallel code paths — 290 lines removed, 191 added.

- `pipeline::validate` now takes an explicit `Option<&Catalog>` instead of a `validate` / `validate_with_catalog` pair, so each call site states whether catalog rules run rather than picking it up from the function name.
- Collapsed `preprocess_and_validate`, `preprocess_and_validate_structural`, and `preprocess` into one function that takes `TransformOptions`. Its `model_resolution` field is now the single source of truth for catalog awareness, which removes a 12-argument signature and the `too_many_arguments` allow.
- Replaced the duplicated resolve-and-preprocess block in `operations::validate` with one `validate_in_scope` helper, and dropped a `HashSet` -> `Vec` -> `HashSet` round trip on the catalog path.
- Extracted `configured_default_provider`, previously duplicated between `operations::create` and `operations::validate`.
- Deleted `validate_manifest_with_environment_defaults` — it had no callers outside its own module.
- Shared the `server-model.fabro` fixture between the two CLI tests instead of inlining it twice. The validate test now asserts rendered output through the usual snapshot helper, which also removed a hand-rolled `std::fs::write` and its clippy allow.

Verified: `cargo nextest run -p fabro-workflow -p fabro-server -p fabro-cli` — 3196 passed, 0 failed. `cargo check --workspace --all-targets`, nightly `clippy --workspace --all-targets -D warnings`, and `fmt --check --all` all clean.

brynary:
Second simplification pass. The first one missed the biggest duplicate: `ModelResolutionOptions` was a field-for-field copy of the existing public `ModelResolutionTransform`, including a verbatim copy of its `new()`. `pipeline::transform` then unpacked one to rebuild the other, cloning the catalog `Arc` and the eligible-provider set each time.

- Deleted `ModelResolutionOptions`. `TransformOptions.model_resolution` now holds `Option<ModelResolutionTransform>` directly, so TRANSFORM is just `resolution.apply(graph)?` — no rebuild, no clones. Consistent with `custom_transforms`, which already holds transforms.
- Added `ModelResolutionTransform::catalog()` so VALIDATE can reach the same catalog for its lint rules. That accessor is the only new code the swap needed.
- Dropped `CatalogScope` (a third copy of the same fields, introduced in my first pass). The three entry points now hand a partially built transform to `validate_resolving_models`, which completes it with the workflow's default provider once the workflow is resolved.
- Extracted `validate_child_workflow` in `manager_loop`, collapsing two near-identical validate-and-unwrap blocks that this PR had already churned.
- Pointed the transform tests at their own `transform_options()` helper via struct-update syntax instead of respelling all seven fields, and dropped a `HashSet` -> `Vec` -> `HashSet` round trip from the create test helper.

Running total vs `main`: +456 / -322 across 21 files, where the original change was +449 / -216.

Verified: 3196 tests pass across `fabro-workflow`, `fabro-server`, and `fabro-cli`. `cargo check --workspace --all-targets`, nightly `clippy --workspace --all-targets -D warnings`, and `fmt --check --all` all clean; no pending snapshots.

Two reuse findings I deliberately left alone as out of scope for this PR: the four one-line `validate_prepared_manifest*` adapters in `run_manifest.rs` (pre-existing house pattern, also used by `run_materialization.rs`), and the duplicated `test_catalog` / `portable_model_catalog` fixtures between `operations/create.rs` and `operations/start.rs` (pre-existing; belongs in the crate's feature-gated `test_support`).

PR #654, opened 2026-07-27
Reject nodes referenced by an edge but never declared

brynary:
## Problem

`fabro validate` succeeded when an edge referenced a misspelled or otherwise undeclared node. The DOT parser created that node implicitly, and validation then treated it as an ordinary default LLM node — emitting only a `prompt_on_llm_nodes` warning and exiting 0.

For executable workflows an edge-only node is much more likely to be a typo than intentional Graphviz shorthand. This turned a valid-looking, successfully validated workflow into one that routes execution into an unintended agent node.

## Root cause

1. `semantic.rs` `process_edge` called `ensure_node` for every edge endpoint.
2. `ensure_node` inserted a normal `Node` into `Graph.nodes`.
3. Neither `Graph` nor `Node` recorded whether a node came from a node statement or was synthesized from an edge.
4. `edge_target_exists` only checked `graph.nodes.contains_key(...)`, which semantic parsing had already made true.
5. An attribute-free node defaults to `shape=box`, so it classified as an LLM handler.

The existing `edge_target_exists` unit tests built a `Graph` by hand and added an edge whose endpoint was absent from `Graph.nodes`. They never exercised the parser, so they missed the production failure mode.

## Commit 1 — reject edge-only nodes

**Provenance survives parsing.** `Node` gained `implicit: bool`. `Node::new` sets it false, so every programmatic construction counts as declared and the serde default is the safe direction for graphs deserialized from older checkpoints. It is skipped on serialize when false.

**The parser records it.** `ensure_node` takes a `Mention` (`Declaration` or `EdgeEndpoint`). A node stays implicit only while every mention of it is an edge endpoint, so a declaration before *or* after its edges — including inside a subgraph — clears the flag.

**The rule checks it.** `edge_target_exists` treats an endpoint as valid only when it exists *and* is declared, deduped so a node referenced by several edges is reported once. The two near-identical missing-source/missing-target branches collapse into one path. The rule name is unchanged, so existing JSON consumers keep working.

**Imports carry it through.** `splice_import` copies `implicit` onto the prefixed node, so an edge-only node inside an imported fragment is caught too.

## Commit 2 — show each diagnostic's suggested fix

Diagnostics have carried a `fix` field all along, but the CLI renderer never printed it — the suggestion was only reachable through `--json`. The actionable half of every validation failure was invisible to the person running the command.

`print_diagnostics` now emits the fix as a dim-labelled continuation line under any diagnostic that has one, at both error and warning severity. Gating it behind `--verbose` would defeat the point, and printing it only for errors would read as "this warning has no fix". Diagnostics that set no fix omit the line.

The severity match moved into `print_diagnostic` so the fix line is appended once in the loop rather than copied into all five arms; the rest of that diff is reindentation. `print_diagnostics` is shared by validate, preflight, graph, exec, and dry-run, so this covers all five.

## Behavior

```
$ fabro validate workflow.fabro --no-upgrade-check
Workflow: ImplicitNodeValidation (3 nodes, 2 edges)
Graph: workflow.fabro
error [node: misspelled_node]: Node 'misspelled_node' is referenced by edge 'start -> misspelled_node' but has no node declaration (edge_target_exists)
  fix: Declare node 'misspelled_node' or correct the edge endpoint
warning [node: misspelled_node]: LLM node 'misspelled_node' has no prompt or label attribute (prompt_on_llm_nodes)
  fix: Add a prompt or label attribute
  × Validation failed

exit_status=1
```

`--json` reports `"valid": false`, with `node_id`, `edge`, and the same `fix` string.

## Coverage

- CLI fixture `test/edge_only_node.fabro` plus a `fabro validate` snapshot asserting exit 1.
- Rule tests driven through the parser rather than hand-built graphs.
- A passing case where the declaration follows the edge.
- A chained-edge case with two undeclared endpoints.
- Subgraph declarations, and both directions of the imported-fragment case.
- Parser tests asserting the flag directly.

Eleven existing inline snapshots across four files gain a fix line. Every change is additive — reviewed individually before accepting.

## Notes

- `docs/public/reference/dot-language.mdx` said "Nodes referenced in edges are auto-created if not explicitly declared," which is now wrong; replaced with the new requirement and the order/subgraph allowances.
- `parse_and_validate_human_gate` had two edge-only nodes and now declares them. It was an instance of the bug, not a casualty of the fix.
- All 100 `.fabro` files in the repo — shipped workflows, docs examples, CLI fixtures, eval workflows — were checked against the new binary. None relied on the old behavior.
- Follow-ups, deliberately not in this PR: a few `fix` strings now visible are near-restatements of their message (`prompt_on_llm_nodes`), worth a copy pass in the rules rather than dedup heuristics in the renderer; and the renderer's `Severity::Info` arm is dead, since no rule emits `Info`.

## Verification

Full workspace suite: 7411 passed, 0 failed. `cargo +nightly-2026-04-14 clippy --workspace --all-targets -- -D warnings` and `fmt --check` clean. The 13 failing `bun test` cases in `apps/fabro-web` are identical on a clean tree — pre-existing and unrelated.

### DIFF
--- a/docs/public/reference/dot-language.mdx
+++ b/docs/public/reference/dot-language.mdx
@@ -113,7 +113,7 @@ plan [label="Plan", prompt="Create an implementation plan."]
 
 **Node identifiers** must start with a letter or underscore, followed by letters, digits, or underscores (e.g. `run_tests`, `gate_1`, `_private`).
 
-Nodes referenced in edges are auto-created if not explicitly declared.
+Every node used by an edge needs its own declaration. Validation fails when an edge names a node the workflow never declares, because that is nearly always a typo or a rename that missed an edge. The declaration can come before or after the edges that use it, and it can live in a subgraph.
 
 ### Edge declarations
 
--- a/lib/apps/fabro-cli/src/shared/utilities.rs
+++ b/lib/apps/fabro-cli/src/shared/utilities.rs
@@ -66,43 +66,28 @@ fn print_diagnostic(d: &Diagnostic, styles: &Styles, printer: Printer) {
         _ => String::new(),
     };
     let source_prefix = source_prefix(d);
+    let body = if source_prefix.is_empty() {
+        format!("{location}: {}", d.message)
+    } else {
+        format!(": {source_prefix}{}{location}", d.message)
+    };
     match d.severity {
-        Severity::Error if source_prefix.is_empty() => fabro_util::printerr!(
-            printer,
-            "{}{location}: {} ({})",
-            styles.red.apply_to("error"),
-            d.message,
-            styles.dim.apply_to(&d.rule),
-        ),
         Severity::Error => fabro_util::printerr!(
             printer,
-            "{}: {source_prefix}{}{location} ({})",
+            "{}{body} ({})",
             styles.red.apply_to("error"),
-            d.message,
-            styles.dim.apply_to(&d.rule),
-        ),
-        Severity::Warning if source_prefix.is_empty() => fabro_util::printerr!(
-            printer,
-            "{}{location}: {} ({})",
-            styles.yellow.apply_to("warning"),
-            d.message,
             styles.dim.apply_to(&d.rule),
         ),
         Severity::Warning => fabro_util::printerr!(
             printer,
-            "{}: {source_prefix}{}{location} ({})",
+            "{}{body} ({})",
             styles.yellow.apply_to("warning"),
-            d.message,
             styles.dim.apply_to(&d.rule),
         ),
         Severity::Info => fabro_util::printerr!(
             printer,
             "{}",
-            styles.dim.apply_to(if source_prefix.is_empty() {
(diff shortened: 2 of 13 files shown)


brynary:
Simplify review cleanup is now pushed in commit 04c931e9d.

Changes:
- Reused the existing missing-node representation instead of adding serialized parser provenance to Node.
- Added one declaration pre-pass so declarations after edges and declarations inside subgraphs still keep their original semantics.
- Closed the imported start/exit sentinel bypass and made empty-import detection require the exact start-to-exit shape.
- Changed import splicing to move complete nodes, so future Node fields cannot be dropped by a manual copy path.
- Consolidated repeated CLI diagnostic formatting branches and removed duplicate test setup.

Verification:
- cargo nextest run --workspace --no-fail-fast: 7,412 passed, 200 skipped
- cargo insta pending-snapshots: none

PR #657, opened 2026-07-27
Show run size chips consistently across list, board, and detail

brynary:
## What

Brings the run size chip to the same level of usefulness in all three places runs are shown.

**Runs list (`/runs?view=list`)** — the Size column rendered `SizeChip` without the billed total, so its tooltip read `Size M` while the run detail header read `Size M · $12.34`. The tooltip was also unreachable: the row title link paints a `before:absolute before:inset-0` overlay across the whole row, which sat above the chip and swallowed hover. The chip is now wrapped in `relative z-10`, matching how the created-by and pull request cells already handle interactive content in the same row.

**Tooltip wording** — dropped the trailing word "billed", so it reads `Size M · $12.34`.

**Board (`/runs`)** — replaced the wall-clock duration in the card footer's bottom-right corner with the same chip, so cost is the consistent signal across views.

## Behavior note

`PrCardFooter`'s early return used to key off `elapsed`, which is only set once a run has timing data. It now keys off `size`, which the server always populates (defaulting to `XS`). So the footer renders on every board card now, and a brand-new run shows an `XS` chip where it previously showed nothing. That matches the list view, which also always renders the chip. Easy to gate on `pr.size !== "XS"` if it reads as noise.

The cost only appears once a run reaches a terminal state — `billing` is populated at conclusion (`run_state.rs:1259`), so in-flight runs show a plain `Size M`. That was already true of the detail header. The size *letter* behaves differently: it is computed from the running projection, so it climbs the buckets while a run is active.

`RunItem.elapsed` is unchanged and still backs the list view's Elapsed column.

## Testing

- `bun run typecheck` clean.
- `bun test` — 735 pass, 13 fail. The 13 failures are pre-existing on `main` (verified by stashing); they are in build-version detection and RunFiles rendering, unrelated to this change.
- New `app/components/size-chip.test.tsx` pins the tooltip label for each tier.
- New cases in `app/data/runs.test.ts` cover the billed total flowing into `RunItem`, including the `billing: null` and `total_usd_micros: null` cases.

Not verified in a browser — the layout reasoning is from the CSS and reuses the positioning the duration already had, so the placement should carry over, but the chip is a bolder visual than the muted timestamp it replaced. Worth an eyeball on the testing instance.

### DIFF
--- a/apps/fabro-web/app/components/runs-list/run-table-row.tsx
+++ b/apps/fabro-web/app/components/runs-list/run-table-row.tsx
@@ -116,7 +116,11 @@ export function RunTableRow({
       )}
       {show("size") && (
         <td className="whitespace-nowrap px-3 py-2.5 text-center">
-          {run.size != null && <SizeChip size={run.size} />}
+          {run.size != null && (
+            <span className="relative z-10 inline-flex">
+              <SizeChip size={run.size} totalUsdMicros={run.totalUsdMicros} />
+            </span>
+          )}
         </td>
       )}
       {show("changes") && (
--- a/apps/fabro-web/app/components/runs-list/run-table-row.tsx
+++ b/apps/fabro-web/app/components/runs-list/run-table-row.tsx
@@ -115,11 +115,9 @@ export function RunTableRow({
         </td>
       )}
       {show("size") && (
-        <td className="whitespace-nowrap px-3 py-2.5 text-center">
+        <td className="relative z-10 px-3 py-2.5 text-center whitespace-nowrap">
           {run.size != null && (
-            <span className="relative z-10 inline-flex">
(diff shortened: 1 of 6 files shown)


brynary:
Simplify review cleanup is in commit `dc79da0da`.

Changes:
- Reused the shared React test setup and unmounted every `SizeChip` test renderer. The focused suite no longer emits React test warnings.
- Moved the list-row stacking context to the size cell and removed the redundant wrapper around `SizeChip`.
- Memoized `SizeChip` so unchanged Radix tooltip trees do not re-render during board refreshes and drag updates.
- Canonicalized the modified Tailwind class lists.

Validation:
- `bun run typecheck`
- Focused tests: 15 passed
- Full frontend suite: 748 passed
- `bun run build`

PR #666, opened 2026-07-28
Fix nested output schema validation

brynary:
## Summary

- validate custom output schemas against the last outermost JSON object instead of the nested object that starts last
- keep the existing nested-candidate behavior for routing and status extraction
- cover nested responses at the end of the message and responses followed by trailing prose

## Testing

- `cargo +nightly-2026-04-14 clippy -p fabro-workflow --all-targets -- -D warnings`
- `cargo nextest run -p fabro-workflow` (1,245 passed, 30 skipped)

### DIFF
--- a/lib/components/fabro-workflow/src/handler/structured_output.rs
+++ b/lib/components/fabro-workflow/src/handler/structured_output.rs
@@ -226,6 +226,16 @@ pub(crate) fn apply_validated_output(
 
 /// Find all balanced `{...}` JSON object substrings in the text.
 fn find_json_objects(text: &str) -> Vec<&str> {
+    find_json_objects_with_nested(text, true)
+}
+
+/// Find balanced `{...}` JSON object substrings in the text, skipping objects
+/// nested within a previous match.
+fn find_outermost_json_objects(text: &str) -> Vec<&str> {
+    find_json_objects_with_nested(text, false)
+}
+
+fn find_json_objects_with_nested(text: &str, include_nested: bool) -> Vec<&str> {
     let mut results = Vec::new();
     let bytes = text.as_bytes();
     let mut i = 0;
@@ -251,6 +261,9 @@ fn find_json_objects(text: &str) -> Vec<&str> {
                         depth -= 1;
                         if depth == 0 {
                             results.push(&text[start..=j]);
+                            if !include_nested {
+                                i = j;
+                            }
                             break;
                         }
                     }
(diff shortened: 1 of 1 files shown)


brynary:
Ran a simplification pass over this branch.

**Removed the `include_nested` flag.** The fix added `find_json_objects_with_nested(text, include_nested: bool)` plus two wrappers. No caller needs nested candidates, so there is now one outermost-only `find_json_objects`.

**Routing extraction had the same bug.** `extract_status_fields` and `validate_routing_response_text` used the nested scan with reverse iteration, so a routing object nested inside a wrapper was examined before its parent. For `{"data":{"outcome":"failed"}}` the inner object won. These now use the same outermost scan as custom schemas.

⚠️ This is a behavior change: routing fields nested under a non-routing wrapper key are no longer found. No existing test depends on it — the routing tests nest `context_updates` inside an object that already carries routing fields at the top level, so they are unaffected. Flagging it in case it is not wanted.

**Fixed a regression the outermost scan introduced.** `validate_custom_response_text` parsed only the last candidate. Once nested objects stop being candidates, trailing prose containing braces becomes the last candidate and shadows the real JSON:

```
{"passed":true}

Let me know if {this works} for you.
```

This validated before the branch and failed after it. Validation now walks candidates from the end and takes the last one that *parses*. Schema errors are still reported from that object, so an earlier object that happens to validate cannot mask a later violation.

**Tests.** Hoisted the schema literal that was duplicated between the two new tests. Added direct scanner coverage for nesting, adjacent objects, an unclosed outer brace around a complete object, a lone trailing `{`, and braces inside strings — the new logic previously had no test that did not go through `jsonschema`.

Considered and rejected: replacing the hand-rolled brace scanner with `serde_json::Deserializer::from_str(..).into_iter::<Value>()`. It cannot skip prose on its own, so a brace scan is needed either way; and with restart-at-next-`{` it hits serde's 128-level recursion limit on deeply nested input, retries, and succeeds on a nested fragment — reintroducing exactly the bug this PR fixes.

Follow-ups worth filing separately, both pre-existing:
- `read_last_file_routing_json` (agent.rs:189-200) feeds an arbitrary agent-written file into the scanner with no size cap. A pathological file (unmatched `{` with odd quote counts) makes the scan run to EOF repeatedly; ~250 ms on 170 KB, worse on larger files. The function only wants an object at the *end* of the file, so a trailing window would do.
- `llm/api.rs:1141-1145` validates the response and discards the result (`Ok(_) => None`), then `prompt.rs:196-198` recompiles the validator and re-validates the identical string.

`cargo nextest run --workspace`: 7404 passed. fmt and clippy clean.

PR #673, opened 2026-07-28
Add Modal as an LLM provider

brynary:
## Summary

- add an opt-in Modal provider for Kimi K3 using the existing OpenAI-compatible adapter and Kimi agent profile
- authenticate with the Modal-Key and Modal-Secret proxy-token headers while keeping the endpoint base URL in user settings
- add catalog, credential-resolution, request-routing, and optional live coverage
- document server vault setup, direct SDK environment setup, pricing, troubleshooting, and the provider icon

## Validation

- [x] `cargo nextest run -p fabro-model -p fabro-auth -p fabro-llm` (888 passed, 24 skipped)
- [x] `cargo build --workspace`
- [x] `cargo +nightly-2026-04-14 fmt --check --all`
- [x] `cargo +nightly-2026-04-14 clippy --workspace --all-targets -- -D warnings`
- [x] `cd apps/fabro-web && bun run typecheck`
- [x] `cd apps/fabro-web && bun run build`
- [x] `cargo nextest run -p fabro-llm --profile e2e --run-ignored only modal_kimi_k3_deep_tool_round_trip` against the live Modal endpoint (1 passed)

## Notes

`fabro provider login --provider modal` remains out of scope because the current login flow accepts one credential value. Modal setup uses two `fabro secret set` commands instead.


brynary:
Ran a simplification pass over the branch (reuse, quality, and efficiency review). Pushed 8e4251b — tests and fixtures only, no change to shipped behavior. Net −65 lines.

**Removed duplication**
- The Poolside, Fireworks, OpenRouter, and Modal deep round trips were four near-identical ~30-line bodies. Extracted `enabled_provider_catalog` and `assert_deep_tool_round_trip`; each test is now ~6 lines.
- Added `ApiCredential::with_extra_headers` for providers that authenticate with request headers instead of an API key. Replaces two 7-field struct literals that set two fields.

**Fixed dead and duplicated test code**
- The three `require_env` guards in the Modal e2e test were unreachable — `#[e2e_test(live(...))]` already emits a guard per declared var before the body runs. Switched to the `std::env::var(...).expect(...)` form used by every sibling, and registered `MODAL_TOKEN_ID` / `MODAL_TOKEN_SECRET` in `EnvVars` instead of raw strings.
- `modal_requires_both_vault_proxy_tokens` rebuilt the full built-in catalog on each of its two iterations. Collapsed to a single case; the symmetric half duplicated `missing_secret_header_fails_without_echoing_value`.
- Dropped assertions that cannot fail or pin nothing: the exact `api_key_url` doc URL, `default_for_provider`/`probe_for_provider` on a provider with one model marked `default = true`, and the `get_on_provider` loop that re-asserted the key it looked up by. Renamed that test to `builtin_kimi_k3_selection_prefers_direct_kimi_then_modal_over_openrouter` — the two `select` assertions are what it actually proves.
- Inlined the single-use `modal_env_catalog` fixture and noted why it overrides the shipped `{{ secrets.* }}` templates.

**Small things**
- Sorted the `MODAL_*` keys in `.env.example` (they were inserted before `MINIMAX_API_KEY`).
- Recorded in `modal.toml` why `api_id` keeps the Hugging Face capitalization while OpenRouter uses a lowercase slug, and pointed the setup comment at `~/.fabro/settings.toml` to match `openrouter.toml`.

Verified: `cargo nextest run --workspace` (7468 passed), nightly clippy `-D warnings` across the workspace, and `cargo fmt --check`.

**Left alone — flagging for your call, not defects I want to fix silently:**

1. **Cost data.** Review flagged that per-token costs describe the Shared API while a dedicated Auto Endpoint bills by GPU-second, and suggested `billing_policy = "none"`. The docs page already draws exactly this distinction, so this reads as deliberate. Leaving it.
2. **`enabled = true` with no `base_url`.** Modal is the only `openai_compatible` provider shipping without a base URL, so enabling it alone builds a catalog where Kimi K3 is listed and selectable but `Client::from_credentials` fails later in `adapter_registry.rs`. Failing at catalog build with the settings key named would be a better error, but that is a behavior change beyond a cleanup.
3. **`estimated_output_tps = 460`.** Deployment-specific, and Modal would be the only Kimi K3 row in `fabro model list` showing a speed — the Kimi and OpenRouter rows omit it, so Modal will look categorically faster rather than merely measured. Worth a source comment or dropping.
4. **`ApiCredential` derives `Debug` and `extra_headers` prints raw.** `ApiKeyHeader` hand-writes a redacting `Debug`, so the API-key channel is protected and the header channel is not. Pre-existing (Portkey uses extra headers too), but Modal is the first shipped provider whose entire credential lives only in `extra_headers`.
5. **`fabro provider login --provider modal`** builds an `ApiKeyStrategy` with an empty env-var list and stores a key nothing reads, since Modal has no `auth` block. The docs note the command is unsupported; the CLI does not.

PR #676, opened 2026-07-29
Remove nonfunctional run agent permissions setting

brynary:
## Summary

- Remove the nonfunctional run.agent permissions setting from TOML parsing and resolved run settings.
- Remove the field from OpenAPI, generated TypeScript models, and public documentation.
- Keep CLI exec permissions working and move their enum into the CLI settings module.
- Reject the removed run setting as an unknown field.

## Compatibility

This is a deliberate configuration and API contract break. Existing run configuration that declares the removed key must delete it.

## Verification

- cargo nextest run -p fabro-config -p fabro-types -p fabro-api
- cargo nextest run -p fabro-server
- cargo build --workspace
- cargo dev docs check
- bun run typecheck in lib/packages/fabro-api-client


brynary:
Ran a simplification pass over the branch. Pushed ef9606e6e.

**Found a regression this branch introduced.** Two `fabro-cli` inline snapshots still carried `"permissions": null` in the serialized run spec. They pass on `main` and failed here:

- `cmd::attach::attach_json_errors_without_prompting_for_human_input`
- `cmd::inspect::inspect_resolves_selector_via_server_endpoint`

Snapshots accepted. The verification list in the PR description didn't include `-p fabro-cli`, which is how it slipped through.

**Collapsed a duplicate type.** `AgentPermissions` was an exact twin of `fabro_types::PermissionLevel` — same variants, same kebab-case wire form, same crate. `PermissionLevel` is strictly richer (`Hash`, strum derives, `clap::ValueEnum`) and is already the `with_replacement` target for the OpenAPI `PermissionLevel` schema, whose values are byte-identical to the `AgentPermissions` schema this branch deletes. Rather than relocate the duplicate from `settings::run` to `settings::cli`, I deleted it and typed the setting as `PermissionLevel`.

That removes the adapter in `fabro exec`:

```rust
// before
let permissions = cli.exec.agent.permissions.map(|p| match p {
    AgentPermissions::ReadOnly => AgentPermissionLevel::ReadOnly,
    AgentPermissions::ReadWrite => AgentPermissionLevel::ReadWrite,
    AgentPermissions::Full => AgentPermissionLevel::Full,
});

// after
let permissions = cli.exec.agent.permissions;
```

along with the `as AgentPermissionLevel` alias that existed only to disambiguate the two names. TOML wire form unchanged, so no config break.

**Smaller items.** Tightened `rejects_removed_permissions_setting` to assert the exact ``unknown field `permissions` `` message — the `|| contains("unknown field")` branch would have passed if `[run.agent]` itself were later dropped. Renamed its module `run_agent_fabro_tools` → `run_agent`, since it now covers more than `fabro_tools`. Dropped three stale doc references to the removed setting (`agents/mcp.mdx` plus two plan docs this branch already touched).

**Verification.** `cargo nextest run --workspace` → 7622 passed, 0 failed. Clippy `-D warnings` clean, `fmt --check` clean, `cargo dev docs check` up to date, `bun run typecheck` clean in `fabro-api-client` and `fabro-web`. Regenerating the TypeScript client produces no diff against what's checked in.

**Two things I left alone.**

The hard-fail on existing config is worth a conscious call. `RunAgentLayer` has `deny_unknown_fields`, so a `settings.toml` with `permissions = "read-write"` now fails to load entirely — for a key that never did anything, and one that was publicly documented until this branch. `lib/foundation/fabro-config/migrations/` has precedent for stripping a key with a warning instead. The PR description calls the break deliberate, so I didn't add a migration, but nothing in the branch records the decision.

No changelog entry, since those are generated from merged `main` history — worth catching on the next changelog run given the break above.

PR #683, opened 2026-07-29
Support comments in model stylesheets

brynary:
## Summary

- support CSS block comments between model stylesheet tokens
- preserve token boundaries and comment-like text inside quoted strings
- reject unterminated comments and keep non-CSS line comments invalid
- document the syntax and test comments through parsed workflow application

## Test plan

- cargo nextest run -p fabro-graphviz
- cargo nextest run -p fabro-workflow stylesheet
- cargo nextest run -p fabro-validate stylesheet
- cargo +nightly-2026-04-14 clippy -p fabro-graphviz --all-targets -- -D warnings
- git diff --check

## Reference

- https://www.w3.org/TR/css-syntax-3/#consume-comments

### DIFF
--- a/docs/public/workflows/stylesheets.mdx
+++ b/docs/public/workflows/stylesheets.mdx
@@ -114,11 +114,14 @@ selector { property: value; property: value; }
 - Properties and values are separated by `:`
 - Declarations are separated by `;`
 - Whitespace is flexible — newlines and indentation are ignored
-- CSS comments (`/* ... */`) are not supported
+- CSS block comments (`/* ... */`) can appear between tokens and are ignored
+- Comments close at the first `*/` and do not nest; an unterminated comment is a syntax error
+- `//` does not start a CSS comment
 
 ### Full example
 
 ```
+/* Use a fast model by default. */
 *            { model: claude-haiku-4-5;reasoning_effort: low; }
 box          { reasoning_effort: high; }
 tab          { reasoning_effort: low; }
--- a/docs/public/workflows/stylesheets.mdx
+++ b/docs/public/workflows/stylesheets.mdx
@@ -116,7 +116,7 @@ selector { property: value; property: value; }
 - Whitespace is flexible — newlines and indentation are ignored
 - CSS block comments (`/* ... */`) can appear between tokens and are ignored
 - Comments close at the first `*/` and do not nest; an unterminated comment is a syntax error
-- `//` does not start a CSS comment
+- `//` starts a comment in the surrounding workflow file, but not inside a stylesheet
 
 ### Full example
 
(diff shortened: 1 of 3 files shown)


brynary:
Ran a simplification pass over this branch. Net effect on `stylesheet.rs`: **42 insertions, 65 deletions**.

**Simplified the comment stripper.** `strip_css_comments` went from a 56-line byte scanner to a 20-line `str::find` loop. Behavior is unchanged for every valid stylesheet.

**Dropped the quote and backslash tracking.** This was the main find, and it was a bug fix rather than only a cleanup. The stylesheet language has no string literals — `parse_declarations` ends a value at the first `;` or `}` with no quote awareness, and values flow into `AttrValue::String` verbatim, so a quoted model name is just an unknown model that the lint rule flags. Meanwhile the tracking created a failure mode the simple scan does not have: one unpaired apostrophe disabled comment stripping for the rest of the input, then blamed a well-formed comment for the resulting error.

```
* { model: don't; } /* c */
before: Err("expected selector ('*', '#id', '.class', or shape name), got: \"/* c */\"")
after:  Ok(rules: [Universal { model: "don't" }])
```

Because `stylesheet_application.rs:18` discards parse errors (`let Ok(..) = .. else { return Ok(graph) }`), that path silently dropped the whole stylesheet at run time. I removed `comment_markers_inside_strings_are_preserved`, which pinned the old behavior.

**Dropped the `Cow` and its `copied_through` watermark.** They avoided one allocation on a graph attribute of a few hundred bytes, parsed once per workflow load, in a function whose caller already clones the attribute and whose parser allocates a `String` per property and per value.

**Fixed a reachable panic while extracting `excerpt()`.** Adding a third error-snippet site made the duplication worth collapsing. The two existing sites sliced raw bytes at index 20, which panics when a multi-byte character straddles the cutoff, and `model_stylesheet` is arbitrary user text:

```
parse_stylesheet("!aaaaaaaaaaaaaaaaaaéxx")
→ panicked: end byte index 20 is not a char boundary; it is inside 'é'
```

This is pre-existing and slightly outside the PR's scope, but it is two lines in the function being changed and I did not want the new error path to be char-safe while its neighbors were not.

**Smaller items.** `parse_error_line_comment` now asserts the message rather than just `is_err()` — `//` was never valid, so the old assertion also passed on pre-PR code. Added a test that the excerpt cuts on a character boundary. Reworded the `//` docs bullet to explain the surprise: `//` is a comment in the surrounding workflow file but not inside a stylesheet.

**Checked and left alone.** `parser/lexer.rs:16-27` strips DOT block comments by joining the tokens on either side, so `a/*x*/b` lexes as `ab` — a real inconsistency with the new CSS rule, but pre-existing and out of scope. One reviewer flagged the repurposed integration test as a coverage regression; that is a false positive, since `stylesheet_applies_model_override` already covers the comment-free `parse` → `validate` → transform path.

Verified: `cargo nextest run -p fabro-graphviz -p fabro-validate -p fabro-workflow` → 1651 passed, `cargo build --workspace`, `fmt --check`, and nightly `clippy --all-targets -D warnings` all clean.

PR #691, opened 2026-07-30
fix(llm): decode Modal reasoning token usage

brynary:
## Summary

- decode Modal’s top-level `usage.reasoning_tokens` field
- prefer the standard nested `completion_tokens_details.reasoning_tokens` field when both spellings are present
- cover both streaming and non-streaming Chat Completions usage payloads

Modal reports `completion_tokens` as an inclusive total and sends the reasoning subset beside it. Fabro previously discarded that sibling field, leaving all completion tokens in the visible-output bucket and reporting zero reasoning tokens.

## Verification

- `cargo nextest run -p fabro-llm` — 671 passed, 24 skipped
- `git diff --check`


brynary:
Merged `main` (resolved a conflict in `wire.rs` with the DeepSeek `prompt_cache_hit_tokens` work, which landed in the same function) and ran a simplification pass over the diff.

Production change is unchanged — a review of reuse, quality, and efficiency found no issues with it. Notably: no helper extraction for the repeated `.as_ref().and_then(...).or(...)` shape, since each of the three reads a different field, so a helper would need a closure per call site and end up longer. Field name and `.or()` also left as-is, matching the DeepSeek fallback next to it.

Cleanups:

- Collapsed the two envelope tests into focused `ApiUsage` tests. They were one test paid for twice — `ApiResponse::usage` and `StreamChunk::usage` are both `Option<ApiUsage>`, so the envelope cannot change the result. Envelope-level usage decoding is already covered by `stream_chunk_usage_parsing`. The payloads also carried fields `ApiUsage` never deserializes (`total_tokens`, `id`, `model`, `choices`) plus a `cached_tokens` split unrelated to the fix.
- Renamed to the `token_counts_*` convention used by the neighboring tests and regrouped them so the usage tests stay contiguous.
- Added a test pinning the precedence rule this PR introduces: nested detail wins over the flat spelling, and an empty `completion_tokens_details` still falls back to it. That behavior was previously untested. Verified both new tests fail when `.or(self.reasoning_tokens)` is removed.
- Documented the precedence on `token_counts`, which covered disjoint bucketing but not the alternate spellings.
- Reverted the unrelated `cost` doc edit that dropped the concrete OpenRouter reference.

Verification: `cargo nextest run -p fabro-llm` (690 passed, 26 skipped), `fmt --check --all`, `clippy -p fabro-llm --all-targets -- -D warnings`, `git diff --check`.

One out-of-scope finding for a possible follow-up: `codec/openai_compatible/response.rs:81` parses the whole response body a second time into a `serde_json::Value` for `Response.raw`, which appears to have no production reader (only two fabro-llm tests). Not touched here.

PR #696, opened 2026-07-31
fix: expose output schemas to agents

brynary:
## Summary

- Add the resolved output contract to API-backed agent task instructions.
- Limit the contract to the final response so intermediate tool use remains unrestricted.
- Repeat the exact JSON Schema in structured-output repair turns.
- Add regression coverage and document the behavior.

## Why

One-shot prompt nodes pass the schema to the provider as a response format. Multi-turn agent nodes validated the final response but did not show the schema to the model. This could cause an agent to ask the user which object shape to return.

Applying a provider response format to every agent turn would interfere with tool use on some providers and is unsupported by others. The task-level final-output contract exposes the schema without changing intermediate agent turns.

## Testing

- cargo nextest run -p fabro-workflow (1,318 passed; 31 skipped)


brynary:
Ran a simplification pass over the branch.

**Changed**
- `agent_prompt_with_output_schema` and `output_expectation` both dispatched on `OutputSchemaKind`, so they moved onto the type as `agent_prompt()` and `expectation()`. Call site is now `schema.agent_prompt(&prompt)`.
- Split the combined agent test. The five `assert!` calls were inside `CustomOutputBackend::run()`, where a failure surfaces as a panic from `execute()`. Prompt assertions now live in a dedicated test using the `PromptCapturingBackend` pattern already in that file; `codergen_handler_custom_output_schema_updates_output_context_key` goes back to asserting only the outcome.
- Added coverage for the `Routing` branch of the contract, which had none — every `output_schema="routing"` agent node gets this new prompt text too.

**Checked, no change needed**
- `parse_node_output_schema` is called in both `agent.rs` and `api.rs`, recompiling the validator twice per node execution. This is pre-existing on `main`, not introduced here — the PR only moved the `agent.rs` call earlier.
- The no-schema path is already guarded: `codergen_handler_no_preamble_when_empty` asserts exact prompt equality, so an unconditional append would fail it.

**Worth a second look before merge**
- `Event::Prompt.text` now carries the contract, so it shows up in the UI prompt panel, `progress.jsonl`, and the `stages/<id>/prompt.md` checkpoint artifact. That matches what was actually sent, but agent nodes will now display Fabro-authored text that prompt nodes do not.
- Both docs pages say "API-backed agent nodes". The code appends the contract above backend dispatch; that is accurate today only because `acp.rs:523` rejects `output_schema` outright.

`cargo nextest run -p fabro-workflow` — 1321 passed. Clippy and fmt clean.

PR #700, opened 2026-07-31
Fix canceled parallel durations and event persistence logging

brynary:
## Summary

- Show the recorded stage duration when cancellation interrupts a parallel summary.
- Log run-event persistence failures at error level so lost durable events are visible.

## Verification

- bun test app/components/stage-renderers/parallel-children.test.tsx
- bun run typecheck
- cargo nextest run -p fabro-workflow (1,318 passed)

### DIFF
--- a/apps/fabro-web/app/components/stage-renderers/helpers.test.ts
+++ b/apps/fabro-web/app/components/stage-renderers/helpers.test.ts
@@ -195,15 +195,11 @@ describe("parseParallelOverview", () => {
     const overview = parseParallelOverview(events);
     expect(overview).toEqual({
       branchCount: 3,
-      successCount: 2,
-      failureCount: 1,
-      durationMs: 12000,
       results: [
         { id: "branch-a", index: null, itemLabel: null, status: "succeeded" },
         { id: "branch-b", index: null, itemLabel: null, status: "succeeded" },
         { id: "branch-c", index: null, itemLabel: null, status: "failed" },
       ],
-      isComplete: true,
     });
   });
 
@@ -249,7 +245,6 @@ describe("parseParallelOverview", () => {
       }),
     ];
     const overview = parseParallelOverview(events);
-    expect(overview.isComplete).toBe(false);
     expect(overview.branchCount).toBe(4);
     expect(overview.results).toEqual([]);
   });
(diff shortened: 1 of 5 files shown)


brynary:
Ran a simplification pass over this branch (reuse / quality / efficiency).

**The duration tile was a duplicate.** `ParallelChildren` renders `StageMetaBar` five lines above the stat grid, and the meta bar already shows the same stage's duration — with a live ticking clock and a started-at tooltip. The two disagreed while running: the meta bar counted up, the tile showed the static word `running`. The cancelled-stage bug existed only in the duplicate.

So the tile is gone rather than fixed. The meta bar was already correct for cancelled, pending and skipped stages. That removed the three-way duration branch, the `"--"` sentinel decode, and the `ACTIVE_STAGE_STATES` / `formatDurationMs` imports. Stat grid is now Branches / Succeeded / Failed at `sm:grid-cols-3`.

Worth confirming you're happy losing that tile — it's the one product-visible change here.

**Dead code it exposed.** With the tile gone `ParallelOverview.durationMs` is dead, and `successCount` / `failureCount` / `isComplete` were already dead before this branch (the renderer counts the branch rows it actually draws). `ParallelOverview` is now just `branchCount` + `results`.

**The `error!` promotion needed a bound.** That loop has no retry or backoff, so a persistently broken sink fails on every event — a bare `error!` would emit one "investigate me" line per event for the life of the run. Now: first failure at `error!` with `run_id` and event name, subsequent ones at `debug!`, and new losses summarized at flush. Flush runs per stage and per agent turn, so the summary reports only losses since the last one. This follows the `mark_metadata_degraded` precedent already in the repo.

**Left for a follow-up:** `Stage.duration` is `string` with a `"--"` sentinel minted at `stage-sidebar.ts:114` and decoded differently at three call sites (`run-stages.tsx:1661`, `stage-popover.tsx:224`, and previously here). Making it `string | null` would delete all three decodes. Also `EMPTY_VALUE = "—"` is defined privately in both `run-billing.tsx` and `run-sandbox.tsx`. Both are pre-existing and touch files outside this branch, so I left them.

Verification: `bun test` (787 pass; the 13 failures are pre-existing on `main` — chunk-loading tests, confirmed by stashing), `bun run typecheck`, `cargo +nightly-2026-04-14 clippy -p fabro-workflow --all-targets -D warnings`, `cargo nextest run -p fabro-workflow` (1318 passed).

PR #707, opened 2026-07-31
Sanitize Bedrock tool identifiers during encoding

### LINKED ISSUE #578: AWS Bedrock tool name sanitation
## Problem

Fabro’s Bedrock Converse codec preserves tool calls emitted by
the model and sends
them back as conversation history on later turns.

Bedrock requires tool names to match its restricted grammar:

- ASCII letters, numbers, `_`, or `-`
- at most 64 characters

Some models can emit tool-call names outside that grammar (for
example,
`search???` or names containing spaces). Fabro previously
preserved those names
unchanged.

That poisons the conversation history: the initial response may
be decoded, but
the next Bedrock Converse request fails when it includes the
invalid historical
`toolUse.name`. The same issue occurs with streamed tool calls.

## Workaround

I've locally patched fabro in the meantime, but I'd rather this be in the upstream build so that i can just pull the container down instead of building one myself

Sanitize tool names at every Bedrock boundary where a tool call
enters or leaves
Fabro history:

1. Decode non-streaming Bedrock `toolUse` blocks.
2. Decode streaming `contentBlockStart` tool uses.
3. Encode tool-call history for the next Converse request.

The sanitizer:

- retains `[A-Za-z0-9_-]`;
- replaces every other character with `_`;
- truncates to 64 characters;
- uses a non-empty fallback name for an empty input.

For example:

```text
search???  -> search___
bad name   -> bad_name

Sanitizing on decode is important: it prevents an invalid model
response from
ever being persisted into Fabro’s internal message history.
Sanitizing again on
  encode protects existing or externally-created historical
  messages.

swerner:
## Summary

- sanitize historical Bedrock Converse `toolUse.name` values to the provider's `[a-zA-Z0-9_-]+` grammar
- sanitize `toolUseId` values at every `toolUse` and `toolResult` encode site with one deterministic function, preserving call/result pairing
- replace invalid characters, provide non-empty fallbacks, and shorten over-length values to 64 characters with a stable SHA-256 suffix
- preserve already-valid identifiers without allocation

## Why

Bedrock can return a model-emitted tool name that its request validator would reject. Fabro previously stored that output verbatim and replayed it verbatim, so the next request failed with a validation error and the conversation could not continue. Cross-provider histories can hit the same failure with longer names or incompatible tool-call IDs.

Sanitization happens only while encoding the Bedrock wire request. Non-streaming and streaming decoders continue to preserve exactly what the provider emitted, keeping the canonical transcript useful for debugging and replay through providers with different identifier rules.

Tool definitions and named tool choices also remain unchanged. Silently renaming a registered tool would break dispatch without reverse-mapping machinery, so invalid definitions still fail through Bedrock's normal validation.

## Verification

- `cargo nextest run -p fabro-llm` — 681 passed, 24 skipped
- `cargo nextest run --workspace --status-level slow --profile ci` — 7,688 passed, 203 skipped, with ambient provider keys unset for hermetic credential tests

## How this compares to other open-source Bedrock adapters

This failure class ("model output is not replayable as input") is well documented across the ecosystem, and the approach here matches where other projects converged — while covering a couple of gaps they still have open.

**LiteLLM** is the only major adapter that ships a fix, and it arrived at the same shape over two years of incident-driven patches: a single deterministic sanitizer (invalid characters → `_`) applied at every Bedrock encode boundary — tool definitions first ([BerriAI/litellm#5138](https://github.com/BerriAI/litellm/pull/5138)), then message-history `toolUse` names and forced tool choice ([BerriAI/litellm#28874](https://github.com/BerriAI/litellm/pull/28874)). Two things it still doesn't cover that this PR does: the 64-character length limit (open as [BerriAI/litellm#26398](https://github.com/BerriAI/litellm/issues/26398)) and `toolUseId` sanitization with call/result pairing preserved (open as [BerriAI/litellm#28825](https://github.com/BerriAI/litellm/pull/28825)). One thing LiteLLM needs that this PR deliberately avoids: because it also renames tool *definitions*, it has to keep an in-memory reverse-map cache to restore original names in responses, and that cache is lossy (bounded size, TTL, per-process). Fabro sidesteps the whole problem by leaving definitions unsanitized (invalid definitions still fail loudly) and never mutating the stored transcript — there is nothing to restore.

**Vercel AI SDK** has the identical bug on file ([vercel/ai#10202](https://github.com/vercel/ai/issues/10202) — a model hallucinated `$READFILE` and the replayed history failed validation) and an unmerged community fix ([vercel/ai#10258](https://github.com/vercel/ai/pull/10258)) that uses the same replacement rule as this PR (`[^a-zA-Z0-9_-]` → `_`) on the history-replay path, but without length handling or ID coverage. Their provider already normalizes replayed tool-call *IDs* to a model-specific grammar for Mistral (`normalizeToolCallId`), which is the same normalize-on-the-wire pattern this PR applies.

**LangChain** (`ChatBedrockConverse`, Python and JS) passes names and IDs through verbatim with no fix on file; users hitting the constraint are directed to rename their tools at the source (e.g. [awslabs/mcp#3181](https://github.com/awslabs/mcp/issues/3181)).

Two ecosystem-consistent design points worth calling out for review:

- **Nobody sanitizes on decode.** Every adapter that solved this did it at request-encode time and kept internal history faithful to what the model emitted. This PR does the same: sanitization is a per-request rendering of the wire body, so the canonical transcript still replays the original identifiers through providers with different grammars.
- **Deterministic re-sanitization instead of stored mappings.** Like LiteLLM's sanitizer, the functions here are pure and stable across turns and processes, so a re-encoded history always agrees with itself — including `toolUse`/`toolResult` ID pairs, which are sanitized through one shared function. The hash-suffix truncation goes a step further than any of the above to keep distinct over-length identifiers distinct (relevant for MCP-style names sharing long namespace prefixes).

Fixes #578

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2766,6 +2766,7 @@ dependencies = [
  "rand 0.9.4",
  "serde",
  "serde_json",
+ "sha2 0.10.9",
  "strum 0.28.0",
  "thiserror 2.0.18",
  "tokio",
--- a/lib/components/fabro-llm/Cargo.toml
+++ b/lib/components/fabro-llm/Cargo.toml
@@ -21,6 +21,7 @@ anyhow.workspace = true
 thiserror.workspace = true
 serde.workspace = true
 serde_json.workspace = true
+sha2.workspace = true
 strum.workspace = true
 tokio.workspace = true
 uuid.workspace = true
--- a/lib/components/fabro-llm/src/codec/bedrock_converse/decode.rs
+++ b/lib/components/fabro-llm/src/codec/bedrock_converse/decode.rs
@@ -280,6 +280,21 @@ mod tests {
     }
 
     #[test]
+    fn tool_use_names_are_preserved_verbatim() {
+        let block = serde_json::json!({
+            "toolUse": {
+                "toolUseId": "tool-1",
+                "name": "search???",
+                "input": {}
+            }
+        });
+        let Some(ContentPart::ToolCall(tool_call)) = decode_content_block(&block) else {
+            panic!("expected tool call");
+        };
+        assert_eq!(tool_call.name, "search???");
+    }
+
+    #[test]
     fn reasoning_text_block_round_trips_signature() {
         let block = serde_json::json!({
             "reasoningContent": {
--- a/lib/components/fabro-llm/src/codec/bedrock_converse/encode.rs
+++ b/lib/components/fabro-llm/src/codec/bedrock_converse/encode.rs
@@ -4,6 +4,7 @@ use base64::Engine;
 use base64::engine::general_purpose::STANDARD as BASE64;
 use serde_json::{Map, Value, json};
 
+use super::sanitize;
 use crate::codec::{CodecCtx, EncodedRequest, extract_system_prompt, merge_named_provider_options};
 use crate::error::Error;
 use crate::types::{ContentPart, Message, Request, Role, ToolChoice};
@@ -129,7 +130,7 @@ fn encode_message(message: &Message) -> Option<Value> {
             let text = message.text();
             blocks.push(json!({
                 "toolResult": {
-                    "toolUseId": tool_call_id,
+                    "toolUseId": sanitize::tool_use_id(tool_call_id),
                     "content": [{ "text": text }],
                 }
             }));
@@ -185,8 +186,8 @@ fn encode_content_part(part: &ContentPart) -> Option<Value> {
             };
             Some(json!({
                 "toolUse": {
-                    "toolUseId": tool_call.id,
-                    "name": tool_call.name,
+                    "toolUseId": sanitize::tool_use_id(&tool_call.id),
+                    "name": sanitize::tool_name(&tool_call.name),
                     "input": input,
                 }
             }))
@@ -197,7 +198,10 @@ fn encode_content_part(part: &ContentPart) -> Option<Value> {
                 other => json!([{ "json": other }]),
             };
             let mut block = Map::new();
-            block.insert("toolUseId".to_string(), json!(result.tool_call_id));
+            block.insert(
+                "toolUseId".to_string(),
+                json!(sanitize::tool_use_id(&result.tool_call_id)),
+            );
             block.insert("content".to_string(), content);
             if result.is_error {
                 block.insert("status".to_string(), json!("error"));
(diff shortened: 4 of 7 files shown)


swerner:
## Cleanup pass (`/simplify`)

Ran a four-angle quality review (reuse, simplification, efficiency, altitude) over this PR and pushed the resulting cleanups in 594bf6e63. Reuse and efficiency came back clean — notably `sha2` was already a workspace dependency used by 9 other crates, and the valid-input path was already cheap.

### Applied

1. **Dropped the `Cow<'_, str>` machinery in `sanitize.rs`** — every call site fed the result straight into `json!`, which allocates an owned string anyway, so the borrowed fast path bought nothing while costing a three-branch/two-scan implementation and tests that pinned the borrowed-vs-owned distinction. `sanitize()` now returns `String` via a single pass.
2. **Made sanitization structural instead of call-site discipline** — the pairing invariant ("every emitted tool-use ID must go through the helper") was enforced only by a doc comment across three scattered `json!` sites in `encode.rs`. Private `tool_use_block()` / `tool_result_block()` constructors now own the sanitize calls, so any future emit site inherits sanitization by construction.
3. **Removed a can't-fail test assertion** — `encode_with(&request)` takes a shared reference, so the "request not mutated" check was guaranteed by the type system. Renamed the test to `historical_tool_names_are_sanitized_on_the_wire`.
4. **De-duplicated the replacement literals** (`"search___"`, `"bad_id_"`) from the encode-level wiring tests — they now assert through the sanitize helpers, so the charset rule is pinned only in `sanitize.rs`'s own unit tests.

No wire-format changes; clippy, rustfmt, and all 681 fabro-llm tests pass.

### Noted but deliberately not changed (would alter intended behavior)

- **Character-replacement collisions**: distinct IDs like `call 1` and `call_1` both sanitize to `call_1`, which could cross-pair `toolUse`/`toolResult` blocks. The hash-suffix disambiguation only kicks in for overlength values; extending it to *any* modified value would fix this but changes wire output the PR's tests deliberately pin (`search???` → `search___`).
- **Definitions/history asymmetry**: a tool still in `request.tools` with a Bedrock-invalid name (e.g. an MCP tool `weird.name`) gets its historical calls sanitized but its definition sent verbatim — the request still 400s, and the sanitized `toolUse` name matches no `toolSpec`. A request-scoped name mapping (applied to definitions, `tool_choice`, and history, reversed on decode via `CodecCtx`) would cover this, but that's a design decision beyond a cleanup pass.

PR #710, opened 2026-08-01
fix(llm): classify provider error codes through one shared table

brynary:
Follow-up to #701, which fixed Kimi quota classification with a vendor-specific arm. This generalizes that fix and closes the same gap for OpenAI.

## The problem

Three places translated a provider error code into a `ProviderErrorKind`:

| Location | Covers |
|---|---|
| `error.rs` `error_from_status_code` | all non-streaming HTTP, all providers |
| `codec/openai_responses/stream.rs` | OpenAI streaming events |
| `codec/anthropic_messages/stream.rs` | Anthropic streaming events |

The tables disagreed, so the same failure classified differently depending on which path saw it.

The visible case: OpenAI returns HTTP 429 with `error.type = "insufficient_quota"` when an account is out of credit. The streaming decoder mapped that to `QuotaExceeded`. The non-streaming path fell through to the plain `429 => RateLimit` arm, so a spent quota was retried with backoff and never triggered failover. Identical error, two behaviors.

#701 added a fourth spelling of "out of credit" (`exceeded_current_quota_error`) to the status table, which did not know any of the other three.

## The change

Move the code table into `error.rs` as `kind_from_error_code`. It returns `None` when the code says nothing beyond what the caller already knows, so each call site keeps its own default: the stream decoders treat an unrecognized code as transient, `error_from_status_code` falls back to the HTTP status. All three call sites now share it.

Precedence in `error_from_status_code`:

- **Unambiguous statuses** (401, 403, 404, 408, 413, 5xx) win outright. The status is more trustworthy than a body field, and there is no reported bug there.
- **429** defers to the code only when it reports a spent quota. This keeps a junk or unexpected code from turning a real rate limit into something non-retryable.
- **Ambiguous statuses** (400, 422, ...) prefer the structured code over the message-substring guessing, which now runs only when there is no code.

## Behavior changes to review

Beyond the OpenAI 429 fix, merging the tables changes two classifications. Neither was covered by a test; both look like improvements, but they are the parts worth a second opinion:

- `not_found_error` → `NotFound` instead of `Server` for OpenAI. The old suffix rule only matched `*_not_found`, so Anthropic's spelling fell through to the transient default.
- `request_too_large` → `ContextLength` instead of `InvalidRequest`. Anthropic's oversized-input code was hitting the generic `*_too_large` rule.

The third arm of `_ => code_kind.unwrap_or_else(...)` is the widest part of the diff: a 400 whose body says `server_error` now classifies as `Server` (retryable) rather than `InvalidRequest`. That requires a provider returning a 400 with a 5xx-flavored code, which I have not seen, but it is the one case where this trades a bounded retry for a better signal.

## Validation

- `cargo nextest run --workspace` — 7717 passed

New tests: all three quota spellings on a 429, non-quota 429s staying retryable (named, unrecognized, and absent codes), ambiguous statuses preferring the code over the message, and a direct table test that pins the `request_too_large` ordering against the `_too_large` suffix rule.

### DIFF
--- a/lib/components/fabro-llm/src/codec/anthropic_messages/stream.rs
+++ b/lib/components/fabro-llm/src/codec/anthropic_messages/stream.rs
@@ -8,7 +8,7 @@
 use super::SYNTHETIC_TOOL_NAME;
 use super::decode::{convert_synthetic_tool_to_text, map_finish_reason, refusal_error};
 use crate::codec::{RawEvent, StreamDecoder, parse_tool_arguments_or_empty};
-use crate::error::{Error, ProviderErrorDetail, ProviderErrorKind, kind_from_error_code};
+use crate::error::{self, Error, ProviderErrorDetail, ProviderErrorKind};
 use crate::types::{
     ContentPart, FinishReason, Message, RateLimitInfo, Response, Role, StreamEvent, ThinkingData,
     TokenCounts, ToolCall,
@@ -358,7 +358,7 @@ fn stream_error_event_to_provider_error(data: &serde_json::Value, provider_name:
     // overloaded_error, api_error, and unknown stream errors are transient.
     let kind = error_code
         .as_deref()
-        .and_then(kind_from_error_code)
+        .and_then(error::kind_from_error_code)
         .unwrap_or(ProviderErrorKind::Server);
 
     Error::Provider {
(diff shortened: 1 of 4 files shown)


brynary:
Cleanup review complete.

Changes in 97aeb5631:
- use module-qualified calls for the shared error classifier
- keep the classifier crate-private
- remove duplicate message-fallback test coverage
- add an adapter-level regression test for a non-streaming OpenAI HTTP 429 with insufficient_quota

Local validation:
- cargo nextest run -p fabro-llm: 688 passed
- cargo +nightly-2026-04-14 clippy -p fabro-llm --all-targets -- -D warnings

PR #726, opened 2026-08-04
Extract workflow bundling from manifest assembly

swerner:
## Summary

- Move the existing recursive workflow, import, template, config, and Dockerfile traversal behind a private, stateful `WorkflowBundler`.
- Have the bundler emit the existing `ManifestWorkflow` map directly, while `build_run_manifest` remains responsible for settings, goal, arguments, top-level configs, Git context, and the existing best-effort branch push.
- Preserve structured error source chains for graph parsing, template discovery, config parsing, and filesystem failures.

## Compatibility

- Keep the public `RunManifest` API and JSON shape unchanged.
- Preserve current `ManifestPath` behavior, including leading parent components for workflows outside the invocation directory.
- Preserve graph, child-workflow, import, inline-file, template-include, Dockerfile, settings, goal, input-override, reference-provenance, and Git behavior.
- Keep this as an internal pre-factor only: no portable package-path model, storage or API types, source abstraction, producer migration, or new alias/collision policy.

## Tests

Coverage includes:

- The complete legacy manifest projection, including root and child workflows, configs, imports, file references, nested template dependencies, Dockerfiles, goal, arguments, and input overrides.
- External sibling paths, reference provenance, repeated-reference deduplication, structured parser/template/I/O error chains, and direct-bundler Git isolation.
- Existing Git observation and best-effort push behavior.

Validation run:

- `cargo nextest run -p fabro-manifest`
- `cargo nextest run -p fabro-cli validate_reports_missing_template_dependency`
- `cargo +nightly-2026-04-14 clippy -p fabro-manifest --all-targets -- -D warnings`
- `cargo nextest run --workspace` — 7,793 passed, 206 skipped.


swerner:
Ran a cleanup review (reuse / simplification / efficiency / altitude) over this PR and applied the contained fixes in 0e703a777 (net −123 lines). No behavior change intended; fabro-manifest tests (including the legacy-projection characterization test), clippy `-D warnings`, and fmt all pass.

## Applied

**Reuse / duplication**
- Deleted the duplicated lexical path-normalization loop: `normalize_absolute_path` in `lib.rs` was a byte-identical copy of the component walk inside `working_tree.rs`'s `lexically_normalize_access_path`. It now delegates to that loop, and both it and `manifest_path_from_absolute` moved into `working_tree.rs` so the module dependency points one way (projection → collector) instead of both ways. Also removed a redundant re-normalization call in `collect_bundled_file` (a no-op on already-normalized output).
- The goal-file and prompt-file branches in `collect_workflow_files` shared a copy-pasted 15-line "bundle then collect template includes" sequence — extracted into one `collect_bundled_template_includes` method, which also removes the doubled `self.document(bundled)` lookups.
- The two identical config-seeding blocks in `collect_working_tree` became one `seed_config_document` helper; the two identical config-reader closures in `build_run_manifest` became one `read_source_input` function, with the doubled `user_settings_path.is_file()` check evaluated once.

**Simplification**
- Replaced ~100 lines of trivial getters on the five `Collected*` output structs with `pub(super)` fields (the module's own tests were already bypassing the getters). The `CollectedPath` newtype keeps its private field since its constructor enforces a real invariant.
- `assemble_current_manifest` no longer resolves the root workflow twice with two distinct "root workflow missing" errors — the second lookup was provably unreachable.
- `CollectionNamespace`, a struct used only inside its own single associated function, is now a free function `finalize_documents` with two local maps.

**Efficiency**
- `assemble_current_manifest` now consumes `CollectedWorkingTree` by value and moves every source string into the wire types instead of `.to_owned()`-copying the entire bundle contents a second time.
- Removed two full-`DraftDocument` clones (including complete file sources) that existed only to satisfy the borrow checker in `collect_workflow_files` and `collect_config_dockerfile` — they now clone only the `access_path`.
- `verify_recorded_template_dependencies` no longer recomputes `manifest_path_from_absolute` per collected file (O(templates × files)); it round-trips the files-map key through `ManifestPath::from_wire`, which is exactly how those keys were built.

## Noted but not applied

- **Cross-crate reuse with `fabro-types`** — `normalize_relative_path`, `virtual_reference_path`, and `CollectedPath::try_new` partially re-implement `ManifestPath::normalize_components` / `from_reference` / `is_portable_logical_path`. Consolidating means widening fabro-types' public API and would make some validations stricter (a behavior change), so it belongs in its own change.
- **Deeper design follow-ups** — four sound but larger findings that would restructure the representation this PR deliberately shaped: (1) the legacy wire-path derivation is computed independently in both the collector and the assembler; (2) `CollectedFileReference` links documents by physical path rather than collected coordinate, forcing consumers to reverse-map; (3) project-config-dockerfile placement is done via remove/mutate/re-insert map surgery in the collector rather than in the compatibility projection; (4) coordinate-conflict detection lives in three places instead of one choke point in `insert_document`. Worth considering as follow-ups.
- **Test-only duplication** — the git2-based fixture in `collector_does_not_push_an_ahead_branch` duplicates the CLI-based git test helpers in `lib.rs`, and the PR mints two copies of a write-fixture helper across the two test modules. Consolidating means a shared `#[cfg(test)]` support module; left alone.
- **`ComponentRole::label` → strum** — the crate has no strum dependency; adding one for a three-variant `const fn` isn't a win.

swerner:
Review turned up one behavior regression, fixed in 04f45b7c6: the extraction dropped the lexical normalization step for the **root** workflow argument. On main, the root argument went through `collect_workflow_entry`, which lexically resolves `..` segments (and rejects `~`) before reading, while `WorkflowLocation::resolve` alone just joins the argument onto cwd. So on this branch, `fabro run link/../wf/workflow.fabro` (where `link` is a symlink) read the graph through OS symlink resolution — a different file than the lexical manifest key (`wf/workflow.fabro`) names — and bundled the wrong source silently.

Worth knowing for future work in this area: root and child workflows really do take different paths here, and that predates this PR. The *bundled content and manifest key* for both root and children are lexically normalized, but the root's **settings resolution** (`workflow.toml` sibling discovery, project-config discovery, `target_path`) uses the unnormalized `cwd.join(arg)` path, which the OS resolves through symlinks. This commit restores main's exact behavior rather than unifying them; normalizing once at entry would be a small behavior change better done deliberately in a follow-up.

The commit also folds in review cleanups: `bundle()` no longer recomputes the root manifest key, the bundler holds one `FilesystemTemplateStore` instead of rebuilding it per template reference, the dead `Clone` on `WorkflowScanInput` is gone, the characterization test now uses an insta snapshot, both test modules share one fixture writer, and the bundler git-push test was removed (the bundler has no git code path, so it couldn't fail).

PR #731, opened 2026-08-06
Persist pre-start worker failures from runnable runs

swerner:
## Summary

- permit runnable runs to persist the four audited failures that can occur before a worker appends run.starting
- keep run.starting worker-owned and leave scheduler admission and mutex behavior unchanged
- add production-path regressions for launch failure and early worker exit, including durable and API projections

## Motivation

A subprocess worker can fail during launch or exit before it records run.starting. The durable run is still runnable at that point, so the lifecycle validator previously rejected the terminal event and left a stale runnable run behind. This change admits only launch_failed, workflow_error, terminated, and bootstrap_failed from runnable. Other direct failures remain invalid.

## Testing

- cargo nextest run -p fabro-types (437 passed)
- cargo nextest run -p fabro-store --no-fail-fast (246 passed)
- cargo nextest run -p fabro-server --no-fail-fast (805 passed)
- cargo +nightly-2026-04-14 clippy -p fabro-types -p fabro-store -p fabro-server --all-targets -- -D warnings

A full workspace sweep ran all 7,790 tests. It exposed two downstream store fixtures that assumed runnable to workflow_error was invalid; those fixtures are updated here and the full store suite is green. Five untouched fabro-workflow tests remain locally failing in credential and OpenAI mock setup paths, including one test that expects no configured credentials and four vault-backed mock tests that receive a 404.

### DIFF
--- a/lib/apps/fabro-server/src/server/tests.rs
+++ b/lib/apps/fabro-server/src/server/tests.rs
@@ -4,7 +4,7 @@ use std::os::unix::fs::PermissionsExt;
 use std::path::{Path, PathBuf};
 #[cfg(unix)]
 use std::process::Stdio;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::{Arc as StdArc, Mutex as StdMutex};
 
 use async_zip::base::read::mem::ZipFileReader;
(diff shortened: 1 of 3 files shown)


swerner:
Closing without merge. Review surfaced that the relaxed transitions interact badly with stale-writer and cancellation races and with event-log replay on older binaries — we'd rather address those structurally if we revisit the run state machine than patch them piecemeal here.

PR #734, opened 2026-08-06
Add [run.limits] run budgets: halt runs on token/cost spend

brynary:
## What

Optional per-run spend budget. A new `[run.limits]` table accepts `max_cost` (USD) and/or `max_tokens`; when a run's accumulated LLM spend exceeds either limit, the run halts immediately and fails with `FailureReason::BudgetExhausted`. A one-shot warning notice fires at 80% of either limit.

```toml
[run.limits]
max_cost = 25.00
max_tokens = 5000000
```

Motivation: a long-running run recently burned ~$1,300 of OpenRouter usage overnight with nothing to stop it.

## How

- **Tracker** — `RunBudget` in fabro-model: a shared thread-safe accumulator. Every site that records LLM usage charges it; `charge()` returns the exceeded limit and queues 80% warnings for the run-notice stream.
- **Enforcement is inline, no background watchdog** — spend only becomes known when a response reports usage, so the checks live at the usage-recording sites: agent turn loop (interrupts exactly like the wall-clock timeout, new `InterruptReason::BudgetExhausted`), subagent sessions, one-shot prompt path including schema-repair retries, context compaction, `web_fetch` summarization, and prompt/agent hooks.
- **Halt semantics** — budget exhaustion is an engine-level error (`fabro_core::Error::BudgetExhausted`, following the `StallTimeout` precedent), so it terminates the run instead of converting to a fail outcome and routing along fail edges. A between-stage gate stops non-LLM stages, and a run-start gate makes a resumed budget-failed run fail fast (spend seeds from the billing projection) until the limit is raised.
- **Config & API** — settings layer + resolver with validation; `limits` exposed on the dense `RunNamespace` OpenAPI schema (`RunLimitsSettings`); Rust and TS clients regenerated; docs updated (`run-configuration.mdx`, generated options reference).

Deliberately out of scope: CLI flags / per-run overrides (budgets come from the settings layers only), parent budgets spanning child runs, ACP usage capture, and billing-side recording of compaction/hook spend (see the review-notes comment).

## Testing

- `cargo nextest run --workspace`: 7,813 passed
- `cargo +nightly clippy --workspace --all-targets -- -D warnings`: clean
- New coverage: tracker unit tests; config merge/resolve/validation; api-level tests for one_shot trip, pre-flight skip, notice-emitted-once, agent fail-fast, and a mid-session halt (tripping turn's follow-up request is never sent); an engine-path integration test proving fail edges are not followed; hook pre-flight skip tests; OpenAPI parity tests; two snapshot updates (`"limits": {}` in resolved settings JSON).


brynary:
## Review notes: coverage audit and things worth scrutinizing

An exhaustive audit of LLM call sites in the run path was done as part of this change (two independent passes). Summary for the reviewer:

### Coverage map

| Site | Charged by budget | In billing today |
|---|---|---|
| Agent turn loop (`fabro-agent/src/session.rs`) | ✅ per response | ✅ |
| Subagent sessions (via session factory in `handler/llm/api.rs`) | ✅ | ⚠️ live projection only — see "known issues" |
| One-shot prompt path incl. schema-repair retries (`one_shot`) | ✅ | ✅ |
| Context compaction (`fabro-agent/src/compaction.rs`) | ✅ new in this PR | ❌ pre-existing gap |
| `web_fetch` summarization (`fabro-agent/src/tools.rs`) | ✅ new in this PR | ❌ pre-existing gap |
| Prompt/agent hooks (`fabro-hooks/src/executor.rs`) | ✅ new in this PR | ❌ pre-existing gap |
| ACP external agents (`handler/llm/acp.rs`) | ❌ | ❌ (`usage: None`) |
| Post-run PR title/body + server run-title generation | ❌ (runs after execution) | ❌ |

Verified clean (no LLM calls): router, conditional, human, parallel, command handlers, fabro-interview, agent memory/skills/question/todo tools. `manager_loop` reuses the charged registry.

### Design points a reviewer should check

1. **No double counting.** The projection reducer adds `AgentMessage` usage incrementally but *replaces* stage totals on `StageCompleted`. The budget deliberately does not mirror billing: it charges once per response at the source (session/one_shot/hooks/compaction), and stage-terminal events never charge. The only rollup-based charge is the one-time resume seed in `build_run_budget` (`pipeline/initialize.rs`).
2. **Termination path.** Non-retryable handler errors normally become fail outcomes so workflows can route fail edges. Budget exhaustion must not — it maps to `fabro_core::Error::BudgetExhausted` in `node_handler.rs` and terminates the executor loop (same shape as `Cancelled`/`StallTimeout`). `budget_exhausted_ends_the_run_without_following_fail_edges` in the workflow integration tests pins this.
3. **No failover on budget errors.** `classify_agent_error` maps `Interrupted(BudgetExhausted)` to Terminal, never `FailoverEligible` — exhausting the budget on model A must not restart the work on model B.
4. **Overshoot is accepted by design.** Checks run when a response's usage arrives, so the run can exceed the budget by the in-flight requests (one per concurrent branch). Agreed in planning; documented.
5. **Hooks fail open on exhaustion.** A spent budget makes prompt/agent hooks skip with `Proceed` — consistent with their existing posture on timeouts and client errors. If a blocking policy hook must never be skipped, that posture deserves discussion.
6. **Spend-equal-to-limit does not trip** (strictly greater comparison); `max_tokens` compares against `total_tokens()` (all five buckets).

### Known issues surfaced by the audit (follow-ups, not fixed here)

- **Subagent billing inversion (pre-existing, now more visible):** subagents are charged by the budget, but `SubAgentResult` carries no usage, so `StageCompleted.billing` overwrites the live stage total with the parent-only figure. The budget can trip at a total the billing UI cannot explain. Fix belongs in billing (fold subagent usage into the parent's accumulator *without* re-charging).
- **ACP is chargeable in principle:** the workspace enables `unstable_session_usage` and Codex sends `SessionUpdate::UsageUpdate` (with cost), but `fabro-acp/src/session.rs` ignores it. Caveat for the future fix: those values are cumulative per session — charge deltas, or spend gets multiplied.
- **Compaction/hook/web-fetch spend is budgeted but still invisible to billing** — no usage event channel exists for them. Budget ≥ billing is the conservative direction, but resume re-seeds from billing, so a resumed run's seeded spend can be slightly lower than what the failed run actually consumed.
- **Untested composition:** `build_run_budget`'s projection seeding has no store-level integration test — its pieces (rollup, charge-on-seed, start gate, tracker seeding semantics) are each unit-tested, but not the composition end-to-end.

PR #741, opened 2026-08-10
Share one blob store across run handles

swerner:
## Summary

- make Database construct and share one BlobStore with every run handle
- route run-scoped blob reads, writes, and listing through that shared store
- remove the duplicate raw-key blob listing path

## Why

Blob bytes already live in a global content-addressed namespace, but each RunDatabase constructed its own BlobStore facade and listing bypassed the facade entirely. Consolidating access behind one Database-scoped owner and implementation boundary makes the later SlateDB-to-SQLite cutover smaller.

## Behavior preserved

- blob keys, bytes, IDs, and public APIs are unchanged
- deleting a run still retains global CAS blobs
- malformed blob IDs are still skipped during listing
- read-only run handles can read and list blobs but cannot write them

## Testing

- cargo nextest run -p fabro-store
- cargo +nightly-2026-04-14 clippy -p fabro-store --all-targets -- -D warnings

### DIFF
--- a/lib/components/fabro-store/src/keys.rs
+++ b/lib/components/fabro-store/src/keys.rs
@@ -1,7 +1,7 @@
 use std::fmt::{self, Write};
 use std::ops::Range;
 
-use fabro_types::{RunBlobId, RunId, SessionId};
+use fabro_types::{RunId, SessionId};
 
 pub(crate) const MAX_EVENT_SEQ: u32 = 999_999;
 
@@ -91,10 +91,6 @@ pub(crate) fn run_events_range(run_id: &RunId, start_seq: u32) -> Range<SlateKey
     run_event_seq_prefix(run_id, start_seq)..end
 }
 
-pub(crate) fn blobs_prefix() -> SlateKey {
-    SlateKey::new("blobs").with("sha256").into_prefix()
-}
-
 pub(crate) fn sessions_by_id_prefix() -> SlateKey {
     SlateKey::new("sessions").with("by-id").into_prefix()
 }
@@ -113,21 +109,6 @@ pub(crate) fn parse_event_seq(key: &str) -> Option<u32> {
         return None;
     }
     segments.next()?.split_once('-')?.0.parse().ok()
-}
-
-pub(crate) fn parse_blob_id(key: &str) -> Option<RunBlobId> {
-    let mut segments = SlateKey::segments(key);
-    if segments.next()? != "blobs" {
-        return None;
-    }
-    if segments.next()? != "sha256" {
-        return None;
-    }
-    let id = segments.next()?;
-    if segments.next().is_some() {
-        return None;
-    }
-    id.parse().ok()
 }
 
 #[cfg(test)]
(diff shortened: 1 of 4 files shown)


swerner:
Ran a `/simplify` pass (4 review agents: reuse, simplification, efficiency, altitude) and applied the surviving findings in f773a24 (net −45 lines; 248 fabro-store tests pass, fmt + clippy clean).

## Applied

1. **Collapsed the constructor plumbing** (`run_store.rs`, `slate/mod.rs`). `open_writer`/`open_reader` were 8-line wrappers that only hardcoded the `read_only` flag, and each of the three `Database` call sites repeated the same dependency litany (`self.blobs()`, two `Arc::clone`s). Replaced by one `pub(crate) RunDatabase::build` plus a private `Database::open_run_database(run_id, read_only)` helper that gathers the shared stores in one place — the next Database-owned store threads through one function instead of four.
2. **Removed a wasted fetch from `open_run`'s hot path.** It fetched the blob store before the active-run cache check, so the common cache-hit case paid for an `Arc` it immediately discarded (`open_run_reader` already had the ordering right). The helper above is now called only on the cache-miss path.
3. **Deduplicated the raw-db test fixture** in `blob_store.rs` — `list_skips_malformed_blob_ids` copied the `slatedb::Db::open` + `BlobStore::new` bootstrap verbatim from `raw_db_reads_exact_blob_bytes`; both now share a `raw_store(name)` helper.
4. **Made `open_run_reader_is_read_only` test what its name says.** `create_run` leaves the run in the active cache, so the test's reader was a `read_only_clone` of the writer and the reader-construction blob plumbing was never exercised. The test now evicts the cached writer first, so `read_blob`/`list_blobs`/`write_blob(ReadOnly)` all run against a genuinely constructed reader.

## Noted but not applied

- **Move the `Error::KeyParse` skip into `Repository::scan_ids_stream`**: the other caller, `RunCatalogIndex::list`, currently propagates parse errors — unifying the semantics is a real design decision that changes behavior outside this diff. `Repository::parse_entry_id` also conflates "foreign key under my prefix" with "my key, corrupt id", so `BlobStore::list` silently drops genuinely malformed blob keys; worth a follow-up.
- **Drop `ids.sort()` in `BlobStore::list`**: it is a no-op today (lexicographic key order equals `RunBlobId`'s derived byte order for fixed-width lowercase hex), but removing it would make correctness depend on a subtle cross-module key-encoding invariant to save an O(n) pass over an already-sorted vec.
- **Allocation-heavy `parse_entry_id`/`scan_entries` on the new list path**: the new path allocates ~3 heap objects per scanned key where the deleted raw scan allocated none, but the waste lives in pre-existing `record/repository.rs` code, and `list_blobs` currently has no production callers.
- **Bigger refactors out of scope**: bundling the Database-owned handles into a shared context struct (which would also make the blob-store sharing unconstructible-otherwise instead of asserted via `Arc::ptr_eq` on private fields), and the naming seam where run-scoped `RunDatabase::list_blobs` lists the global CAS namespace (pre-existing).

PR #742, opened 2026-08-10
Add immutable workflow version resource

swerner:
## Summary

> **Stacked on #744** (walker unification). This PR now contains only the new resource; the reference-walker refactor and its three semantic alignments were split into #744, which must land first. Once #744 merges, this PR's base flips back to main. The pre-split combined branch is archived as `codex/workflow-version-resource-combined`.

- Add the immutable WorkflowVersion domain resource with exactly entrypoint, files, and workflow_dependencies, plus strict WorkflowPath validation and deterministic canonical raw JSON (serialized once at construction).
- Validate graph imports, templates, file references, workflow.toml source rules and graph selection, Dockerfile paths, and exact child-workflow dependency bindings without reading the host filesystem.
- Keep WorkflowVersionId as the API and domain name while using BlobHash as its private, lossless content-addressed value; IDs parse case-insensitively and serialize canonical lowercase.
- Validate the complete stored workflow dependency closure before writing a root or returning a stored version, deduplicate repeated dependency reads, cap direct workflow dependencies at 512, and detect file/directory path collisions via exhaustive ancestor-prefix checks.
- Add the authenticated create-only POST /api/v1/workflow-versions endpoint, OpenAPI contract, Rust type replacements with JSON-parity tests, and generated TypeScript client.

The implementation enforces 512 files, 512 workflow dependencies, 512 KiB per file, 240-byte and 16-component paths, and a 2 MiB canonical artifact and raw request limit. Dependencies must resolve through the shared blob store to readable, valid, canonical workflow versions before the root is written.

The API returns invalid_json for malformed JSON, workflow_version_too_large for oversized bodies, workflow_version_invalid for local validation failures, and workflow_version_dependency_not_found for absent, invalid, or non-canonical dependencies (422). Persistence failures are logged with their internal chains and return a curated 500 response.

## Scope

This intentionally does not migrate producers, consume workflow versions from run intents, change POST /runs, add project-settings admission, add read/list/delete/catalog/GC surfaces, or introduce speculative workflow wrapper types. Workflow packages store canonical raw source rather than parsed workflow IR. Known deferred follow-ups: dependency-closure validation is O(closure size) per request pending a validated-marker index, and the transport body limit equals the canonical size limit (no headroom for non-canonical JSON formatting).

## Verification

- Tree is byte-identical to the reviewed combined branch (`git diff` against `codex/workflow-version-resource-combined` is empty)
- cargo nextest run -p fabro-types -p fabro-workflow-version -p fabro-api -p fabro-server: 1486 passed
- cargo +nightly-2026-04-14 fmt and clippy clean on touched crates
- High-effort code review completed; all confirmed findings fixed in the stacked commits


swerner:
## Simplification pass

Three follow-up commits from a quality review (reuse / simplification / efficiency / altitude) of this PR.

### 970e72f — Simplify workflow version validation and supporting types

- **Validate once, serialize on demand.** `canonical_bytes()` used to re-run the full structural validation (graph parse, template-closure discovery, O(n²) collision scan) on every call, so each create request validated twice and each store `get` twice more (plus once per dependency on `put`). The type is immutable and only constructible via `new()`/`Deserialize`, so validation now happens once at construction.
- **One collision loop.** The three copy-pasted pairwise loops (files×files, deps×deps, files×deps) collapsed into a single loop over the chained key sets.
- **Allocation-free `WorkflowPath::is_ancestor_of`** (string prefix + `/`-boundary check instead of two `Vec`s per comparison — it runs ~131k times per validation at the 512-file cap), and `validate` no longer collects components into a `Vec`.
- **Dead code removed:** `WorkflowPath::resolve_from_root`, the `missing_dependencies`/`unused_dependencies` error accessors, and `WorkflowVersionStoreError::is_dependency_unavailable` (folded into the handler's match).
- **Convention alignment:** `ReferenceKind` `Display` via strum; `WorkflowVersionId` uses the `#[serde(into/try_from = "String")]` newtype pattern and delegates length/charset checks to `RunBlobId` (keeping only the lowercase guard); the `fabro-workflow` `static_reference` re-export shim is gone; the handler's JSON-rejection match collapsed from five arms to two.

### 0b8273a — Split workflow version wire type from semantic validation

The PR's original shape made `Deserialize` run full semantic validation, which coupled the wire type to the parser stack and broke crate layering in three places:

- `fabro-api` (foundation) depended on `fabro-workflow-version`, pulling graphviz-sys (native C) and minijinja into every API consumer (`fabro-client`, `fabro-tool`, `fabro-mcp-server`, CLI).
- `fabro-store` gained its first domain-validation dependency; `fabro-checkpoint` transitively built the graphviz binding.
- `fabro-graphviz` gained all of `fabro-template` for a three-line `contains_template_syntax` probe.

Now: the `WorkflowVersion` wire type + structural invariants (size limits, entrypoint presence, unique keys, path collisions, canonical form) live in `fabro-types`; the wire shape is unchanged. Semantic validation stays in `fabro-workflow-version` behind a `ValidatedWorkflowVersion` newtype — the only door to a semantically-valid version — and `WorkflowVersionStore` moved there too (`put` takes validated versions; `get` still re-validates blobs from shared storage). The static-reference vocabulary moved to `fabro_types::graph`, template-syntax validation to `fabro-template`. `cargo tree` confirms `fabro-api`, `fabro-store`, `fabro-checkpoint`, and `fabro-graphviz` are all parser/template-free again.

### 1fc36f8 — Unify the workflow graph reference walkers

The manifest bundler and workflow-version validation each maintained a parallel walk over the graph reference vocabulary (goal files, inline goal/prompt templates, imports, child workflows, `@`-file attributes) — a new reference-bearing attribute had to be added twice or the two would drift. `fabro_template::visit_graph_references` is now the single per-graph walker (validates references are template-free, emits typed events); both consumers keep their own IO/resolution/recursion. Likewise `SettingsLayer::image_layers[_mut]` is now the one definition of where images live in a settings layer, replacing three hand-rolled dockerfile traversals (run compiler, manifest bundler, workflow-version validation).

Behavior note: the manifest bundler now also follows `stack.child_dotfile` (already in the shared vocabulary and treated as a child-workflow reference by workflow-version validation). Nothing in the engine emits or reads that attribute today.

### Considered and not done

- **Concurrent dependency verification in `put`** (`try_join_all`): dependency counts are typically 0–2; not worth the complexity yet.
- **`spawn_blocking` for CPU-heavy validation in the handler**: a maximal payload costs real CPU on a runtime thread, but after de-duplicating validation it's one pass; revisit if payloads grow.
- **Borrowed `BundleTemplateStore`**: validation still clones file contents once per pass because the store takes owned `String`s; a borrowed variant would need a `fabro-template` API change.
- **`WorkflowPath` vs `ManifestPath` shared validation core**: the two portable-path validators overlap and are bridged by an `expect`; delegating would lose per-check error messages, so left as is.
- **`Database::workflow_versions()` accessor**: the handler assembles the store from `blobs()` directly; fine until a second endpoint needs it.
- **Dropping `stack.child_dotfile` from the vocabulary**: possibly vestigial (only the classifier knows it); left for a separate discussion since it's a behavior question, not a cleanup.

All verified with workspace build, clippy `-D warnings`, `fmt --check`, and the full test suite (the only failures are pre-existing, environment-dependent `fabro-workflow` credential tests that also fail on the original head). The intermediate commit was built and tested in isolation for bisectability.

swerner:
Follow-up on the last open item above: `stack.child_dotfile` turned out to be a zombie alias, not just vestigial vocabulary. It was the original attractor-spec name for the manager-loop child workflow path, kept as a backward-compat alias after the rename to `stack.child_workflow`, deliberately removed in #281 — and then accidentally resurrected the same day when #290 introduced `reference_kind_for_attribute` with the alias baked in. Since the engine never reads it, validation treating it as live was actively misleading (a workflow version using it would demand a dependency pin for a child that never spawns). Dropped in 8188e1a.

PR #748, opened 2026-08-14
Close workflow goals over version dependencies

swerner:
## Summary

- require workflow-config goal files and their static template dependencies to live inside immutable workflow versions while keeping goal contents template-capable
- validate config and graph template roots in one aggregate traversal, including every authored root occurrence when logical paths collide
- expose an opaque, validated root-plus-dependencies closure from the workflow-version store and make ordinary reads project the root from that path
- preserve canonical bytes, content IDs, storage layout, existing typed errors, and the current HTTP schema while making uploads and stored-version reads stricter

## Scope

This does not add RunIntent admission, runtime lowering, compiler seams, project acquisition, producer cutover, manifest compatibility changes, or OpenAPI changes.

## Verification

- workspace check and pinned-nightly formatting
- workspace Clippy with warnings denied
- 7,851 workspace tests passed; 206 expected skips
- docs refresh and docs check
- 805 web tests plus web and API-client typechecks on Bun 1.3.14
- optimized fabro-cli release build

### DIFF
--- a/lib/apps/fabro-server/src/server/handler/workflow_versions.rs
+++ b/lib/apps/fabro-server/src/server/handler/workflow_versions.rs
@@ -116,7 +116,7 @@ mod tests {
     use axum::body::{Body, to_bytes};
     use axum::http::{Method, Request, StatusCode, header};
     use axum::response::IntoResponse;
-    use fabro_types::WorkflowVersionId;
+    use fabro_types::{BlobHash, WorkflowVersion, WorkflowVersionId};
     use serde_json::{Value, json};
     use tower::ServiceExt;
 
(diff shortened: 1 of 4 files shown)


swerner:
## /simplify review (ecbf07fed)

Four parallel review passes (reuse, simplification, efficiency, altitude) over this PR's diff, with the surviving findings applied. Net **-5 lines**.

### Applied

- **Wasted closure accumulation** (`store.rs`): `put()` and `get()` were building a `BTreeMap` of every transitive dependency's full contents just to discard it — peak memory went from O(one version) to O(entire closure) per request. The walk is now a single `walk_dependency_closure(deps, visit)` with a visitor sink: `put`/`get` pass a no-op, only `get_closure` collects.
- **Root held structurally in `LoadedWorkflowVersionClosure`**: the root now lives in its own field instead of inside the map guarded by two `expect()`s and a doc-comment promise. `root()` and the (now-deleted) `into_root()` panic sites are gone; `get()`/`versions()` compose root + dependencies.
- **Dead `visited: HashSet` parameter removed**: it existed only to pre-seed the root id, guarding against a dependency cycle that content-addressing makes impossible (it would require a hash fixpoint).
- **Cross-crate error dissection moved to the owner**: `template_discovery_path()` exhaustively matched `TemplateDiscoveryError`/`TemplateLoadError` internals with an `expect` and an `unreachable!` borrowed from another crate's invariants. `TemplateDiscoveryError` now exposes `source_name()` in fabro-template, and the consumer shrank to three lines. Future variants get handled where they're defined.
- **5× repeated `TemplateSource::new(manifest_path(...), template_root.clone(), ...)` collapsed** into a small `TemplateRoots` collector that also absorbs the two threaded parameters (`&ManifestPath` + `&mut Vec<TemplateSource>`) into one, and removes the root-path `expect` from `ValidatedWorkflowVersion::new`.
- **Goal-file validation now shares `validate_dockerfile`'s pipeline** via a new `validate_config_file_reference()` (validate static ref → resolve → require file), instead of a second hand-rolled copy of the same sequence.
- **Test cleanups**: `version_with_config` takes `impl Into<String>`, `rejects_missing_workflow_goal_file` uses the `version_with_goal_file` helper added in this PR, and the new `version_id()` helper in store tests is now used at the four pre-existing hand-rolled sites too.

### Noted but not changed

- **Root/path collision gap in `discover_static_dependency_closure`**: a loaded dependency sharing a path with a root is never queued, so its transitive includes silently skip validation, and "last root wins" makes `closure.sources` ambiguous under collision. Fixing this changes validation behavior (could newly reject versions), so it's left for a correctness-focused follow-up.
- **`ReferenceKind`'s growing dead arms in the graph walker** (`static_reference.rs`): `RunGoalFile` follows the existing `Dockerfile` precedent exactly; splitting classifier-output from error-label roles is a larger refactor than this PR warrants. Three dead arms is a trend worth watching.
- **A `WorkflowVersion::id()` production method in fabro-types** would remove the last hand-rolled id derivation (one site in the server tests), but adding public API for test convenience didn't seem worth it.

### Verification

`cargo build`, `cargo nextest run` (883/883 passed across fabro-workflow-version, fabro-template, fabro-server), nightly clippy `-D warnings`, and `fmt --check` all clean.

PR #749, opened 2026-08-14
Persist workflow version lineage on runs

swerner:
## Summary

- Adds optional durable WorkflowVersionId lineage to RunSpec and the canonical run.created event/replay path.
- Threads the exact version identity through server compilation, persistence, retry, fork, and rewind.
- Adds only an optional RunSpec response field to OpenAPI and the generated TypeScript client.

## Compatibility

- Historical JSON defaults the missing field to None.
- None is omitted during serialization, preserving existing event and run-spec shapes.
- Existing manifest and general create paths continue to persist no workflow-version identity.

## Deliberately out of scope

This does not add RunIntent, target checkout, producer migration, lifecycle changes, or accepted-definition cleanup.

## Rebased onto the shared RunSpec test fixture (#762)

Now based on `main` including #762, which added `fabro_types::test_support::test_run_spec()`. Test literals that #762 converted to `..test_run_spec()` no longer need the new field spelled out — the fixture carries it, so five files dropped out of this diff entirely (`fabro-dump`, `serializable_projection`, `billing_rollup`, `run_projection`, `run_spec_methods`).

The reduction is smaller than expected: **50 files → 45**, and 7 of 46 `RunSpec`-literal field additions absorbed. #762 deliberately converted 13 of 64 literals, leaving production paths and deliberately-exhaustive round-trip tests explicit, so most of this PR's remaining churn is in literals it didn't touch. 52 of the added field lines are in `Event::RunCreated` literals, which a separate prefactor would address — see the follow-up notes in the comments.

Conflict resolution kept #762's spread form at every site and added the new field to the fixture once, so `test_run_spec()` covers it for every current and future adopter.

## Verification

- `cargo build --workspace` (without `test-support`) — clean
- `cargo nextest run -p fabro-types -p fabro-api -p fabro-store -p fabro-dump -p fabro-cli` — 1889 passed
- `cargo nextest run -p fabro-workflow -p fabro-server` — 2146 passed, 5 failed
- `bun run generate` / `bun run typecheck` in `lib/packages/fabro-api-client`

The 5 failures (`initialize_prepares_sandbox_and_uses_persisted_run_dir` and four `pull_request` vault/title tests) reproduce identically on unmodified `main` — they assert credentials do not leak into resolved LLM sources and are sensitive to ambient `.env` values in a local shell. Unrelated to this change.


swerner:
## `/simplify` pass

Reviewed for reuse, simplification, efficiency, and altitude (not correctness). Applied in 3ff7a445d — net **−116 lines**, with `cargo build --workspace`, `cargo +nightly fmt --check --all`, `cargo clippy --workspace --all-targets -D warnings`, and the touched crates' tests all clean.

### Fixed

**Redundant serde-omission coverage.** The `skip_serializing_if` behavior for this one field was asserted five times across three crates. Kept the two assertions in `fabro-types`, which owns the attribute, and dropped the rest:

- Deleted `convert.rs::run_created_omits_absent_workflow_version_id` — a 38-line copy of the test above it, re-testing another crate's serde attribute. `convert.rs`'s actual responsibility (copying the field) is covered by the one-line assert already added at `convert.rs:2688`.
- Deleted `create.rs::legacy_create_input_persists_without_workflow_version_id` — a 45-line async test that built a tempdir, in-memory store, and catalog and ran the full `create()` pipeline to prove that `create.rs:103`'s hardcoded `workflow_version_id: None` is `None`. `CreateRunInput` has no such field, so no input could change the result; the test could only fail if someone edited that literal.
- Folded `run_spec_serde.rs::run_spec_omits_absent_workflow_version_id` (a fresh 14-field `RunSpec` literal) into the adjacent `run_spec_defaults_automation_for_legacy_specs`, which already had an all-`None` record in hand.
- Dropped the two off-topic `spec_json` re-serialization lines from `run_state.rs`'s `retried_from` test, keeping the on-topic `assert_eq!`.

**Shared fixture for the magic seed.** `WorkflowVersionId::from(BlobHash::new(b"workflow"))` was copy-pasted into 8 sites across 5 crates, and two assertion sites recomputed `BlobHash::new(b"workflow").to_string()` as a separate expression that could drift from the constructed value. Added `test_support::test_workflow_version_id()` next to the existing `test_run_provenance()` and adopted it everywhere — including `retry.rs`'s private `fn workflow_version_id()`, which was the same helper one level too low. No `Cargo.toml` changes needed: `fabro-types` with `features = ["test-support"]` is already a dev-dep in every affected crate.

**Reverted a no-op parameterization.** `run_projection_round_trip.rs::run_spec_json` gained an `Option<WorkflowVersionId>` parameter, but `RunProjection` is a `with_replacement` alias for the canonical type, so the `Some` and `None` call sites exercise byte-identical code and can't detect spec/Rust drift on this field. Restored the parameterless helper with the field populated.

**Made two spec mirrors honest.** `pipeline/pull_request.rs:1123` and `server/tests.rs:7113` read `run_spec` for every sibling field but hardcoded `workflow_version_id: None`. No bug today (both specs are `None`), but they stop mirroring reality the moment a producer populates the field. They now read `run_spec.workflow_version_id`.

### Skipped

**The field is write-only in production.** Both producers (`handler/runs.rs:733`, `create.rs:103`) pass `None`, and nothing reads it outside tests — so the retry/fork/projection propagation is currently unobservable. The PR body scopes out producer migration explicitly, so this is disclosed staging, not an oversight. Left intact.

**`CreateRunPersistenceInput::workflow_version_id()`** was flagged as a public getter with only a test caller, but it mirrors the equally test-only `workflow_slug()` accessor and enables the one test (`run_compiler.rs:1020`) that actually proves the field survives normalize → prepare → compile → pin → assemble. Kept.

**Exhaustive destructure in `fork.rs::persist_forked_run`** — `retry.rs` destructures `RunSpec` exhaustively so a new field is a hard compile error there, while `fork.rs` reads `spec.field`. Not worth changing: the `Event::RunCreated` struct literal is itself exhaustive, so the compiler still catches an added field. No silent-drop risk.

### Follow-ups worth their own PR

Three pre-existing findings, flagged because they're the reason a one-field change touched 49 files and 68 `: None` insertions:

1. **`Event::RunCreated` is a 17-field duplicate of `RunCreatedProps`.** The only difference is that `Event` stores `settings`/`graph` as `serde_json::Value`, which `convert.rs:47-50` immediately deserializes back to the typed form behind an `.expect()` panic — and every producer serializes typed→`Value` on the way in. It's the only variant of 86 doing this round trip. Collapsing to `Event::RunCreated { run_id, props: RunCreatedProps }` would have made this PR ~10 files instead of 49, and deletes two panic paths.
2. **No `RunSpec → Event::RunCreated` constructor.** `create.rs:566`, `fork.rs:152`, and `retry.rs:70` each rebuild the event field-by-field from a spec, differing only in title / provenance / `retried_from` / `parent_id` / `web_url` / `manifest_blob` and in whether the JSON gets normalized. One `Event::run_created(spec, overrides)` collapses all three plus most of the 39 test literals.
3. **`EventBody` is 2096 bytes** because `RunCreatedProps` is unboxed and the next-largest variant is `StageCompletedProps` at 640. The server's `broadcast::channel(4096)` of `EventEnvelope` is a preallocated ring, so that's ~8.6 MB resident for payloads typically under 100 bytes, plus per-clone cost for every SSE subscriber. `Box<RunCreatedProps>` (one alloc per run — there's exactly one `run.created` each) cuts the ring to ~2.7 MB, at the cost of match ergonomics at every use site.

Also noted, out of scope: `RunSpec` itself isn't in `build.rs`'s `with_replacement` list (only `RunProjection` is), so a parallel generated `fabro_api::types::RunSpec` exists with `run_id: String` and nothing tests yaml↔Rust parity — which is why this PR had to hand-edit both `fabro-api.yaml` and `run-spec.ts`. `WorkflowVersionId` itself is correctly replaced (`build.rs:727`).

### Verification note

5 tests fail in `fabro-workflow` (`initialize_prepares_sandbox_and_uses_persisted_run_dir` and 4 `pull_request` vault/title tests), but all 5 fail identically on the pre-change commit — they assert credentials don't leak into resolved LLM sources and are sensitive to ambient `.env` values in a local shell. Unrelated to this diff.

swerner:
Once #762 lands, this PR should be rebased onto it. #762 adds a shared `test_run_spec()` fixture in `fabro-types::test_support` and converts the highest-traffic `RunSpec { … }` test literals to `..test_run_spec()` spreads, which removes a large share of the mechanical test-literal churn in this PR — the converted sites no longer need an edit just to add `workflow_version_id`. https://github.com/fabro-sh/fabro/pull/762

swerner:
## Corrections to my follow-up notes above

I measured the three "follow-ups worth their own PR" items properly afterwards. Four numbers in that section were wrong — the recommendations stand, but don't act on these figures:

**Item 1 — "would have made this PR ~10 files instead of 49."** It's **~33**. I attributed the whole diff to the event half and ignored that the `RunSpec` half is nearly as large. Attributing every added field line to its enclosing struct: 52 land in `Event::RunCreated`/`RunCreatedProps` literals, **46 land in `RunSpec` literals**, 8 are the real field declarations, 32 are plumbing and assertions. By file: 16 touched only by an event literal, 13 only by a `RunSpec` literal, 5 by both. Collapsing the event variant fixes the first group only. Getting to ~15 files needs a shared `RunSpec` test fixture as well — that's #762, filed separately.

Also, "the only variant of 86 doing this round trip" undersold it. Classifying all 92 variants against their payload structs (normalizing module paths, without which twelve variants falsely look like type changes): 24 are field-for-field identical, 42 differ only by routing fields that get hoisted onto the envelope, and 14 do real work — 8 collapse many variants into one payload with a synthesized discriminator, 3 enrich, 3 fan out into 71 `EventBody` variants. So 66 of 92 are purely mechanical, and `RunCreated` is the only one of three genuine type changes that is a serialize-then-parse round trip rather than a real conversion.

**Item 3 — three wrong figures.**

- The next-largest payload is **`AgentMessageProps` at 752**, not `StageCompletedProps` at 640. (`StageCompleted` is third; clippy's own diagnostic reports 752.)
- The rings hold `EventEnvelope`, which is **2512** bytes, not `EventBody` at 2096 — so the global SSE ring reserves **~10.3 MB**, not ~8.6 MB, and a 10k-event listing runs ~25 MB.
- Boxing takes `EventEnvelope` to **1168** and the ring to **~4.8 MB**, not ~2.7 MB. Those smaller numbers require boxing `AgentMessage` as well, which is a separate decision.

**Item 3 — "at the cost of match ergonomics at every use site" is wrong.** `Box` deref means every match and `let`-else site keeps compiling untouched. There are 8 `EventBody::RunCreated` occurrences across 5 files and only **one** needs editing — the construction at `convert.rs:46`. Verified: the variant change plus `Box::new(…)` gives a clean `cargo check --workspace --all-targets` with zero warnings. Two bonuses found while checking: the `#[allow(clippy::large_enum_variant)]` on `EventBody` can be deleted outright afterwards (the largest/second-largest gap falls to 112 bytes, under clippy's 200 default), and wire-neutrality is verified empirically, not just argued — the same `run.created` JSON re-serialized through `RunEvent::to_value` is byte-identical boxed vs unboxed.

**One thing I understated:** I said the event refactor was "mostly neutral" for the SlateDB→SQLite events cutover. That's true of the `RunCreated` variant alone and wrong as a general claim. Every indexed column in the planned `run_events` table except `run_id`/`seq`/`event_name` is a routing field (`node_id`, `stage_id`, `session_id`, with `run_events_by_stage` and `run_events_by_session` built on them), so the column set is already determined by `StoredEventFields` — while today those values are derived by 340 lines of per-variant matching in `stored_fields.rs`. The routing half of the refactor is genuinely load-bearing for that cutover.

None of this changes anything about the code in this PR.

PR #750, opened 2026-08-14
Check out exact admitted commits in clone-based sandboxes

swerner:
## Summary

- add an optional exact-commit capability to Docker and Daytona sandbox requests while leaving every existing production caller on branch-only behavior
- require a non-empty repository branch whenever an exact 40-hex commit SHA is supplied
- keep Docker's efficient exact-target path: initialize an empty repository, shallow-fetch the admitted SHA, detach HEAD, and verify the result
- use Daytona's official SDK clone path with both `branch` and `commit_id`, preserving its existing retry, credential, layout, and event behavior

## Target semantics

This is execution-layer capability for the later RunIntent admission cutover. The intended public Git target requires a repository and branch, with an optional SHA:

- without a SHA, admission resolves the named branch's HEAD once
- with a SHA, admission verifies that the commit is reachable from the named branch
- execution receives the validated branch plus resolved SHA and never falls back to a newer branch HEAD

A normal branch advance does not invalidate an older admitted commit. A destructive branch rewrite or commit deletion may make an admitted SHA unavailable by setup time; that run fails rather than changing what it executes.

Docker and Daytona intentionally keep different transport implementations. Daytona's documented `commit_id` behavior clones the selected branch and then checks out the commit, which matches the branch-reachability contract. It retains Daytona's existing full-clone and SDK timeout characteristics. Docker can continue using its native shallow exact-fetch path.

This PR does not introduce or wire the RunIntent API, persistence, producer migrations, or event changes.

## Verification

- `cargo +nightly-2026-04-14 clippy -p fabro-sandbox --features docker,daytona --all-targets -- -D warnings`
- `cargo nextest run -p fabro-sandbox --features docker,daytona` (238 passed, 10 skipped)


swerner:
Ran a simplification pass over this PR (reuse / simplification / efficiency / altitude) and pushed the resulting cleanups as 65616b4 (−230 net lines). Test suite (236 tests, including the hermetic git proof), clippy `-D warnings`, and `fmt --check` are all green.

## Applied

**Daytona no longer downloads the repository twice.** The exact-checkout path used to SDK-clone the default branch, then discard that work with a shallow fetch of the admitted SHA. It now mirrors the Docker approach — `git init` + shallow fetch + detach through the process service — extracted into a `checkout_exact_commit` method. This also removed the `daytona_clone_branch` branch-suppression helper, the special-case `Err(e) if commit_sha.is_some()` clone-error arm, and the six-times-repeated `report_clone_failure` + `fail_init` pattern (now one `map_err` at the call site). The symlink/completion/push-credentials epilogue is shared by both paths instead of nested inside the clone match.

**Docker's ~175-line `checkout_exact_github_commit` is gone.** It duplicated the entire `clone_github_repo` skeleton (auth resolution, events, prepare, retry closure, symlink, bookkeeping). `clone_github_repo` now takes `commit_sha` and branches only on the acquisition step; a new `retry_git_transfer` helper holds the previously copy-pasted retry closure, and `exact_checkout_exec_error` merged into a label-parameterized `clone_failure_error`. The two drifting auth-error styles were unified on the structured form.

**Redundant validation removed.** The spec-level `decide_clone` pre-checks in `sandbox_spec.rs` duplicated the identical guard the constructors run before connecting to Docker/Daytona (which this PR's own tests assert). Both blocks are deleted and the `anyhow::Context` cfg widening reverted; the spec test now asserts the constructor's error context.

**Smaller items:**
- Checkout + `rev-parse HEAD` are now one shell command (`exact_checkout_verify_command`), saving an exec round trip per init — the hermetic git test confirms stdout is exactly the SHA.
- The `git -c maintenance.auto=0 -c gc.auto=0` prefix now comes from the shared `sandbox::GIT` constant in the three command builders that hardcoded it.
- `DockerCloneFailure` / `DaytonaExactCheckoutFailure` merged into one `clone_retry::CloneAttemptFailure`.
- The pass-through `daytona_exact_exec_error` helper was inlined.

## Considered and skipped

- **Daytona SDK's `GitCloneOptions.commit_id`** as an alternative: a plain clone + checkout can't reach commits a force-pushed branch left behind, which is exactly the case this PR guards against; init + fetch-by-SHA is the robust choice.
- **A `CloneSource` struct / `CloneTarget` enum** to replace the loose `Option<String>` triple in the spec enums and constructors: the right long-term move, but it reshapes the `SandboxSpec` API across fabro-server and fabro-workflow — better as a follow-up.
- **The constructor-level `decide_clone` guard stays** alongside the one in `initialize()`: it's what makes invalid SHAs fail before any daemon/client connection, and tests pin that behavior.

PR #753, opened 2026-08-17
Add bounded GitHub repository reads

swerner:
## Summary

- Add a repository-scoped `GitHubRepositoryReader` that resolves refs to canonical 40-hex commit SHAs and streams caller-capped UTF-8 files at an explicitly supplied exact SHA.
- Resolve App authentication once per reader while preserving direct PAT and installation-token behavior.
- Require GitHub raw-file media for successful content reads, rejecting directory and object metadata before consuming the body.
- Migrate the live branch-head lookup to the raw commits endpoint while preserving its public signature, retry behavior, and 404-to-`None` compatibility.
- Extend Twin GitHub with exact ref/file state, commit and content endpoints, stored branch SHAs, and inspectable token counts.

## Safety and provider semantics

- Repository paths, ref selectors, API bases, and exact SHAs are validated before their corresponding requests.
- Successful SHA and file bodies are bounded on the wire; invalid UTF-8 errors retain only `std::str::Utf8Error`, not repository bytes.
- New errors omit bearer tokens, raw request URLs, upstream response bodies, ref/path inputs, and file contents while preserving safe source chains.
- GitHub can use 404 to hide inaccessible private resources, so not-found means not observable with the current credentials.
- Rate-limit detection is intentionally header-based. A body-only secondary-limit 403 remains permission-denied because error bodies are never consumed.

## Scope

This establishes the network boundary for later repository-backed `RunIntent` admission. Admission orchestration, materialization policy, aggregate budgets, overall deadlines, retries and caching, clone/tree APIs, persistence, OpenAPI, and generated clients remain out of scope.

## Verification

- `cargo +nightly-2026-04-14 clippy -p fabro-github -p twin-github -p fabro-test -p fabro-workflow --all-targets -- -D warnings`
- `cargo nextest run -p twin-github`
- `cargo nextest run -p fabro-github`
- `FABRO_TEST_MODE=twin cargo nextest run -p fabro-github --test integration --run-ignored only`
- `cargo nextest run -p fabro-workflow -- stale_remote_branch_is_rejected_before_pull_request_creation`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2665,11 +2665,11 @@ dependencies = [
  "jsonwebtoken",
  "serde",
  "serde_json",
+ "strum 0.28.0",
  "thiserror 2.0.18",
  "tokio",
  "tracing",
  "tracing-subscriber",
- "url",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2669,6 +2669,7 @@ dependencies = [
  "tokio",
  "tracing",
  "tracing-subscriber",
+ "url",
 ]
 
 [[package]]
--- a/lib/apps/fabro-server/src/server/tests.rs
+++ b/lib/apps/fabro-server/src/server/tests.rs
@@ -7028,6 +7028,8 @@ async fn pr_test_app_with_completed_run(
     (state, app, run_id)
 }
 
+const COMPLETED_RUN_FINAL_SHA: &str = "0123456789abcdef0123456789abcdef01234567";
+
 async fn create_run_with_pull_request_record(
     state: &Arc<AppState>,
     run_id: RunId,
@@ -7134,7 +7136,7 @@ async fn create_completed_run_ready_for_pull_request(
             status:               "succeeded".to_string(),
             reason:               SuccessReason::Completed,
             total_usd_micros:     None,
-            final_git_commit_sha: Some("final-sha".to_string()),
+            final_git_commit_sha: Some(COMPLETED_RUN_FINAL_SHA.to_string()),
             final_patch:          Some(final_patch.to_string()),
             diff_summary:         None,
             billing:              None,
@@ -9726,11 +9728,12 @@ async fn create_run_pull_request_creates_and_persists_record() {
     let github = MockServer::start();
     let branch_mock = github.mock(|when, then| {
         when.method("GET")
-            .path("/repos/acme/widgets/branches/fabro/run/42")
+            .path("/repos/acme/widgets/commits/heads%2Ffabro%2Frun%2F42")
+            .header("accept", "application/vnd.github.sha")
             .header("authorization", "Bearer ghu_test");
         then.status(200)
-            .header("content-type", "application/json")
-            .body(json!({ "commit": { "sha": "final-sha" } }).to_string());
+            .header("content-type", "application/vnd.github.sha")
+            .body(COMPLETED_RUN_FINAL_SHA);
     });
     let create_mock = github.mock(|when, then| {
         when.method("POST")
@@ -9930,11 +9933,12 @@ async fn create_run_pull_request_persists_generation_failure() {
     let github = MockServer::start();
     let branch_mock = github.mock(|when, then| {
         when.method("GET")
-            .path("/repos/acme/widgets/branches/fabro/run/42")
+            .path("/repos/acme/widgets/commits/heads%2Ffabro%2Frun%2F42")
+            .header("accept", "application/vnd.github.sha")
             .header("authorization", "Bearer ghu_test");
         then.status(200)
-            .header("content-type", "application/json")
-            .body(json!({ "commit": { "sha": "final-sha" } }).to_string());
+            .header("content-type", "application/vnd.github.sha")
+            .body(COMPLETED_RUN_FINAL_SHA);
     });
     let find_mock = github.mock(|when, then| {
         when.method("GET")
--- a/lib/components/fabro-github/Cargo.toml
+++ b/lib/components/fabro-github/Cargo.toml
@@ -26,7 +26,7 @@ tracing.workspace = true
 tokio = { workspace = true }
 base64.workspace = true
 thiserror.workspace = true
-url.workspace = true
+strum.workspace = true
 
 [dev-dependencies]
 fabro-macros = { path = "../../foundation/fabro-macros" }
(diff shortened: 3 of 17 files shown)


swerner:
## `/simplify` pass — quality cleanup

Two commits pushed. No correctness bugs were hunted here; this is reuse, simplification, efficiency, and altitude only.

### `8dfda7e` — Reduce duplication in GitHub repository reads

**Twin GitHub server**

`handlers/commits.rs` and `handlers/contents.rs` shipped byte-identical copies of six items — `AuthorizationError`, `authorize`, `authorization_error`, `accepts`, `is_exact_commit_sha`, `message` (56 lines, verified with `diff`) — and two of those response mappers already existed a third time in `pulls.rs`. All of it now lives in `handlers/support.rs`. `AuthorizationError` itself is gone: it re-encoded the existing `BearerTokenError` + `InstallationTokenAccessError` into a flat enum that a second match immediately mapped back to a `Response`, and nothing ever inspected the intermediate value.

Also: `AppState::find_repository` / `repository_mut` replace four open-coded repository lookups; `head_refs` / `heads_selector` single-source the `heads/{branch}` mapping that was spelled out in three places; and `Repository.files` is keyed by SHA then path instead of a `(String, String)` tuple, which drops two allocations and two full-map scans per content request.

**Repository reader**

- Switched to `fabro_redact::DisplaySafeUrl`, which retires the file-scope `#![expect(clippy::disallowed_types)]` and the new direct `url` dependency. That suppression covered all 725 lines and everything later added to the module, and the invariant in its reason string ("never logged or included in errors") was enforced by nothing — a later `tracing::debug!(url = %url)` would have passed clippy silently.
- The reader stores `repository_base` (`{api_base}/repos/{owner}/{repo}`) built once at `open()`. The URL builders became infallible methods, deleting three copies of a `cannot be a base` error path that `parse_api_base` had already made unreachable.
- Collapsed `RevisionNotFound`/`ContentNotFound` and `RevisionUnavailable`/`ContentUnavailable` into `NotFound { operation }` / `Unavailable { operation, status }` — the same distinction `UpstreamUnavailable` and `UnexpectedStatus` already modeled as a field. `Operation` is public with `strum::Display` per CLAUDE.md's enum guidance, replacing the hand-written variant→string map, and the enum is now `#[non_exhaustive]`.
- Smaller: `classify_status` is one `Err(match ...)` rather than ten `Err(...)` wrappers; `parse_api_base` returns just the URL instead of `(Url, String)` where the `String` was derivable from the `Url`; `collect_bounded` sizes its buffer from `Content-Length`; `parse_resolved_commit_sha` lowercases in place instead of rebuilding the string byte by byte.

**The one non-cosmetic fix**

`verify_remote_head` retries the branch-head read up to `BRANCH_HEAD_ATTEMPTS` times, and each attempt called `branch_head_sha`, which opened a full `GitHubRepositoryReader` session. With App credentials that re-minted an installation token every time — two extra HTTP round trips per retry, so three token mints to read one branch head. It now opens one reader before the loop. This is the property the type was built for; the PR's own test is named `repository_reader_reuses_one_app_token_for_ref_and_file_reads`, and this was the one production path violating it.

Note the deliberate behavior change: only the ref lookup is retried now. A credential failure surfaces immediately rather than being retried three times with 500 ms sleeps, which matches the documented rationale for the retries (remote ref visibility lag), not auth.

### `3ab5846` — Drop the branch-head compatibility wrapper

With the hoist above, `branch_head_sha` had no production callers left — only an integration test named `branch_head_compatibility_wrapper_uses_repository_reader`, i.e. a test of the shim itself. Preserving its signature was migration safety, and that migration is complete.

What remained only made the typed API worse: it joined an already-validated owner and repo into a slug so `open` could split them apart again, invented an "invalid repository coordinate" error for a value `parse_github_owner_repo` had already validated, opened a fresh credential session per call, and flattened `RepositoryReadError` into `anyhow` while keeping exactly one variant — leaving callers unable to distinguish "rate limited, back off" from "401, stop now". The concrete risk in keeping it: the next person needing a branch head finds the more discoverable of two paths and opens a second credential session without noticing.

Its test now asserts the same 404-means-not-observable semantics through `resolve_commit`, where that contract actually lives.

### Deliberately not done

- **Rebuilding the reader on the crate's `HttpClient` trait.** Real altitude issue: the reader forks a second HTTP/auth/URL stack because the shared trait ends in `resp.text().await` and cannot express a bounded read — which is why four `MockHttpClient` unit tests for `branch_head_sha` were deleted with no unit-level replacement. Fixing it properly means adding a bounded method to the trait and implementing it for both the real client and the mock. Worth doing, but well outside this diff.
- **Moving the `fabro-server` / `fabro-workflow` httpmock stubs onto the twin.** Those suites now hand-encode `commits/heads%2F...`, the `application/vnd.github.sha` media type, and a raw-body response — GitHub transport details leaking into two consuming crates. The twin models this generically now; collecting that payoff is a separate change.
- **Dropping the content-read path.** `read_utf8_file_at` has no production caller yet, but the PR scopes itself to establishing the network boundary for later admission work. Intended, not accidental.
- **Reusing `fabro-types`' `workflow_path` validation** for `validate_repository_path`. The rules overlap closely but the limits differ materially (4096/256 vs 240/16); a general repo-file reader shouldn't inherit workflow-path constraints.
- **A `RepositoryReadLimits` config struct.** `max_bytes` being a bare per-call `usize` will become shotgun surgery once there are real call sites, but with none yet this is premature.

### Verification

`cargo +nightly-2026-04-14 fmt --check --all` and `clippy -p fabro-github -p twin-github -p fabro-workflow --all-targets -- -D warnings` both clean. `cargo nextest run -p twin-github -p fabro-github -p fabro-workflow -p fabro-server` gives 2256/2261 passing, plus 13/13 on `FABRO_TEST_MODE=twin ... --test integration --run-ignored only`.

The 5 failures are pre-existing and environment-dependent — they fail identically with these changes stashed, on ambient LLM credentials leaking into tests that assert credentials are empty (`initialize_prepares_sandbox_and_uses_persisted_run_dir`, `build_pr_content_uses_vault_only_openai_codex_source`, `open_pull_request_caps_fallback_title_at_72_chars`, `open_pull_request_falls_back_to_goal_title_when_llm_returns_empty_title`, `workflow_run_with_vault_only_openai_codex_builds_pr_body`). Four of the five don't touch any changed code.

Net across both commits: **−227 lines**.

swerner:
Closing this as we've decided to go in a different, simpler direction

PR #758, opened 2026-08-19
Plane ticket factory

hendrixmar:


brynary:
Hi @hendrixmar -- We're going to close this out because it doesn't have a PR description and we are assuming it may have been an accidental PR launched by an AI agent. But please feel free to let us know if you would like to discuss any potential enhancements to Fabro!

hendrixmar:
It was definetly and accident. Sorry for that
that i currently working to have native support for ACP for various coding agents. Currently is inly scoped for oh my pi and codex. Im very interested in this project. I will like to contribute.

brynary:
No worries! There is some [ACP support](https://docs.fabro.sh/core-concepts/agents#acp-backend) in Fabro today -- are you working on extending that in some way?

hendrixmar:
is not in this PR but this is what i have been working since yesterday

1. Exact token billing for ACP stages (the headline)

Per-turn token usage from ACP PromptResponse.usage is now preserved end-to-end — the Rust ACP helper previously reduced responses to StopReason and threw usage away, which is why ACP stages billed as zero
Semantics kept honest: usage_update.used (context occupancy) is explicitly not treated as consumption, and USD cost renders only when the harness actually reports it — missing cost is never displayed as $0
Usage projects into the existing stage/billing UI; totals saturate correctly; web presentation is test-covered
2. Canonical tool inventory + catalogs

ACP tool calls are now grouped by stable tool identity instead of display title (previously Read file X and Read file Y showed as different tools in the dropdown)
OMP's official 23-built-in catalog is seeded so tools appear before first use; observed tools get marked invoked, unknown/plugin tools appended; Codex stays observed-only since its surface varies by preset
Structured tool kinds preserved through the pipeline; observed tools surface in the stage UI — reusing the existing agent.tools.available projection rather than adding a parallel ACP-specific UI
3. Per-phase harness configuration on nodes

New node attributes harness, reasoning_effort, and skills; stylesheet classes gain harness + skills properties — so an ACP node can now declare which external harness runs it and with what effort, where previously ACP nodes couldn't carry model controls at all
Validator rules added: harness_valid (codex|omp on ACP nodes), reasoning_effort_enum (caught at validate time instead of silently killing the node mid-run — the failure mode we hit ourselves), skills_known
Env translation to harness-native knobs: FABRO_ACP_MODEL/REASONING_EFFORT/HARNESS, CODEX_CONFIG JSON merge (node attrs override baked config), PI_MODEL for OMP
4. Skills materialization into harnesses

Engine library or repo-local skills are materialized into harness-native directories (~/.codex/skills, ~/.omp/agent/skills) before spawn, with an agent.skills.materialized event; API nodes get skill_dirs + skill_allowlist pinning
5. Session reliability fixes

Prevents live-session starvation and drains buffered response updates — the class of bugs that made long ACP turns (like our 15-minute GLM dead-code phase) fragile
In-band ACP errors (a "type":"error" output) now fail the turn instead of passing silently
6. Normalized agent events from ACP sessions — tool_call updates become standard agent.tool.started/completed events plus one skill:// activation per skill per stage visit, so ACP phases look native in the event log; RunManifest.external_agent_harness flows into run metadata and worker env

PR #770, opened 2026-08-21
Rescue release pushes when origin/main moves mid-release

brynary:
## Problem

`cargo dev release` checks out main, runs the release smoke (~15 minutes), then pushes a version-bump commit to `main` plus the release tag in a single unguarded `git push`. Any commit landing on main during that window makes the push fail as non-fast-forward and kills the whole release — this is what broke the manually dispatched `v0.332.0-nightly.1` run (GH run 32473204303).

Reading the code also surfaced a latent worse failure: the push was not atomic. If main were rejected while the tag ref was accepted (possible whenever the racing commit doesn't touch workflow files), the Release workflow would ship from an orphan tag not reachable from main, and main would never receive the version bump.

## Changes

- **Atomic push**: `git push --atomic origin main <tag>` — both refs land or neither does.
- **Bounded rescue loop** (4 attempts): when the push is rejected, drop the bump commit and tag this run created, fast-forward onto the updated origin/main, recompute the version against freshly fetched tags (handles a concurrent release claiming the tag name), rebuild the bump commit on the new tip, and push again. Each retry takes seconds, shrinking the race window from ~15 minutes to a couple of seconds per attempt.
- **Divergence guard**: the rescue resets only to the pre-bump HEAD and uses `git merge --ff-only origin/main`, so unpushed local commits fail loudly with a clear error instead of being reset away.
- Extracted a `ReleaseVersions` struct so the compute-bump-tag cycle can be repeated per attempt; dry-run output reuses the real push command (now shows `--atomic`).

The retried tag can include commits this run's smoke did not test. That is acceptable: those commits passed CI to land on main, and the Release workflow re-runs the full workspace test suite on the exact tagged commit across all targets before publishing anything.

## Tests

New unit tests in `release.rs` drive `commit_tag_and_push` against fixture repos with a bare origin and a concurrent writer:

- push is rejected because origin/main moved → rescue rebuilds the bump on the new tip and succeeds
- concurrent release claims the tag name → rescue recomputes to the next nightly number
- local main has unpushed commits and origin moved → fails with a divergence error and preserves the local commits

`cargo nextest run -p fabro-dev --features dev`: 37 passed (the `policy::*` and `docs::*` failures are pre-existing on unmodified main in this environment). Clippy and fmt clean.

### DIFF
--- a/lib/foundation/fabro-dev/src/commands/release.rs
+++ b/lib/foundation/fabro-dev/src/commands/release.rs
@@ -8,6 +8,7 @@ use super::{PlannedCommand, capture_command, run_command, spa_refresh, workspace
 
 const RELEASE_EPOCH: &str = "2026-01-01";
 const RELEASE_TEST_SEGMENT_WRITE_KEY: &str = "fake-for-local-smoke";
+const MAX_PUSH_ATTEMPTS: u32 = 4;
 
 #[derive(Debug, Args)]
 pub(crate) struct ReleaseArgs {
@@ -36,6 +37,12 @@ struct ReleasePlan {
     root:         PathBuf,
 }
 
+struct ReleaseVersions {
+    current: String,
+    next:    String,
+    tag:     String,
+}
+
 #[expect(
     clippy::print_stdout,
     reason = "dev release command reports progress and dry-run commands directly"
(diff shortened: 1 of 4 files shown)


lib/foundation/fabro-dev/src/commands/release.rs
```diff
@@ -156,6 +118,156 @@ impl ReleasePlan {
...
+
+    /// Commits the version bump, tags it, and pushes `main` plus the tag
+    /// atomically. When the push is rejected because origin/main moved while
+    /// the release ran, rebuilds the bump commit and tag on the fresh tip
+    /// and retries.
+    #[expect(
+        clippy::print_stdout,
+        reason = "dev release command reports push retry progress directly"
+    )]
+    fn commit_tag_and_push(
```
brynary, line 183: Review note: a rescued push tags a tree that this run's release smoke did not test — the smoke ran before the rescue fast-forwarded onto the commits that landed on origin/main mid-release.

I think this is the right tradeoff: those commits passed CI to land on main, and the Release workflow re-runs the full workspace suite on the exact tagged commit across all five targets before anything publishes, so an actually-broken tree fails there without shipping. The alternative — re-running the smoke on each retry — reopens the ~15-minute race window every attempt, which defeats the purpose of the rescue.

Flagging it since it slightly weakens what the pre-push smoke guarantees. If we ever want the strict property back, the retry loop is the place to add a re-smoke.

lib/foundation/fabro-dev/src/commands/release.rs
```diff
@@ -156,6 +118,156 @@ impl ReleasePlan {
...
+        &self,
+        cargo_toml: &Path,
+        mut versions: ReleaseVersions,
+    ) -> Result<String> {
+        let mut attempt = 1;
+        loop {
+            let start_head = self.head_commit()?;
+            self.create_bump_commit_and_tag(cargo_toml, &versions)?;
+            let Err(error) = self.push_main_and_tag(&versions.tag) else {
```
brynary: Review note: the rescue triggers on *any* push failure, not just non-fast-forward rejections. A permanent failure (bad credentials, unreachable remote) will burn all 4 attempts before surfacing, at a cost of a few seconds per attempt.

I kept it this way deliberately: classifying rejections would mean parsing git stderr, which is brittle across git versions, and the rescue is harmless when the failure isn't a race — it resyncs, recomputes the same version, and fails again with the real error. If we'd rather fail fast on non-race errors, `git push --porcelain` output is the sturdier thing to parse, but I didn't think the added complexity paid for itself.

PR #775, opened 2026-08-21
Additional GitHub repository access

brynary:
Implements `.ai/plans/2026-07-22-additional-github-repositories.md` (revised 2026-08-21).

A run can now declare additional GitHub repositories that its stages may access through the one managed `GITHUB_TOKEN`:

```toml
[run.integrations.github]
additional_repositories = ["fabro-sh/keystone"]
permissions = { contents = "read" }
```

The run origin stays implicit. Fabro mints **one** installation token scoped to the origin plus every declared repository with the shared permission map, and stages reach the declared set through `gh`, raw GitHub API calls, plain Git over HTTPS, and the common SSH URL spellings (bridged to authenticated HTTPS with no secret in Git config).

## Commits

1. **feat(config)** — validated `additional_repositories` field: `GitHubRepositorySlug` gains `FromStr`/`Display`/string serde and case-insensitive identity; resolution validates slug grammar, duplicates, single owner, the 499 cap, and the required `contents` permission with indexed error paths. Wholesale list replacement across layers; `[]` clears.
2. **feat(github)** — `GitHubRepositoryAccess` (validated effective set, primary-first, secret-free), per-repository installation resolution requiring one shared installation ID, multi-repository mint, and `InstallationTokenSource::for_access` with unchanged caching/refresh/single-flight.
3. **feat(workflow)** — one `ResolvedGithubIntegration` value carried from CLI/server workers through `SandboxEnvSpec`; hard initialization errors for declared sets (missing creds/origin, inaccessible repository) with eager installation + token validation; legacy permissions-only stays best-effort. New `git_bridge` module injects secret-free `GIT_CONFIG_*` entries (env-reading credential helper, per-repo SSH→HTTPS rewrites, `GIT_TERMINAL_PROMPT=0`), appended after any valid user Git-config overlay. Contract tests drive the real `git` binary against local fixtures.
4. **feat(server)** — preflight resolves each repository's installation, mints once, probes every effective repository with `git ls-remote … HEAD` (token only in the child env; same-token retry for replication lag), and reports one check per repository in deterministic primary-first order. `GH_TOKEN` in the resolved run environment warns without failing. Primary-only behavior unchanged.
5. **docs(api)** — OpenAPI schema + canonical-type reuse via `with_replacement` with type-identity witnesses, regenerated TS client, integration/run-configuration docs (including the corrected `GITHUB_TOKEN` claim and the `contents = "write"` blast-radius warning), changelog entry, and an opt-in live GitHub App e2e test.

## Scope notes

- All repositories must share one owner and one App installation; cross-owner/cross-installation access is an explicit follow-up.
- Managed cloning of additional repositories is out of scope — workflows add their own clone step.
- PAT mode: the list drives validation and probes but cannot narrow the PAT's inherent scope (documented).

## Verification

- `cargo nextest run --workspace` — 8017 passed
- `cargo +nightly-2026-04-14 fmt --check --all` / `clippy --workspace --all-targets -- -D warnings` — clean
- TS client regenerated; `bun run typecheck` clean
- Live e2e (`fabro-github/tests/live_access.rs`) and the post-deploy acceptance run from the plan still need real App credentials.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2663,6 +2663,7 @@ dependencies = [
  "fabro-static",
  "fabro-test",
  "fabro-types",
+ "futures",
  "jsonwebtoken",
  "serde",
  "serde_json",
--- a/docs/public/api-reference/fabro-api.yaml
+++ b/docs/public/api-reference/fabro-api.yaml
@@ -14447,6 +14447,17 @@ components:
           type: object
           additionalProperties:
             type: string
+        additional_repositories:
+          type: array
+          description: |
+            Additional GitHub repositories, beyond the implicit run origin,
+            that the minted GITHUB_TOKEN must cover. Each entry is a full
+            `owner/repository` slug; every repository must share one owner
+            with the run origin. Omitted when empty; settings persisted
+            before this field existed deserialize to an empty set.
+          items:
+            type: string
+          uniqueItems: true
 
     RunGoal:
       oneOf:
--- a/docs/public/changelog/2026-08-21.mdx
+++ b/docs/public/changelog/2026-08-21.mdx
@@ -1,0 +1,20 @@
+---
+title: "Additional GitHub repositories"
+date: "2026-08-21"
+---
+
+## One token for the whole repository set
+
+A run can now declare additional GitHub repositories that its stages may access through the managed `GITHUB_TOKEN`:
+
+```toml
+[run.integrations.github]
+additional_repositories = ["fabro-sh/keystone"]
+permissions = { contents = "read" }
+```
+
+The run origin stays implicit, and Fabro mints one installation token scoped to the origin plus every declared repository with the shared permission map. Inside stages, `gh` commands, raw GitHub API calls, plain Git over HTTPS, and the common SSH URL spellings (`git@github.com:owner/repo` and `ssh://git@github.com/owner/repo`) all work against the declared set — the SSH forms are transparently rewritten to authenticated HTTPS with no secret placed in Git configuration.
+
+Every repository must share one owner and be reachable by the origin's GitHub App installation. Preflight resolves each repository's installation, mints the scoped token once, and probes every repository with `git ls-remote`, naming the exact repository when something is not accessible; run initialization enforces the same checks. A declared-but-inaccessible repository fails the run before its first stage.
+
+Declaring additional repositories requires `contents = "read"` or `contents = "write"`. With `contents = "write"`, any stage can push to any declared repository — declare the smallest set and weakest permissions that work. See [Additional repositories](/integrations/github#additional-repositories) for details, including layering rules and `GH_TOKEN` precedence.
--- a/docs/public/docs.json
+++ b/docs/public/docs.json
@@ -295,6 +295,13 @@
         "icon": "clock-rotate-left",
         "groups": [
           {
+            "group": "August 2026",
+            "icon": "clock-rotate-left",
+            "pages": [
+              "changelog/2026-08-21"
+            ]
+          },
+          {
             "group": "July 2026",
             "icon": "clock-rotate-left",
             "pages": [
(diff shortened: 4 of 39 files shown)


brynary:
## Reviewer notes

Things worth extra scrutiny, in rough priority order:

### Behavior changes to existing code

- **`GitHubRepositorySlug` identity is now case-insensitive** (`Eq`/`Ord`/`Hash` over ASCII-lowercased owner/repo; `Display`/serde keep the submitted spelling). Existing users — `fabro-automation`, `git_checkout.rs`, `automation_materializer.rs` — compile unchanged, but any code that relied on exact-byte equality of two case-variant slugs would now see them as equal. I audited the call sites and found none that depend on the old semantics; please double-check `git_checkout.rs`, where the slug feeds bare-repo directory naming (naming uses `owner()`/`repo()` components directly, so case is preserved on disk).
- **The mint's 422 message changed** from naming one repository to listing all requested names (`lib/components/fabro-github/src/lib.rs`). Multi-repo callers resolve installations first, so in practice a 422 should be rare and already attributed.
- **ACP notice text** was generalized to say all declared repository access expires together. It's emitted for single-repo runs too — I judged the wording still accurate there.

### Deliberate design decisions

- **Serialization: `additional_repositories` is omitted when empty** (`skip_serializing_if`). This keeps single-repository settings byte-identical to earlier releases (persisted `run.created` events, snapshots, JSON parity tests). The plan's phrase "additive empty JSON field" could be read as always-serialize; I chose omission for wire compatibility. Cheap to flip if you prefer explicit `[]`.
- **Where validation lives, twice on purpose.** Config resolution rejects what it can offline (slug grammar, duplicates, cross-owner within the additional set, cap, literal `contents` values) with indexed paths; `GitHubRepositoryAccess::new` re-enforces everything at the runtime boundary including origin-dependent rules and interpolated `contents`. Preflight and workflow initialization both construct the same access value so they cannot disagree.
- **`GitHubRepositoryAccess::new` returns `anyhow::Result<Option<Self>>`.** `Ok(None)` encodes the legacy "no origin, nothing declared" state so both call sites keep their existing best-effort behavior; every declared-set invariant is a hard `Err`. PAT/static permissions-only runs skip access construction entirely to preserve their origin-agnostic behavior — see the `declares_additional || App` gate in `build_sandbox_env`.
- **Bridging goes into `base_env` once at initialization**, not into every per-stage resolve. The credential helper reads `$GITHUB_TOKEN` at invocation time, so token refresh flows through the existing per-stage env resolution with no bridging update. The helper is host-scoped to `https://github.com` (documented failure-mode change for undeclared private repos: authorization error instead of missing-credential error).
- **`insteadOf` rewrites match by prefix** — `owner/repo` also rewrites `owner/repo-other`'s SSH spelling. Authority is unchanged (the token is server-side scoped), only the transport changes. Documented in the module header, the docs page, and covered by a contract test.
- **Preflight probes retry with the same token** (never re-mint) on auth-shaped failures, classified via the existing `fabro_sandbox::classify_failure`, because replication of a given token only makes progress. The probe keeps the token out of argv/URLs by passing it through the child env only.

### Test seams

- `run_github_token_check_with` in `run_manifest.rs` mirrors the existing `run_repository_access_check_with` pattern: the mint and per-repo probe are injected, so the new preflight tests script them. The real installation-resolution/mint behavior is covered by `fabro-github` unit tests with a mock HTTP client.
- `git_bridge` contract tests run the installed `git` binary against local `file://` fixtures (a test-only base-URL parameterization of the entry builder), with `GIT_SSH_COMMAND=false` proving SSH is or isn't attempted — no network involved. They create a `.git`-suffixed bare fixture plus a unix symlink for the bare spelling.
- The live e2e (`fabro-github/tests/live_access.rs`) needs `FABRO_TEST_GITHUB_APP_ID`, `GITHUB_APP_PRIVATE_KEY`, `FABRO_TEST_GITHUB_ORIGIN`, `FABRO_TEST_GITHUB_ADDITIONAL_REPO`. It includes a negative check: a primary-only token must NOT read the additional repository.

### Not done here

- The post-deploy acceptance run from the plan (preflight against `fabro-sh/keystone`, in-stage SSH clone, `gh api`, negative run) — needs a deployed server with App credentials.
- Cross-owner/cross-installation sets, per-repository permissions, and managed multi-repo checkout remain explicit follow-up scopes.

PR #780, opened 2026-08-21
Add version-backed run intent creation

swerner:
## Summary

- let `POST /api/v1/runs` temporarily accept either the existing self-contained manifest or a strict, version-backed `RunIntent`, so callers can migrate independently
- load and lower the immutable workflow-version closure, select the requested server environment (defaulting to `default`), validate the first `git` target shape without GitHub/network admission, and create a durable `submitted` run without starting it
- persist the canonical accepted target alongside its operational Git projection, preserve it through retry/fork/replay, and activate an optional exact SHA only when the clone-based sandbox starts
- regenerate the Rust and TypeScript API clients and document the compatibility and exact-checkout behavior

## Explicit design choice for team review

This PR makes the selected server environment authoritative for sandbox provider, working directory, and image. Workflow-version settings may still overlay resources, network, lifecycle, labels, and environment variables. Please confirm that this field-level authority split is the contract we want before merge.

Related boundaries kept intentionally explicit:

- creation returns `submitted`; starting remains a separate operation
- admission does not contact GitHub, resolve a branch, prove branch/SHA ancestry, or fall back from an unavailable exact SHA to branch HEAD
- intent compilation does not read repository `project.toml` settings or target-project files; version-owned workflow config, Dockerfiles, goal files, and dependency files are lowered from the immutable closure
- `git` is the only live target arm in this change; Git targets currently require clone-enabled Docker or Daytona environments, while `none`, `folder`, and Local target support remain additive follow-ups
- the submitted definition blob remains available, and legacy manifest parsing stays on the same route until production callers migrate and the API compatibility/removal gate is decided

## Verification

- `cargo build --workspace`
- `cargo +nightly-2026-04-14 clippy -p fabro-types -p fabro-api -p fabro-client -p fabro-environment -p fabro-sandbox -p fabro-workflow -p fabro-server --all-targets -- -D warnings`
- `cargo +nightly-2026-04-14 clippy -p fabro-server --all-targets --features test-support -- -D warnings`
- targeted RunIntent, closure-lowering, exact-checkout, retry, fork, API identity/parity, and legacy manifest tests
- `cargo nextest run -p fabro-server --features test-support all_spec_routes_are_routable`
- `cargo dev docs check`
- `bun run typecheck` in `lib/packages/fabro-api-client`


swerner:
Ran a four-angle cleanup review (reuse, simplification, efficiency, altitude) over this PR and pushed the resulting fixes as four commits. No behavior changes intended; the wire contracts, error messages, and log lines of both create lanes are preserved.

## Applied

- **Git-target grammar unified** (`7fd2f2c83`): the slug/branch/SHA rules and derived origin URL were implemented twice — server admission and sandbox start — with no shared code path. They now live once as `RunTarget::validate()` in fabro-types, returning the canonical target plus its derived `GitContext`. Start re-derives the expected clone source through the same rules before checking the persisted projection, so admission and activation can no longer drift. Admission also now runs the pure in-memory checks (target grammar, environment id) before the blob-store closure fetch, so malformed requests don't pay for version-store I/O.
- **Shared create-run pipeline tail** (`7e811c89c`): the intent and legacy-manifest handlers each carried a ~120-line copy of the post-admission sequence (LLM readiness → compile/pin → persist → summary → managed-run insert → title generation → 201). They had already drifted on when the run ID is resolved. Both lanes now share one `finalize_created_run`, with a small `CreatedRunErrorStyle` keeping each lane's pinned error mapping intact, and both resolve identity before compilation — which also let the `PinnedRun` copy of `PreparedRun`'s identity accessors be deleted.
- **Request-parsing overhead** (`88c2b5759`): the dispatcher deep-cloned the parsed JSON body twice per request (once per candidate shape), so every legacy manifest create copied its full workflow bundle twice. Both shapes now deserialize from a reference to the parsed value. Also simplified `inline_goal_file` to bind the goal slot once instead of re-navigating and `expect`-ing.
- **Client duplication** (`81b22cf2e`): the two byte-identical `create_run_from_*` client methods now share one private `submit_create_run(CreateRunRequest)`.

## Considered and deliberately skipped

- **Reusing `manifest_path.rs`'s `relative_path_from` for closure-lowering rebasing**: its contract returns `None` for an empty base, which the lowering path hits for top-level entrypoints (covered by this PR's tests), so switching would break behavior.
- **`RunCompilerSettingsInput::LegacyManifest` enum in the compiler**: it marks the temporary legacy seam this PR explicitly keeps until callers migrate; removing the lane later is a one-arm delete, so the explicit enum seems worth its cost.
- **Relocating `pin_workflow_environment_authority` / expressing environment authority as a merge-engine rule**: the PR asks the team to confirm the field-level authority contract first; moving the mechanism before that decision seems premature.
- **A shared provider/image compatibility predicate, typed scalar inputs for `RunIntentArgs.inputs`, and sharing file contents across repeated mounts**: each is a reasonable follow-up but changes pre-existing call sites, wire error details, or existing type design beyond this PR's scope.

## Verification

- `cargo build --workspace`
- `cargo +nightly-2026-04-14 clippy -p fabro-types -p fabro-workflow -p fabro-server -p fabro-client --all-targets -- -D warnings` (and `-p fabro-server --features test-support`)
- `cargo nextest run -p fabro-types -p fabro-workflow -p fabro-client -p fabro-api -p fabro-server` — 2,899 passed
- Each of the four commits builds independently.

PR #783, opened 2026-08-22
Activate verified SQLite blob storage at server startup

swerner:
## Summary

- gate server startup on a verified SQLite blob authority, including legacy inventory/import, retained backup validation, disk preflight, full source/target verification, SQLite integrity checking, and the final WAL checkpoint
- require every runtime run handle, including manager-loop children, to reuse the same SQLite-backed blob store; keep Slate-backed blob access test-only
- add activation and migration regression coverage plus operator documentation for the forward-only recovery boundary and evidence-gated compatibility window

## Validation

- full Rust workspace: 7,993/7,993 tests passed on the implementation base after excluding `validate_reports_missing_template_dependency`; that existing snapshot is checkout-path-sensitive and renders this temporary `/private/tmp` checkout instead of `[FIXTURES]`
- post-rebase focused suites: legacy blob import 25/25, startup activation 10/10, manager loop 21/21, workflow-version store 9/9
- `cargo dev docs check`
- Bun 1.3.14 web tests: 807/807; web and API-client typechecks
- `cargo dev build -- -p fabro-cli --release`


swerner:
Ran a simplification pass over this PR (reuse / simplification / efficiency / altitude). Applied in 65d495ce9 and 5cfca6c88 — net −153 lines. Verified with a full workspace build (without `test-support`), `check --all-targets`, clippy `-D warnings`, fmt, and the affected crates' test suites.

## What was fixed

**Reuse of existing `fabro-db` helpers** — `blob_activation.rs` had verbatim copies of `append_to_path`, `remove_file_if_exists`, and the `set_private_permissions` cfg-pair that already lived privately in `fabro-db`. Those are now `pub` in `fabro-db` (with `set_private_permissions` returning `io::Result` so callers own their error context) and `blob_activation.rs` calls them instead of redefining them.

**Discarded report wrapper removed** — `activate_blob_storage` returned an `ActivatedBlobStorage` whose `BlobActivationReport` payload was explicitly dropped by its only accessor (`let _ = report;`). It now returns `Arc<fabro_store::Database>` directly and emits the final `info!` from locals.

**`BlobDiskPreflight` echo struct removed** — `compute_disk_preflight` mostly returned its own inputs back for logging. It now takes three args and returns just the computed `required_free_bytes`; the debug log uses locals.

**Double backup validation removed** — the freshly `VACUUM INTO`-ed staging file was integrity-checked, atomically renamed, then fully integrity-checked again on identical bytes. The post-publish validation now runs only in the `AlreadyExists` branch, where it genuinely examines a different file.

**Duplicated integrity-check protocol unified** — the "exactly one `ok` row from `PRAGMA integrity_check`" loop existed twice (backup vs live pool); now one `integrity_check_is_ok` helper generic over `sqlx::Executor`. The two-variant `PublishOutcome` enum also collapsed to a bool.

**`resource_sampler.rs` copy-paste** — `available_space_for_path` duplicated the `Disks` → `DiskCandidate` enumeration block byte-for-byte from `sample_disk_resources`; both now call a shared `refreshed_disk_candidates()`.

**~32 test fixtures routed through the PR's own helper** — the PR added `fabro_store::test_support::test_database(...)` but bypassed it at ~32 sites that hand-assembled `Database::new(..., test_blob_store())` across fabro-workflow, fabro-server, and their integration tests. All now call the helper, and its fabricated `Result` return (unconditional `Ok`) is gone, so the 19 pre-existing callers dropped their pointless unwraps too. The next `Database::new` signature change touches one helper instead of ~30 files.

**Fragile cross-crate `include_str!`** — `test_blob_store` reached four directories up into fabro-db's migration files by relative path. fabro-db now exports `BLOBS_MIGRATION_SQL`, and fabro-store's `test-support` feature activates an optional `fabro-db` dependency (production builds unchanged).

**`serve.rs` double error reporting** — the activation `map_err` hand-rolled `anyhow::Error::new(...).context(...)` plus an `error!` of the flattened chain, unlike every other returned startup error. Now a plain `.context("activating SQLite blob storage")?`; the unused `error_util` import is gone.

## What was skipped (and why)

- **Startup does three full scans/SHA-256 passes over legacy blobs, and full-store verification recurs on every restart.** The fail-closed, re-verify-everything behavior is documented intent in `server-operations.mdx` for this activation bridge — weakening it changes intended behavior, so it deserves a deliberate decision rather than a cleanup pass. (If wanted later: the inventory pass doesn't need the digest check, and the verify source pass's byte-equality against a digest-validated target row already transitively proves the source digest.)
- **`tokio::join!`-ing the independent startup I/O** (inventory scan, backup validation, disk sampling): real but one-time startup latency; joining adds ordering subtlety to deliberately paranoid code.
- **Consolidating the triplicated `checked_add`/`usize_to_u64` helpers and the three parallel failure enums in `legacy_blob_import.rs`**: making them generic means threading the overflow variant through ~35 call sites (or adding a trait), which costs more than the two duplicated 4-line helpers it saves — especially in a module scheduled for deletion with the bridge.
- **Dropping the `&BlobStore` parameter from `import_legacy_blobs_into` / `verify_legacy_blobs_in`**: verification tests legitimately pass a target distinct from the database's own store, so it's not clearly vestigial.
- **Shrinking `LegacyBlobVerificationReport`'s write-only fields**: they're read by tests and plausibly deliberate migration evidence for the observation window.
- **Follow-up observation, not changed here**: `manager_loop.rs` now correctly shares the parent's run store with sub-workflows, but the sibling `artifact_store` on the same line is still a throwaway `InMemory` store, so child-workflow artifacts vanish. Fixing it needs a wider seam (the parent's `ArtifactSink` lives on `Initialized`, not `EngineServices`).

Note: 5 fabro-workflow tests fail locally with real LLM provider credentials in the environment (they assert no credentials resolve) — pre-existing on this branch, unrelated to these changes, and they pass in a clean environment/CI.

swerner:
A code review of this PR surfaced ten issues; each is fixed in its own commit, pushed to this branch. All fixes verified with `cargo build --workspace`, the full workspace test suite (8077 passed), `cargo fmt --check`, and workspace clippy with `-D warnings`.

**Correctness / operability**

- **Forward fabro-store/test-support through dependent test-support features** (57d0efa) — `fabro-workflow` and `fabro-server` import `fabro_store::test_support` from their feature-gated `test_support` modules but never forwarded `fabro-store/test-support`, so per-crate builds like `cargo check -p fabro-cli --tests` failed with E0432 (workspace builds passed only through feature unification). One line in each Cargo.toml.
- **Size the disk preflight to the remaining import work** (cb5bc6f) — the preflight demanded ~1.5× the *full* legacy inventory free on every boot, but a successful activation itself consumes ~2× those bytes and the legacy source is retained for the whole window, so a large activated server could become unable to restart. The inventory now reports which rows are still missing from SQLite and the preflight sizes only those; a warm restart needs only the fixed 64 MiB headroom.
- **Skip the preflight when no mount matches the database path** (2a551c8) — `available_space_for_path` returning `None` (tmpfs/squashfs roots, network filesystems, unreadable mount table) was a fatal `UnknownFilesystem` even with zero legacy rows, while the resource sampler treats the identical condition as benign. It now warns and skips the capacity check; import, verification, and integrity checks still run.
- **Continue startup when the final WAL truncate checkpoint reports busy** (a05f03e) — any external reader outliving the 5 s busy timeout (backup tool, replication agent, operator shell) crash-looped every boot over a condition with no integrity impact. Busy now logs a warning and startup continues; also restores the checkpoint failure-path test coverage that was lost in the relocation.
- **Make snapshot and backup publication durable** (93fa52a) — neither the `.pre-blob-activation.bak` nor the pre-migration snapshot fsynced the staged file or the parent directory around the publishing rename, so a crash after the import committed could silently lose the documented rollback artifact (and the next boot would write a "pre-activation" backup that already contains the imported blobs). Staged files are now synced before publish and both publishers sync the destination directory after their rename.

**Startup cost**

- **Cut redundant blob scans and hashing from startup** (cc30dc0) — every boot scanned the legacy keyspace three times, SHA-256-hashing every value each time, then re-read and re-hashed every live SQLite blob row. The inventory pass now validates key shapes only (the import pass still validates every digest before persisting), and the independent verification sweep runs only on boots whose import actually inserted rows — a no-op restart is already fully byte-compared by the import pass itself.

**Test infrastructure**

- **Isolate the shared test blob store** (6a38790) — `test_blob_store` was a process-wide `OnceLock` singleton, so blob state bled across tests in one process; plain `cargo test -p fabro-workflow-version --lib` failed 8/24 (nextest's process-per-test masked it). Each call now gets an isolated store; reopen-style tests share a durable per-store-directory blob database explicitly via `test_blob_store_at` / `test_database_with_blobs`.

**Design / reuse**

- **Pass the SQLite pool directly to import and verification** (8976e3f) — removes the always-`Some` `sqlite_pool_for_legacy_import` accessor, its clippy suppression, and both production-unreachable `WrongTargetBackend` variants; the activation path no longer round-trips a pool it already owns.
- **Share one snapshot-staging helper between fabro-db and activation** (11c6388) — `create_backup` had hand-copied fabro-db's stage-a-VACUUM-INTO sequence (which is why those fs helpers had been made pub). Both now stage through `fabro_db::write_snapshot_to_staging`; the publish halves stay separate on purpose (overwrite rename vs. `persist_noclobber` + validation), and the copied helpers are private again.
- **Register the activation bridge as a server migration** (e39e1a0) — the self-described temporary compatibility bridge now follows `docs/internal/migrations-strategy.md`: it lives at `migrations/2026082301_sqlite_blob_activation.rs`, is registered through `src/migrations.rs`, carries a `REMOVAL_DEADLINE` (an eligibility floor — removal still requires the evidence and explicit approval in the module docs), and logs `removal_deadline` on every activation.

The server-operations doc was updated to match the new preflight and verification behavior.

PR #789, opened 2026-08-23
Add empty workspace run target

swerner:
## Summary

- add a strict `{"kind":"none"}` run target for repository-free workspaces
- admit the target for Docker and Daytona while forcing clone suppression at start
- reject the target for Local environments until scratch-workspace allocation exists
- keep Git fields absent through persistence, retry/replay, sandbox startup, and pull-request routing
- publish the OpenAPI contract, generated TypeScript client, docs, and changelog entry

## Base

This PR is rebased directly onto `main` after #780 merged. Its single commit contains only the empty-workspace target addition and preserves #780's final admission and clone-projection design.

## Behavioral notes

- Creation remains a durable admission step and returns a submitted run; it does not start execution.
- A workflow's clone setting remains persisted unchanged, but the `none` target is authoritative and prevents cloning for Docker and Daytona.
- Local scratch workspaces and local-folder targets remain separate follow-up capabilities.

## Verification

- `cargo check -p fabro-types -p fabro-api -p fabro-workflow -p fabro-server --all-targets --features fabro-server/test-support`
- focused `cargo nextest` run: 27 passed
- `cargo +nightly-2026-04-14 clippy -p fabro-types -p fabro-api -p fabro-workflow -p fabro-server --all-targets --features fabro-server/test-support -- -D warnings`
- `bun run generate` and `bun run typecheck` in `lib/packages/fabro-api-client`
- `cargo dev docs check`
- `git diff --check`


swerner:
## Simplification pass (commit 3e0adde73)

Ran a reuse / simplification / efficiency / altitude review over this PR and applied the fixes. Net **−104 lines**, no behavior change.

**`fabro-types/src/run_intent.rs`**
- Replaced the hand-rolled `Deserialize` (three wire structs + untagged dispatch + manual impl) with the plain derive. Serde only ignores sibling fields on internally tagged *unit* variants, so `None` is now an empty struct variant `None {}` and `deny_unknown_fields` applies directly. `{"kind":"none","x":1}` is still rejected, and error messages are now serde's field-level ones instead of "did not match any variant".
- Added `RunTarget::kind_name()` (strum `IntoStaticStr`) for diagnostics.

**`fabro-workflow/operations/start.rs`**
- `clone_source_for_run` is now the single owner of the empty-workspace decision: `CloneSourceForRun` gained `skip_clone`, and the Docker/Daytona arms consume it uniformly. Removed `target_requires_empty_workspace` and the duplicate `None` check / unreachable `ok_or_else` error path.
- The two near-identical Local-provider error arms collapsed into one using `kind_name()`.

**`fabro-server/handler/runs.rs`**
- Destructure `ValidatedRunTarget { target, git }` directly.
- `validate_intent_environment`: two back-to-back matches over `target` → one match yielding `(incompatible, detail)`.

**`fabro-workflow/operations/create.rs`**
- Dropped the redundant `git = None` override; admission already derives `git` from the target. The `source_directory` override stays.

**`fabro-server/server/tests.rs`**
- The two byte-identical unavailable-integration request loops became `assert_run_intent_targets_unavailable(&state)`.

**Skipped**
- Nulling `source_directory` for every `Some(target)` (would change Git-target behavior).
- Removing the `target.clone()` in the handler (needed since environment validation runs after the input is consumed; a bool would cost readability for a tiny clone).

Verified: `cargo fmt --check`, nightly clippy `-D warnings`, and `cargo nextest run` for fabro-types / fabro-api / fabro-workflow / fabro-server all pass.

PR #790, opened 2026-08-23
Add local folder run target

swerner:
## Summary

- add the strict `{ "kind": "folder", "path": "/absolute/server/path" }` run target for existing server directories
- admit folder targets only for Local environments, canonicalize and validate the path before run allocation, persist it as the target-authoritative source directory, and revalidate it before sandbox construction
- keep intent compilation isolated from target-project files while observing optional named-branch Git metadata through read-only, no-network libgit2 calls after provider admission
- preserve the canonical folder identity through event replay and retry, reject Docker and Daytona at both admission and start, and publish the OpenAPI/generated-client/docs contract

## Stack

This PR is stacked on #789, which is stacked on #780. Review only this PR's commit; after the parent PRs merge, this branch can be retargeted to `main`.

## Trust and lifecycle boundaries

- An authenticated run-management caller may name any server-accessible absolute directory. This target is intended only for trusted single-tenant deployments, and the Local-provider gate is the deployment safety boundary.
- The workflow compiler continues to resolve version-owned files from the immutable workflow closure at `/workspace`; it does not read goals, settings, or Dockerfiles from the folder target.
- Folder runs execute in place without Fabro Git checkpoints. Retry retains the canonical target, while fork and rewind fail with an explicit unsupported error. This PR does not add Local Git cloning, checkpoint restoration, or Local scratch workspaces for the `none` target.

## Verification

- `cargo check -p fabro-server -p fabro-workflow`
- focused four-crate `cargo nextest` run: 27 passed
- `cargo +nightly-2026-04-14 clippy -p fabro-types -p fabro-api -p fabro-workflow -p fabro-server --all-targets --features fabro-server/test-support -- -D warnings`
- `cargo +nightly-2026-04-14 fmt --all -- --check`
- regenerated the TypeScript API client and ran `bun run typecheck`
- `cargo dev docs check`
- two independent final diff reviews with no remaining blockers


swerner:
## `/simplify` pass (9dc39ce9f)

Four cleanup reviewers (reuse, simplification, efficiency, altitude) went over the diff; findings were deduped and applied in 9dc39ce9f. Net +99/−184 across 4 files.

### Fixed

- **`run_intent.rs`** — Folder admission was dispatched three times (`prepare_intent_target` pass-through, a `matches!` in the handler, a `let Folder = …` in `observe_folder_git_context`). Collapsed into one `prepare_intent_target` that canonicalizes, checks the directory, and observes Git under the same provider gate. Removed `observe_folder_git_context`, its `PathBuf` clone, and the two copy-pasted `warn!` arms.
- **`handler/runs.rs`** — Single call, no `matches!`. Dropped the `git.clone()` / `target.clone()` fed into `RawRunCompilerInput` only to be overwritten by `with_target_and_git` (the compiler only forwards those fields).
- **`operations/start.rs`** — Hoisted the identical Folder-rejection blocks out of the Docker and Daytona arms into one pre-match check; restored `target.kind_name()` for the Git/None arm (the PR added `kind_name` but left it unused). In `folder_working_directory_from_record`, removed the unreachable `is_absolute` and post-`canonicalize` symlink checks (the "redirected" test already lands on "no longer canonical") and switched to `fs::metadata`.
- **Tests** — `canonical_folder(&temp)` fixture replaces six copies of the create/canonicalize/`to_str` ritual in `start.rs`; `local_test_app_state()` and `folder_intent()` replace 4+5 duplicated blocks in `server/tests.rs`; `post_run_manifest` now delegates to `post_run_intent_response`, which uses the existing `json_request` helper.

### Skipped

- Sharing one folder-validation helper between admission (`fabro-server`) and startup (`fabro-workflow`) — the sites want different error shapes (422 codes vs engine errors) and startup is intentionally stricter; not worth the cross-crate error plumbing.
- Folding `sanitized_origin_url`'s query/fragment/any-scheme stripping into `fabro_github::normalize_repo_origin_url`, and making `fabro-manifest::build_git_context` reuse `observe_git_context` — both change behavior for callers outside this diff.
- Removing the `source_directory == target.path` drift check at startup — deliberate defensive invariant.
- A `RunTarget` policy method to unify admission/start/fork target-vs-provider rules — admission also depends on `clone.enabled`, so a shared method wouldn't replace all three sites.

### Verification

`cargo fmt` and `clippy -D warnings` clean on `fabro-server` + `fabro-workflow`; both crates' suites pass except pre-existing failures (`auth::cli_flow`, `pipeline::pull_request`/`finalize`) that reproduce identically with these changes stashed.

PR #792, opened 2026-08-24
Move pending CLI authorizations to SQLite

swerner:
## Summary

- move pending CLI authorizations from SlateDB to SQLite
- persist only SHA-256 hashes of one-time authorization codes and atomically consume them with `DELETE ... RETURNING`
- leave legacy SlateDB authorization-code records inert at cutover; no new code reads them and they expire within 60 seconds
- preserve the existing OAuth status codes, response bodies, and one-attempt exchange behavior
- document the brief upgrade interruption for a login between browser approval and token exchange

## Testing

- `cargo build --workspace`
- `cargo +nightly-2026-04-14 clippy -p fabro-db -p fabro-store -p fabro-server --all-targets --features fabro-server/test-support -- -D warnings`
- `cargo nextest run -p fabro-db -p fabro-store`
- `cargo nextest run -p fabro-server -E 'test(/auth::cli_flow::tests/) | test(authorization_code_retirement_failure_is_fatal_before_startup)'`
- `cargo nextest run -p fabro-server --features test-support -E 'test(/cli_auth_token/)'`


swerner:
Ran a quality/cleanup review over this PR (reuse, simplification, efficiency, and altitude lenses) and pushed the fixes as 2f204af99. No behavior changes intended; `fabro-store`, `fabro-db`, and `fabro-server` suites all pass, fmt + clippy clean.

**Fixed**

- `retire_authorization_code_keyspace` was a line-for-line copy of `retire_refresh_token_keyspace`; both are now one-line wrappers over a private `Database::retire_keyspace(prefix)`.
- Inlined the single-use `retire_authorization_code_records` wrapper in `serve.rs` (same `.context(...)` + fatal `?`), dropping its test — which mostly asserted anyhow's context chaining — and the `test_close_slate` hook that existed only for it.
- The two independent SlateDB retirement scans at startup now run concurrently via `tokio::join!`. Error policies are unchanged (codes fatal, refresh tokens best-effort) and the asymmetry is now called out in a comment.
- New crate-private `sqlite_row` module with typed `identity_from_row`/`timestamp_from_row`; both `AuthorizationCodeStore` and `AuthSessionStore` use it, so the session store's stringly `Error::Other` corruption errors became the typed `InvalidStoredIdentity`/`InvalidStoredTimestamp` variants (previously introduced with a `record` parameter but only one user).
- Deleted `Repository::gc`, which had no production callers left and was kept alive by its own test; updated the record-layer module docs that still taught `gc` as the production pattern and noted `JsonCodec` is currently test-only.
- Deduplicated the SQLite test-support bootstrap (third copy) into a shared `sqlite_test_pool()` helper.
- The invalid-timestamp test now uses `issue()` plus a one-line `UPDATE` instead of a copied 10-column `INSERT`.
- Folded `oauth_authorization_codes` into the existing table-existence loop in the fabro-db schema test instead of a bespoke 8-line check.

**Considered and skipped**

- Dropping `oauth_authorization_codes_by_expiry`: the table is tiny (rows live ≤60s), but the index mirrors `refresh_tokens_by_expiry`, so consistency wins.
- Hashing the code at the HTTP boundary instead of inside the store (matching how refresh tokens do it): hashing in the store is this module's documented, deliberate design; changing it is a design decision, not a cleanup.
- Routing the keyspace retirement through the dated-migrations convention (`REMOVAL_DEADLINE` etc.): the scan runs on every boot forever and has no removal trigger, but the same applies to the pre-existing refresh-token retirement — flagging for a deliberate decision rather than restructuring here.
- Consolidating the per-store pool-close test hooks (`AuthorizationCodeStore::test_close` vs `RunSummaryStore::close_pool`): the right home is a shared helper at the pool owner's layer, which is beyond this diff.

swerner:
Follow-up to the cleanup pass: 999fb9eda drops the SlateDB `auth/code` keyspace retirement entirely rather than restructuring it into the migrations convention. The records it deleted are inert — at most a handful exist at cutover, every binary (old or new) rejects them within 60 seconds of issue via the expiry check, and nothing reads the keyspace after the move to SQLite — so the PR now cuts over with no one-shot compat code to remember to remove later. The pre-existing refresh-token retirement from #648 keeps its original shape. Rationale is noted in a comment at the startup call site.

PR #817, opened 2026-08-27
Add inactive SQL run event storage foundation

swerner:
## Summary

- Make the SQLite `RunSummaryStore` an explicit, database-owned dependency of every run handle (replacing the `OnceLock` attach dance), and introduce the private `PreparedRunSummary` row type.
- Add the final inactive `run_events` schema with its constraints and candidate indexes, without changing the existing `runs` table.
- Add caller-owned SQL primitives for atomic first-event creation, later event/current-row transitions, and the required event reads, with hermetic coverage for rollback, stale heads, corruption, filters, paging, gaps, and query plans.

The store keeps its `RunSummaryStore` name for now: today it is still the summary read model and only grows an inactive events table. The rename to a run-authority name happens once, at the SQL cutover.

## Behavior boundary

Production run history, run existence, session reverse lookup, and event paging remain on SlateDB. The existing best-effort runs-row refresh behavior is preserved. This PR does not call the new run_events write path, import or dual-write data, activate a runtime switch, transfer authority, enforce unique session ownership, or add restart recovery.

## Verification

- cargo +nightly-2026-04-14 clippy -p fabro-db -p fabro-store -p fabro-server --all-targets -- -D warnings
- cargo nextest run -p fabro-store -p fabro-db -p fabro-server (1150/1153; the 3 failures are graphviz SVG rendering tests that also fail on main in this environment)


swerner:
## `/simplify` pass — ff2aef4

Four cleanup reviewers (reuse, simplification, efficiency, altitude) ran over this diff; findings were deduped and applied in ff2aef4. Quality-only — no correctness bugs were hunted.

### Fixed

**`run_record_store.rs`**
- Extracted `bind_run_columns()` — the 23-column bind ladder was duplicated verbatim between `write_insert_shaped_run` and `update_run_on_connection`; one helper now feeds INSERT/UPSERT/UPDATE.
- Seq guard computed once per append via `next_event_seq_after()`; removed the recompute in `update_run_on_connection` and the redundant `ensure_prepared_head` calls in `insert_run_on_connection` / `insert_event_on_connection` (entry points already check). Dropped the unreachable `seq == 0 || seq > MAX` check in `validate_event_for_record`.
- `stored_seq()` replaces the twice-written `u32::try_from(..).filter(range)` decode.
- `list_events_on_connection` reads the run head *before* decoding events, so a head mismatch bails without paying JSON-parse cost.
- `decode_event_rows` computes `run_id.to_string()` once per result set instead of per row.
- First-visit stage listing rewritten as `UNION ALL` instead of `stage_id = ? OR (stage_id IS NULL AND node_id = ?)`. The OR shape fell back to the primary-key range scan and never used `run_events_by_stage` / `run_events_by_legacy_node`; the query-plan test now asserts the real shape hits both indexes.
- Shared `INSERT_EVENT_SQL` between the production writer and the test seeder; tests gained a `seqs()` helper and lost a `needless_pass_by_value` lint exemption.

**`test_support/mod.rs`** — `test_blob_store` and `test_run_record_store` share one `lazy_in_memory_pool(migrations)` builder.

**Removed two tautology tests** (`required_run_record_store_is_shared_with_run_handles` + its `#[cfg(test)]` accessor, `run_record_store_is_database_owned`) — they only asserted a constructor stored an `Arc`; the behavioral tests already cover sharing.

**`fabro-db/tests/sqlite.rs`** — `insert_run_event` dropped three always-`None` params; `insert_run_with_id` / `insert_minimal_run` share one `insert_run_row`.

### Skipped (design-level or behavior-changing — flagging for the author)

- Making the denormalized `run_events` columns SQL `GENERATED` and dropping the read-time `RunEventMismatch` field checks; merging `RunEventMismatch` / `RunSummaryMismatch` into one variant (the `seq: 0` sentinel suggests the shape doesn't fit all call sites).
- Dropping `AppStores.run_records` in favor of `stores.runs.run_record_store()` at call sites.
- Replacing the `#[cfg_attr(not(test), expect(dead_code))]` staging with a real dispatch arm.
- Skipping the full summary rewrite on events that don't change the summary (per-event hot path once activated); removing the `json_valid` CHECK (double parse per insert with a single trusted writer).
- A `fabro_db::test_support::migrate_in_memory` entry point to replace exporting `*_MIGRATION_SQL` consts per table.
- Centralizing the `run.session.*` prefix shared between Slate's `is_run_session_event()` and the SQL GLOB/index literal.
- Converting static `QueryBuilder` uses to `sqlx::query(const)` (marginal).

### Verification
`cargo fmt --check`, clippy `-D warnings`, and `cargo nextest run -p fabro-store -p fabro-db -p fabro-server`: 1150/1153 pass. The 3 failures are graphviz SVG rendering tests that fail identically on the branch without these changes (local environment, not this PR).

swerner:
Rolled back the `RunSummaryStore → RunRecordStore` rename in 57bbb92. The store is still the summary read model today (it only gains an inactive `run_events` table here), and the rename made `run_summary_store.rs` show as a 932-line delete plus a 1920-line add while touching nine otherwise-unrelated files. The diff is now 11 files / +1453 −94 with the store file shown as a modification. The final rename to `RunStore` happens once, when the SQL store becomes the run authority. Title and description updated to match.

PR #820, opened 2026-08-27
Move automation runs to workflow versions and RunIntent

swerner:
## Summary

- Package each checked-out automation workflow as an immutable, validated workflow-version closure and store dependencies before parents.
- Submit automation runs through the same in-process RunIntent admission path as POST /api/v1/runs, while preserving the existing automation schema and repository-local workflow selection.
- Keep run creation and start separate: API and scheduled triggers still queue exactly one start only after successful creation.

## Behavior and compatibility

- Local packages use checkout-relative paths, and optional workflow.toml is discovered beside each entrypoint. Workflow-owned goal files, template includes, imports, child workflows, and Dockerfiles are included without rewriting source.
- Branch and tag targets resolve once per firing; the resulting run preserves branch/tag identity and is pinned to the exact peeled commit SHA used for packaging.
- Project and user settings are no longer packaged into automation run creation. With no automation environment selector, the server default environment remains authoritative.
- CLI, Playground, and run-create-tool producers remain on legacy manifests in this change.

## Validation

- Pinned-nightly formatting and workspace clippy with warnings denied
- 8,286 workspace Rust tests passed under the CI nextest profile
- Documentation refresh/check passed
- Bun 1.3.14 lockfile install, web typecheck, all 808 web tests, and API-client typecheck passed
- Optimized fabro-cli build passed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2827,11 +2827,13 @@ dependencies = [
  "fabro-test",
  "fabro-types",
  "fabro-workflow",
+ "fabro-workflow-version",
  "git2",
  "insta",
  "serde_json",
  "temp-env",
  "tempfile",
+ "thiserror 2.0.18",
  "toml 0.8.23",
 ]
 
--- a/lib/apps/fabro-server/src/automation_materializer.rs
+++ b/lib/apps/fabro-server/src/automation_materializer.rs
@@ -2,11 +2,12 @@ use std::path::PathBuf;
 use std::sync::Arc;
 
 use async_trait::async_trait;
-use fabro_api::types::RunManifest;
 use fabro_automation::AutomationId;
-use fabro_config::{EnvironmentLayer, MergeMap};
-use fabro_manifest::ManifestBuildInput;
-use fabro_types::{GitHubRepositorySlug, GitRunTarget, RunId, RunTarget, TargetValidationError};
+use fabro_manifest::{CollectedWorkflowClosure, WorkflowVersionCollectError};
+use fabro_types::{
+    GitHubRepositorySlug, GitRunTarget, RunId, TargetValidationError, WorkflowVersionId,
+};
+use fabro_workflow_version::{WorkflowVersionStore, WorkflowVersionStoreError};
 use tokio::{fs, task};
 
 use crate::git_checkout::{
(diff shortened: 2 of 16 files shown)


swerner:
## `/simplify` pass (077d02022)

Four review lenses (reuse, simplification, efficiency, altitude); findings deduped and applied where they didn't change behavior.

### Fixed
- **Duplicated `RunIntent` literal** in the scheduler and automations handler → `AutomationRunMaterialized::into_run_intent()` owns the shape once.
- **Pass-through wrappers** `PackageFromCheckoutInput` / `PackagedAutomationWorkflow` removed; the materializer calls `collect_workflow_versions` directly inside `spawn_blocking`.
- **`VersionIdMismatch`** removed — the store derives the same content hash the collector did, so it's a `debug_assert_eq!` tripwire instead of an unreachable error variant.
- **`workflow.toml`-beside-entrypoint rule** was duplicated in `run_intent::mount_version` and `fabro-workflow-version` → single `WorkflowVersion::config_path()`.
- **Version-ID derivation** duplicated in the collector and test fake → `WorkflowVersion::id()`.
- **Collector `VersionAssembler`** now owns the collected sources and moves file contents instead of cloning them; drops the redundant second `workflows.get`, the `dependency_keys` Vec copy, and the always-true `emitted` set.
- **Double `WorkflowLocation::resolve`** per run → `collect_versions` takes the resolved location; the not-found probe copy-pasted from `build_run_manifest` is a shared `resolve_existing_workflow_location`.
- **Goal file read twice** in `collect_config_goal_files` → reuses the content `collect_bundled_file` just inserted.

### Skipped (noted for follow-up)
- `workflow_version_projection: bool` mode flag on `WorkflowBundler` — turning it into a constructor-time policy / making goal-file collection unconditional changes the manifest lane's closure; separate PR.
- `CreateRunFromManifestRequest.{explicit_run_id, automation, target, explicit_title_supplied}` are now production-dead but exercised by six tests; separate cleanup.
- `WorkflowVersionStore::put` re-walks/re-validates dependencies per version (O(N²) for a chain) and `create_run_from_intent` re-reads the closure the materializer just wrote — real cost, but needs a `put_closure` / typed admission result beyond this diff.
- `explicit_run_id` pre-allocation seam, `MINIMAL_DOT` fixture consolidation, caching the production materializer on `AppState` — minor / out of scope.

Clippy clean; 1413/1416 tests in the touched crates pass (the 3 failures are the SVG tests needing a local `dot` binary).

PR #823, opened 2026-08-28
Activate atomic SQLite run history storage

swerner:
## Summary

- activate SQLite as the sole authority for run existence, event history, and current projections
- import and fully verify legacy Slate run history behind a fail-closed startup gate with a retained pre-import backup and durable activation marker
- make run creation, event append, and deletion atomic in SQLite, including deletion tombstones that prevent legacy resurrection
- serve run reads, pagination, cache warming, and SSE catch-up from committed SQLite state
- document activation behavior, operator recovery boundaries, and backup retention

## Verification

- `cargo nextest run -p fabro-db -E 'test(run_history_activation) | test(run_events)'`
- `cargo nextest run -p fabro-store -E 'test(run) | test(event) | test(legacy_run_history) | test(activation)'`
- `cargo nextest run -p fabro-workflow -E 'test(create) | test(retry) | test(fork) | test(resume)'`
- `cargo nextest run -p fabro-server --features test-support -E 'test(activation) | test(startup) | test(delete_run) | test(events)'`
- `cargo +nightly-2026-04-14 clippy -p fabro-store -p fabro-workflow --all-targets -- -D warnings`
- `cargo +nightly-2026-04-14 clippy -p fabro-server --all-targets --features test-support -- -D warnings`


swerner:
## `/simplify` pass (commit 0574c818c)

Four cleanup reviews (reuse, simplification, efficiency, altitude) over this PR's diff, deduped and applied. `cargo check --workspace`, nightly clippy `-D warnings`, and `cargo nextest run -p fabro-store -p fabro-server` (1141 tests) pass.

### Fixed

**Reuse / duplication**
- Hoisted the backup machinery duplicated between the blob and run-history activation migrations (`backup_exists`, `create_backup`, `validate_backup`, `integrity_check_is_ok`, permission checks, ~130 lines) into a shared `migrations/sqlite_activation_backup.rs` with one `BackupError`. Both migrations now carry a single `Backup(#[from] BackupError)` variant instead of eight copies each. `final_truncate_checkpoint` stays per-migration because the busy-checkpoint policies intentionally differ.
- `RunDatabase::build` calls `build_cached_projection` instead of re-implementing it; `build_cached_projection` drops its redundant `contains` query (the events query already yields `RunNotFound`).
- `run_store::next_event_seq` removed in favor of the identical `run_summary_store::next_event_seq_after`.
- `create_run` / `create_run_with_first_event` share a `reserve_new_run` prelude.
- `test_blob_store_at` / `test_run_summary_store_at` collapsed onto one `lazy_file_pool` builder.
- Catalog key layout (`runs/_index/by-start/<id>`) lives in `keys::run_catalog_prefix` / `run_catalog_key` / `parse_run_catalog_key` rather than being hand-spelled at three sites.

**Simplification / dead code**
- Removed the write-only `recent_events` ring buffer (no readers survived the diff); `DEFAULT_EVENT_TAIL_LIMIT` renamed to `EVENT_BROADCAST_CAPACITY`, its only remaining job.
- `activate_run_history` takes `&Database` and returns `()` instead of passing an `Arc` through unchanged.
- `persist_activation_record` is a single `INSERT … ON CONFLICT DO NOTHING`; no wrapping transaction and no re-read/re-verify of a row just verified.
- Dropped the discarded `legacy_run_history_is_activated` call and its `ActivationState` variant from verification.
- Tombstone preamble: `TombstoneCheckError` plus a per-tombstone query loop replaced by one `tombstoned_destination_present` EXISTS query.
- `From<LegacyRunHistorySourceFailure>` replaces five hand-built `LegacyRunHistorySourceIdentityError { failure }` constructions.
- Unreachable `continue` in `refill_from_sql`; `backup_exists` local no longer shadows the function.

**Efficiency**
- `decode_event_row_with_json` fetches `event_json` once per row instead of twice.

**Docs**
- `record/` module docs now say it is test-only for the retired Slate layout rather than the recommended production pattern.

### Skipped (flagged, not applied)
- **Startup rescans the legacy prefix three times and replays every run ~4×** (identity → import → verify → integrity check → `warm_projection_cache`), on every boot for the life of the bridge. Fast-pathing on a verified marker changes the fail-closed policy, and materializing the source once is a real refactor of `LegacyRunHistorySource`. Highest-value remaining item; deserves its own PR.
- **Double payload validation on the append path** (`append_event_envelope` validates, then `*_on_connection` validates again). The `_on_connection` helpers have ~15 direct test callers exercising standalone validation, so changing their contract is beyond a cleanup.
- **Test-only `Database::create_run` on the production type** shapes `build_empty` and the `expected_last_seq == 0` branch in `commit_event_locked`; ~60 call sites across crates.
- **`delete_canonical` branching on the migration's marker table** (unconditional tombstone would remove the `BEGIN IMMEDIATE` read-then-write and `test_mark_run_history_activated`) — behavior/policy call.
- **`Runs::find` reads SQL directly while `Runs::list` goes through the projection cache** — pick one read path.
- **`reconcile` / `upsert_projection` / `UPSERT_RUN_SQL` kept alive under `cfg(test)`** only to test themselves.
- **Migration conventions**: no `REMOVAL_DEADLINE` const or report type (sibling migrations have both); public `server-operations.mdx` describes mechanism and internal approval process rather than the operator contract.

PR #825, opened 2026-08-29
Add independent workflow sources to automations

swerner:
## Summary

- add optional branch, tag, or commit workflow-source coordinates to automations while preserving omission compatibility
- independently resolve credentials, check out, and package workflows without adding source data to RunTarget or RunIntent
- persist the source in SQLite, expose it through the existing automation API and UI, regenerate the client, and document the behavior

## Verification

- cargo nextest run --workspace --status-level slow --profile ci: 8,339 passed, 208 skipped
- focused automation, database, API, server, and UI tests
- docs check, TypeScript typechecks, frozen Bun install, and release CLI build

## Known baseline test issue

The full web suite consistently passes 810 of 813 tests but has a pre-existing async timing race in InstallApp tests. Repeated full and isolated runs fail shifting subsets while all new automation UI tests pass; no install-flow files are changed by this PR.


swerner:
## Simplify pass

Ran a four-lens cleanup review (reuse / simplification / efficiency / altitude) over this PR and applied the fixes in the second commit. Also rebased onto `main` to pick up the automation environment selector (#828).

### Fixed

**`automation_materializer.rs`**
- Collapsed the four role-paired error variants (`TargetCredentials`/`WorkflowSourceCredentials`, `TargetCheckout`/`WorkflowSourceCheckout`) into `Credentials { role, .. }` / `Checkout { role, .. }` with a `CheckoutRole` enum; only the message varies.
- Replaced the `#[cfg(test)] clone_urls` field and the test-only branch inside the production `prepare_checkout` with a single `AutomationGitRemoteResolver` seam returning `GitRemote { clone_url, auth }`. No test-only state remains on the production struct.
- Replaced `reuse_target_checkout: bool` + two `.expect()`s with one `separate_source: Option<(slug, &source)>` computed up front.
- A workflow source in the target's repository (different ref) now reuses the already-resolved credentials instead of minting a second token.
- Test failure enum converts via `From`; the pre-existing test that built a version store inline now uses `test_version_store()`.

**`git_checkout.rs`** — the GitHub-URL-deriving `prepare_worktree` wrapper lost its only caller and was removed; the remaining method is `prepare_worktree(args, clone_url)`.

**`model.rs` / `store.rs`** — inlined the one-line `normalize_workflow_source` wrapper; dropped the `as_str()` wrapper on the new kind enum (store binds via `IntoStaticStr`).

**`fabro-db/src/lib.rs`** — removed the unused `AUTOMATION_WORKFLOW_SOURCES_MIGRATION_SQL` constant.

**`automation_scheduler.rs`** — `workflow_source`/`workflow` are moved rather than cloned; `create_automation` is a thin wrapper over `create_automation_with_source`.

**Web form/routes** — one `isWorkflowSourceRefValid(kind, ref)` predicate shared by `isFormValid` and the `aria-invalid` display; the three parallel `switch (kind)` helpers plus the hard-coded `<option>` list collapsed into a `WORKFLOW_SOURCE_KINDS` table; `workflowSourceLabel()` / `RUN_TARGET_CHECKOUT_LABEL` replace the repeated "run target checkout" literal.

### Rebase notes

- The workflow-source migration was renumbered `2026082801` → `2026082802` because `main` already uses `2026082801` for automation environments.
- The canonical-bytes fixture in `model.rs` now includes `environment_id`, and its expected revision hash was updated accordingly.

### Considered but skipped

- **Move the branch/tag/commit selector into `fabro-types` and wrap `TargetValidationError`** instead of adding four `InvalidWorkflowSource*` variants — a fair altitude point, but it changes the public error vocabulary and wire-adjacent types; too large for this pass.
- **Run target and source checkouts concurrently** — small latency win only for different-repo sources, and it competes with the credential-reuse fix above.
- **Have `validate()` return the parsed slug** to avoid the re-parse in the materializer — needs a `fabro-automation` API change; the re-parse is cheap.
- **`TestMaterializeFailure::InvalidWorkflowSource` carrying the real error** — `AutomationValidationError` isn't `Clone`, so it stays a unit variant (now documented).
- **`useMemo` around `mapAutomations`** — extends a pre-existing pattern, not introduced here.
- **Shared git-fixture seeding between `git_checkout` and materializer tests** — test-only duplication.

PR #826, opened 2026-08-29
Add local RunIntent producer support

swerner:
## Summary

- add deterministic local workflow package resolution with canonical containment and collected-closure output
- add collected workflow validation and lowering without introducing a new public source abstraction
- add truthful Git run-target observation while preserving legacy run-manifest behavior

This intentionally does not migrate CLI, `fabro_run_create`, automations, or HTTP registration consumers.

## Verification

- `cargo nextest run --workspace --status-level slow --profile ci` — 8,332 passed, 208 skipped
- `cargo nextest run -p fabro-manifest`
- `cargo nextest run -p fabro-server manifest_validation`
- `cargo nextest run -p fabro-server run_intent`
- `cargo dev docs check`
- `cd apps/fabro-web && bun run typecheck`
- `cd lib/packages/fabro-api-client && bun run typecheck`

## Known unrelated gate issue

`cd apps/fabro-web && bun run test` currently has three stable `InstallApp` failures that remain at “Connecting to install session…”. This branch does not modify `apps/fabro-web`; 11 tests in that file pass and 3 fail. The workflow therefore short-circuited before its release-build stage.

The workflow's forbidden-symbol command also names the nonexistent `lib/crates` directory. Running the same audit over the repository's actual `lib`, `apps`, package, and API-spec paths found no matches.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2826,6 +2826,7 @@ dependencies = [
  "fabro-template",
  "fabro-test",
  "fabro-types",
+ "fabro-util",
  "fabro-workflow",
  "fabro-workflow-version",
  "git2",
--- a/lib/apps/fabro-server/src/manifest_validation.rs
+++ b/lib/apps/fabro-server/src/manifest_validation.rs
@@ -1,11 +1,14 @@
 use std::collections::HashMap;
+use std::path::PathBuf;
 
-use anyhow::Result;
+use anyhow::{Result, anyhow};
 use fabro_api::types;
-use fabro_config::RunLayer;
+use fabro_config::{RunLayer, WorkflowSettingsBuilder};
+use fabro_manifest::CollectedWorkflowClosure;
+use fabro_workflow::operations::{ValidateInput, WorkflowInput, validate};
 use fabro_workflow::pipeline::TEMPLATE_UNDEFINED_VARIABLE_RULE;
 
-use crate::run_manifest;
+use crate::{run_intent, run_manifest};
 
 /// Validate a manifest without a model catalog.
 ///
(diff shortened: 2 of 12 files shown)


swerner:
## `/simplify` pass — e1c19ac02

Quality-only cleanup (reuse / simplification / efficiency / altitude); no behavior changes intended. fmt, clippy `-D warnings`, `fabro-manifest` and the affected `fabro-server` tests pass.

### Fixed

**Reuse / duplication**
- `canonicalize_location` was implemented twice (`local_workflow_package.rs`, `workflow_version_collector.rs`) → one shared `pub(super)` helper taking an error-mapping closure.
- `resolve_location` re-implemented `resolve_existing_workflow_location` → the latter now returns `fabro_config::Error` directly and both callers reuse it; the collector's `downcast_ref` became a plain `match`.
- `ensure_location_is_within_root` + `LocalWorkflowPackageError::EscapesSourceRoot` duplicated the bundler's `read_package_file` containment check → removed; the test asserts on `Collect { .. }` with "escapes source root" in the chain.
- `PackageFileReadError` (thiserror enum only ever `.into()` anyhow, never matched) → `with_context` / `bail!`.
- Test helpers: `error_chain` → `fabro_util::error::collect_chain`; added `root_version()` replacing four copies of `.closure().versions().last().unwrap().1.version()`; deduped `owner_input()` in `manifest_validation` tests.

**Simplification**
- `GitRunTargetObservation`: five accessors over two fields with no invariant → plain `pub` fields.
- `detect_manifest_repo_info` 3-tuple → named `ManifestRepoInfo` struct.
- `remotely_available.then(|| sha.clone()).flatten()` → `sha.clone().filter(|_| remotely_available)`.
- `WorkflowClosureView`: `root()` is a provided method; `CollectedWorkflowClosureView` stores the root ref at construction, dropping the `.expect(...)` re-lookup.
- Removed unused `ResolvedLocalWorkflowPackage::into_closure`.

**Efficiency**
- HEAD SHA now comes from git2's `head.target()` instead of a separate `git rev-parse` subprocess.
- `validate_collected_workflow` deep-cloned the root `BundledWorkflow` (all file contents) → new `WorkflowBundle::into_workflows()` + `remove()`.

### Skipped (flagged, not applied)
- **Unify `CollectedWorkflowClosure` into `LoadedWorkflowVersionClosure`** would remove the view trait, but callers hold the collected closure by reference and reuse it for upload, so a by-value conversion would force cloning every version's files.
- **`resolve_named_workflow` vs `fabro_config::resolve_workflow_arg_impl`**: the new resolver has deliberately different precedence (markerless worktree first, explicit user root only, no "did you mean"). Worth confirming the divergence from `fabro run` is intended.
- **Reusing `observe_git_context` / `branch_needs_push` being run twice via `sync_status`**: real, but pre-existing shape and touches fabro-workflow's git API.
- **`layer_settings` dedupe in `validate_collected_workflow`**: `NormalizedRun` has ~15 fields; wrapping it here is more ceremony than the duplication.
- **Slug round-trip in `github_run_target`**: `GitRunTarget.repo` is a `String` wire type; changing it is an API decision.
- Unrelated pre-existing failure: `fabro-workflow pipeline::initialize::tests::initialize_prepares_sandbox_and_uses_persisted_run_dir` fails identically on the base branch (environmental LLM credential assertion).

PR #827, opened 2026-08-29
Add RunIntent registration support

swerner:
## Summary

- Adds the dry_run, auto_approve, and preserve_sandbox tri-state RunIntent overrides, preserving omission as inheritance and explicit false as an override.
- Reuses the shared build_run_overrides settings translator during intent admission.
- Adds one-version creation and ordered workflow-version registration methods using canonical workflow-version types.

## Guarantees

- Registration preserves caller order and validates every returned ID against the expected content-derived ID.
- Registration stops at the first mismatch or request failure, retains the underlying API error chain, and performs no rollback of earlier successful registrations.
- Producer migration, workflow-version collection, run creation, and run start remain outside this change.

## Verification

- Focused Rust tests: 22 passed across fabro-types, fabro-api, fabro-server, and fabro-client.
- Pinned-nightly formatting and workspace Clippy passed.
- Documentation checks, TypeScript typechecks, and the release build passed.

## Local full-suite exceptions

- Workspace Rust tests: 8,321 passed and 5 failed in untouched workflow initialization and OpenAI-Codex pull-request generation paths. The initializer failure was reproduced in isolation; the other failures were external LLM Not found responses.
- Web tests: 805 passed and 3 timed out in untouched install-app tests. An isolated rerun shifted which cases failed, indicating existing nondeterminism.

### DIFF
--- a/docs/public/api-reference/fabro-api.yaml
+++ b/docs/public/api-reference/fabro-api.yaml
@@ -9296,6 +9296,21 @@ components:
           type: object
           additionalProperties:
             type: string
+        dry_run:
+          type: boolean
+          description: >-
+            Overrides `run.execution.mode`: true selects `dry_run`, false
+            selects `normal`, and omission inherits the lower-precedence setting.
+        auto_approve:
+          type: boolean
+          description: >-
+            Overrides `run.execution.approval`: true selects `auto`, false
+            selects `prompt`, and omission inherits the lower-precedence setting.
+        preserve_sandbox:
+          type: boolean
+          description: >-
+            Overrides `run.environment.lifecycle.preserve`; omission inherits
+            the lower-precedence setting.
 
     RunTarget:
       description: Workspace content and location requested for a run.
--- a/lib/apps/fabro-server/src/server/handler/runs.rs
+++ b/lib/apps/fabro-server/src/server/handler/runs.rs
@@ -18,10 +18,11 @@ use fabro_api::types::{
     BoardColumn, ManifestConfigType, ManifestGoalType, RunIntent, RunManifest, SubmitAnswerRequest,
     UpdateRunParentRequest, UpdateRunRequest,
 };
-use fabro_config::{CliLayer, ReplaceMap, RunEnvironmentLayer, RunLayer, RunModelLayer, Storage};
+use fabro_config::{CliLayer, RunLayer, Storage};
 use fabro_environment::{DEFAULT_ENVIRONMENT_ID, EnvironmentId};
 use fabro_interview::AnswerSubmission;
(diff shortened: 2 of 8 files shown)


swerner:
## Simplification pass (637d6a0d8)

Reviewed the diff for reuse, simplification, efficiency, and altitude cleanups. Net −77 lines; fmt/clippy clean, all touched tests pass.

### Fixed

**`lib/foundation/fabro-client/src/client.rs`**
- The content-derived id check moved from the batch loop into `create_workflow_version` itself (it computes `version.id()?` and compares the response), so every caller gets the guarantee. `register_workflow_versions` now takes `impl IntoIterator<Item = &WorkflowVersion>` — callers no longer pass an `expected_id` the method can derive, and can no longer pass a mismatched id/body pair that would blame the server.
- Dropped the `entry`/`entries` pluralization in error messages; batch context is now `"failed to register workflow version at index {index}"`.
- Tests: one `mock_create_workflow_version` + `created_workflow_version` helper replaces six hand-spelled 201 mocks (and the single-use `HttpMockResponse` helper); removed the `Arc<AtomicBool>` ordering machinery and `*_response_id` copies — the sequential `.await` loop plus `later_mock.assert_calls_async(0)` already prove ordering; imported `WorkflowPath`/`WorkflowVersion`/`WorkflowVersionId` once instead of ~10 `fabro_types::` paths; the empty-input turbofish became a typed `[&WorkflowVersion; 0]`.

**`lib/apps/fabro-server/src/server/tests.rs`**
- Bare `RunMode::`/`ApprovalMode::` (added `ApprovalMode` to the existing `settings::run` import) replacing six fully-qualified paths; the two 9-line `open_run_reader` chains collapsed to the 2-line idiom neighbouring tests use.

### Skipped (noted, deliberately not done)

- `..RunOverrideInput::default()` in `create_run_from_intent` — all three sibling callers spell every field explicitly; consistency (and compiler-caught drift on new fields) wins.
- `From<&RunIntentArgs> for RunOverrideInput` and the `Option<&str>` → re-allocation nit in `build_run_overrides` — real but belongs with the `ManifestArgs` retirement, not this PR.
- A shared `run_projection` test helper — the `open_run_reader(..).state()` pattern exists at 20+ pre-existing sites; that's a file-wide refactor.
- The `WorkflowVersion` clone inside the `send_api` closure — inherent to `send_api`'s `FnOnce + Clone` retry design, and already leaner than the sibling `body.clone()` pattern.

One heads-up for the producer wiring: `CollectedWorkflowClosure::versions()` yields `(id, &ValidatedWorkflowVersion)`, so the future call site will want `.map(|(_, v)| v.version())` now that the id parameter is gone.

PR #814, opened 2026-08-26
Migrate automations to canonical run targets

swerner:
## Summary

- replace the automation-specific repository/ref target with Fabro's canonical `RunTarget` plus an explicit workflow selector
- migrate legacy file and SQLite selectors transactionally, preserving recoverable mappings and aborting unsupported values with repair guidance
- materialize branches, tags, and exact commits with SHA > tag > branch precedence, then persist the exact checked-out commit on created runs
- update the OpenAPI contract, generated TypeScript client, automation UI, and public migration documentation

## Migration behavior

Legacy 40-hex selectors become exact commits, tag/head selectors become canonical tags or branches, `HEAD` becomes `main`, and other valid bare values become branches. Unsupported `refs/*` values and invalid branch/tag names abort the migration without changing the prior schema or data; the pre-migration snapshot remains available for repair.

## Verification

- `cargo nextest run --workspace --profile ci --no-fail-fast` — 8,279 passed, 208 skipped
- `cargo nextest run -p fabro-automation` — 21 passed
- `cargo nextest run -p fabro-db automation` — 3 passed
- `cargo nextest run -p fabro-server automation` — 18 passed
- `cargo nextest run -p fabro-server git_checkout` — 9 passed
- `cargo nextest run -p fabro-server --features test-support --test it automation` — 29 passed
- `cargo nextest run -p fabro-api` — 220 passed
- `cargo +nightly-2026-04-14 fmt --check --all`
- `cargo +nightly-2026-04-14 clippy --workspace --all-targets -- -D warnings`
- automation web test — 6 passed
- web and generated-client TypeScript checks
- documentation refresh/check and `git diff --check`

## Local environment notes

The full web test command currently reports 17 failures in existing install/build/stale-build/run-files coverage; the changed automation test and web typecheck pass. A local `cargo dev build --release` smoke also stops in `apps/fabro-web/scripts/build.ts` because Bun 1.2.16 returns no `metafile`; this occurs before Rust compilation and is in the same shared web-build path.

PR #828, opened 2026-08-30
Add environment selector to automations

swerner:
## Summary

Automations now select a server-managed Docker or Daytona environment instead of relying on whatever the server happens to resolve at fire time.

- `Automation` gains a required `environment_id` (create/replace) that must reference an enabled, ready Docker or Daytona environment. Local environments are rejected because automation Git targets need a clone-based provider.
- The automation stores only the environment ID. Each trigger fire resolves the current environment definition and snapshots its settings into the new run.
- Deleting an environment still referenced by an automation returns `409 Conflict`.
- Scheduler failures are recorded on the automation as `last_error` and cleared after the next successfully queued scheduled run.
- Web: the create/edit forms require an environment; the list and detail pages show incomplete automations and the latest schedule error.

## Migration

Existing automations are backfilled conservatively (`2026082801_environment_selectors` / `2026082801_automation_environments.sql`): a compatible environment named `default` is selected when present, otherwise the sole compatible environment. Ambiguous cases are left with a null environment; those automations display as incomplete and cannot run until an operator selects an environment in the UI.

## Follow-on to

#814 and #820, which moved automations to canonical run targets and the RunIntent/workflow-version pipeline.

## Test plan

- [x] `cargo build --workspace`
- [x] `cargo nextest run -p fabro-server -p fabro-automation -p fabro-db -p fabro-api -p fabro-manifest` (1163 passed)
- [x] `cargo +nightly-2026-04-14 clippy --workspace --all-targets -- -D warnings`
- [x] `cargo +nightly-2026-04-14 fmt --check --all`
- [x] `bun run typecheck` and `bun test` for automations routes (unrelated pre-existing failures in importChunk/InstallApp/RunFiles also fail on main)

### DIFF
--- a/apps/fabro-web/app/components/automation-form.tsx
+++ b/apps/fabro-web/app/components/automation-form.tsx
@@ -16,6 +16,7 @@ import {
   gitTarget,
   type GitRunTarget,
 } from "../lib/automation";
+import { isCloneBasedEnvironment, providerLabel } from "./environment-form";
 import { Panel, Row } from "./settings-panel";
 import { INPUT_CLASS } from "./ui";
 import { sandboxRuntime } from "../lib/run-sandbox-lifecycle";
@@ -104,8 +105,9 @@ export function automationFormValuesFromRun(
   const cloneBranch = sandboxRuntime(run.sandbox)?.clone_branch;
   const sourceEnvironment = settings?.run?.environment;
   const environmentId = sourceEnvironment
-    && sourceEnvironment.provider !== "local"
-    && environments?.some((environment) => environment.id === sourceEnvironment.id)
+    && environments?.some(
+      (environment) => environment.id === sourceEnvironment.id && isCloneBasedEnvironment(environment),
+    )
       ? sourceEnvironment.id
       : "";
   return {
@@ -186,10 +188,6 @@ function firstPresentString(...values: Array<string | null | undefined>): string
   return "";
 }
 
-function providerLabel(provider: string): string {
-  return provider.charAt(0).toUpperCase() + provider.slice(1);
-}
-
 function githubRepositoryFromSettings(
   settings?: WorkflowSettings | null,
 ): string | null {
@@ -260,7 +258,7 @@ export function AutomationFormFields({
   const slugTouchedRef = useRef(values.id.length > 0);
   const shaValid = isOptionalShaValid(values.sha);
   const compatibleEnvironments = environments
-    .filter((environment) => environment.provider === "docker" || environment.provider === "daytona")
+    .filter(isCloneBasedEnvironment)
     .sort((left, right) => left.id.localeCompare(right.id));
   const selectedEnvironmentMissing = values.environmentId !== ""
     && !compatibleEnvironments.some((environment) => environment.id === values.environmentId);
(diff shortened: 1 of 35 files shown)


swerner:
## /simplify pass (commit 88dae80c6)

Four cleanup lenses (reuse, simplification, efficiency, altitude) reviewed the diff; findings that don't change intended behavior were applied.

### Fixed
- **Web — duplicated provider rule/label:** `automation-form.tsx` had two different "compatible provider" predicates (`!== "local"` vs `=== "docker" || === "daytona"`) and a `providerLabel` copied from `settings-environments.tsx`. Added `isCloneBasedEnvironment` (built on the existing `CREATABLE_PROVIDERS`) and a shared `providerLabel` in `components/environment-form.tsx`; both call sites now import them.
- **Web — repeated query-state expressions:** `automations-new.tsx` recomputed `environmentsQuery.data?.data` / `isLoading && !data` / `Boolean(error)` in each render branch; hoisted to three consts.
- **Model:** `normalize_replace` now trims *and* drops empty `environment_id`, so `validate_fields` needs a single `is_none()` check instead of two overlapping ones.
- **Scheduler:** merged `record_scheduler_error`/`clear_scheduler_error` into one `set_scheduler_error(Option<&str>)`, and only issue the clearing `UPDATE` when the automation actually has a stored `last_error` (saves a write per successful fire).
- **Backfill migration:** added a `SELECT 1 … WHERE environment_id IS NULL LIMIT 1` guard so steady-state startups skip the environments query and the full `store.list()` hydration.
- **Store:** dropped an unnecessary `id.clone()` in the `map_err` closure.
- **Handler:** `EnvironmentId::new(value.to_string())` → `value.parse::<EnvironmentId>()`, matching `runs.rs`.
- Removed a no-op "rebuild this crate…" comment above `sqlx::migrate!` in `fabro-db`.

### Skipped (flagged, not applied — behavior or scope changes)
- Folding fire-time `resolve_automation_environment` into run-intent admission (`select_intent_environment_id` / `validate_intent_environment`). The two lanes overlap and use different error-code vocabularies, but unifying changes which errors are recorded/returned and when. Reasonable follow-up.
- Mapping SQLite FK violations to typed store errors (`EnvironmentStoreError::InUse`, `AutomationStoreError::UnknownEnvironment`) instead of the handler-level `references_environment` / existence pre-checks.
- `require_environment: bool` → a named enum; removing `EnvironmentSelectorBackfillReport` (only tests read it); moving the backfill next to `import_legacy_directory_once` in `serve.rs`; sharing the `insert_environment` test fixture across crates via `test-support`.

### Verification
`cargo fmt --check`, clippy `-D warnings` (fabro-server/fabro-automation/fabro-db), fabro-automation + fabro-db + scheduler unit tests, and server `it` automation/environment tests all pass; `bun run typecheck` passes. `bun test` has 16 pre-existing failures (InstallApp/importChunk/RunFiles) identical on the clean tree.

swerner:
Follow-up 7c261f2ae: the CI bun test failure came from the simplify pass — importing the provider helpers from components/environment-form.tsx pulled headlessui Disclosure into automation-form.tsx, which automations-new.test.tsx mocks without it. Moved CREATABLE_PROVIDERS and the two helpers into app/lib/environment-providers.ts (no component imports).

PR #829, opened 2026-08-31
Index session ownership from creation events

swerner:
## Summary

- Make the typed run.session.created event the sole durable session ownership fact and enforce one owner per session with a unique partial SQLite index.
- Add a count-only migration preflight, resolve flat session routes through a fully decoded and validated indexed event, and remove normal SlateDB reverse-key writes, reads, and delete scans.
- Reject run.session.created through generic event append so callers use the dedicated session operation; legacy reverse rows remain diagnostic-only.

## Testing

- Focused fabro-db, fabro-store, and fabro-server session checks
- Full workspace nextest sweep: 8,323 tests passed in the main post-rebase run; the two load-sensitive timeout cases passed separately with one test thread and the repository CI timeout profile, covering all 8,325 cases

## Deferred

- Cross-run list, scheduler, and pending pull-request query changes remain deferred.
- Projection-cache and cache-warmup work remains deferred.

### DIFF
--- a/lib/apps/fabro-server/src/server/handler/events.rs
+++ b/lib/apps/fabro-server/src/server/handler/events.rs
@@ -218,9 +218,9 @@ async fn append_run_event(
     if event.run_id != id {
         return ApiError::bad_request("Event run_id does not match path run ID.").into_response();
     }
-    if let Some(denied) = denied_lifecycle_event_name(&event.body) {
+    if let Some(denied) = denied_dedicated_operation_event_name(&event.body) {
         return ApiError::bad_request(format!(
-            "{denied} is a lifecycle event; clients must call the corresponding operation endpoint instead of injecting it via append_run_event"
+            "{denied} must be performed through its dedicated operation endpoint instead of injecting it via append_run_event"
         ))
         .into_response();
     }
@@ -576,7 +576,7 @@ async fn attach_run_events(
 /// (e.g. "archive only from terminal") that a direct event append would
 /// bypass. Other run-lifecycle events flow through this endpoint legitimately:
 /// the worker subprocess emits state transitions during execution.
-fn denied_lifecycle_event_name(body: &EventBody) -> Option<&str> {
+fn denied_dedicated_operation_event_name(body: &EventBody) -> Option<&str> {
     match body {
         EventBody::RunArchived(_)
         | EventBody::RunUnarchived(_)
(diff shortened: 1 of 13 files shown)


swerner:
## Simplification review (bf05b01)

Four parallel cleanup reviews (reuse, simplification, efficiency, altitude) over this PR's diff. Applied fixes and noted skips below.

### Fixed

1. **Dead re-validation in `find_session_owner`** (`run_summary_store.rs`) — the three post-decode checks (`run_id`, `session_id`, event body) were unreachable: `decode_event_row` already verifies every stored column against the decoded envelope, and the `WHERE` clause pins `session_id` and `event_name` to the requested values. The corrupt-row tests are caught inside the decode layer with identical error fields, and still pass with the checks deleted.
2. **Duplicated SELECT column list** — the inline seven-column query was a verbatim copy of the existing `SELECT_EVENT_COLUMNS` constant; rebuilt with `QueryBuilder::new(SELECT_EVENT_COLUMNS)` like the six sibling event queries.
3. **`"<invalid>"` sentinel error** — the unparseable-run_id error now carries the actual stored value, matching how sibling queries report corrupt ids.
4. **Double `applied_migration_versions` fetch per startup** — `migrate()` fetched the applied-migration set in the new preflight and again in `snapshot_before_new_migrations` (each an acquire + ensure-table + list query). Now fetched once and shared.
5. **Preflight check ordering** — the applied-version check (zero I/O with the shared set) runs first, so steady-state startups skip the `sqlite_master` probe; the expensive `GROUP BY` scan was already correctly gated. Also documented the preflight as temporary compatibility code, removable with the run-history compatibility window.
6. **Hand-rolled legacy session keys in three tests** — restored `session_by_id_key` as a `#[cfg(test)]` helper (same pattern as `run_catalog_key`) so the legacy key shape can't drift from the production `sessions_by_id_prefix()` scan.

### Skipped

- Generalizing the migration preflight into a declarative version→check table — over-engineering for a single temporary guard; the removal note covers it.
- Exporting one ordered migration-SQL slice so test fixtures stop hand-listing migration constants — real fragility, but the pattern predates this PR; separate refactor.
- A shared constant for the `'run.session.created'` literal — inline event-name literals are the established convention in this store and in existing migrations.
- Test-fixture dedup across modules (`session_created_payload` et al.) — per-module event fixtures are the crate's existing convention.

One heads-up rather than a fix: the migration version now lives in three hand-synced copies (the Rust constant, the SQL filename, and a literal in the fabro-db test). If this migration gets renumbered before merge, all three must move together.

Verified: `cargo nextest run -p fabro-store -p fabro-db` (309 tests pass), fmt and clippy clean.

PR #830, opened 2026-08-31
Move cross-run summaries and pull request recovery to SQLite

swerner:
## What changed

- move complete and status-filtered cross-run summary reads to focused SQLite queries with deterministic ordering and no hidden result cap
- rebuild canonical run summaries at the SQL transition boundary while keeping full run projections cached for existing projection consumers
- recover pull-request creation from indexed request-event candidates, replay only candidate runs, and process a bounded, deduplicated, oldest-first queue
- enqueue newly committed pull-request requests directly so the normal path does not wait for periodic recovery
- remove the cache-backed current-summary and pending-request scan responsibilities

The existing request-event index supports the recovery query, so this change adds no SQLite migration.

## Testing

- focused `fabro-store` run-summary tests
- focused `fabro-db` request-event query-plan test
- focused `fabro-server` pull-request recovery and parent-link tests
- focused `fabro-workflow` run-lookup test
- pinned nightly rustfmt check across the workspace
- pinned nightly clippy across all workspace targets with warnings denied
- `cargo build --workspace`
- full workspace nextest run with provider credentials removed: 8,323 passed, 208 skipped; one load-sensitive test passed on its second attempt

## Intentionally deferred

- session reverse-index and session lookup work remains separate
- global projection-cache removal and lazy replay of inactive runs remain separate

### DIFF
--- a/lib/apps/fabro-server/src/run_files.rs
+++ b/lib/apps/fabro-server/src/run_files.rs
@@ -1195,7 +1195,7 @@ async fn load_projection(
     state: &Arc<AppState>,
     run_id: &RunId,
 ) -> std::result::Result<Arc<fabro_store::RunProjection>, ApiError> {
-    Ok(state.cached_run(run_id).await?.projection)
+    state.cached_run_projection(run_id).await
 }
 
 async fn reconnect_run_sandbox(
--- a/lib/apps/fabro-server/src/server.rs
+++ b/lib/apps/fabro-server/src/server.rs
@@ -2638,7 +2638,7 @@ async fn delete_run_internal(
     };
     let had_managed_run = managed_run.is_some();
     let durable_status = if managed_run.is_some() {
-        load_durable_run_status(state, &id).await
+        durable_run_status(state, id).await.ok().flatten()
     } else {
         None
     };
@@ -2703,16 +2703,6 @@ async fn delete_run_internal(
         SandboxDeleteOutcome::Absent if had_managed_run => Ok(DeleteRunOutcome::Deleted),
         SandboxDeleteOutcome::Absent => Ok(DeleteRunOutcome::AlreadyAbsent),
     }
-}
-
-async fn load_durable_run_status(state: &AppState, id: &RunId) -> Option<RunStatus> {
-    state
-        .stores
-        .run_summaries
-        .get(id, Utc::now())
-        .await
-        .ok()?
-        .map(|summary| summary.lifecycle.status)
 }
 
 async fn delete_run_sandbox_resource(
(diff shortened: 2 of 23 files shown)


swerner:
Ran a cleanup review (reuse / simplification / efficiency / altitude) over this PR and pushed the resulting fixes in 7cda4dcc.

## Fixed

1. **Recovery no longer replays full event histories** (`pull_request_supervisor.rs`). `recover_pending_pull_request_creations` called `open_run_reader().state()` per candidate — a per-run SQL existence query plus a deep clone of the entire projection, on every 30s scan, for every run that ever requested a PR. It now uses `get_cached_projection()`, the same warm-cache accessor `process_pull_request_creation` in the same file already uses. This also collapsed the nested two-arm error match into one.
2. **Dispatch-eligibility predicate deduplicated and hoisted**. The `active`/`failures` check was written out verbatim in both the recovery scan and the pop loop, and in the scan it ran *after* the expensive projection load. It's now one `can_dispatch()` helper, checked before any I/O.
3. **`load_durable_run_status` deleted** (`server.rs`). The migration made it byte-identical to the existing `durable_run_status`; the single call site now uses `durable_run_status(...).ok().flatten()`.
4. **Run-id parse deduplicated** (`run_summary_store.rs`). The new candidate query added a third copy of the stored-id → `RunId` conversion; all three sites now share `parse_stored_run_id()`.
5. **Canonical ordering reused**. `list_all` and `list_by_statuses` hardcoded `ORDER BY created_at_ms DESC, id DESC` — exactly what the existing `push_order()` produces — so they now call it, keeping the tie-break rule in one place.
6. **Redundant test accessor dropped**. `clear_pull_request_creation_queue` duplicated `drain_pull_request_creation_queue`; its one caller now drains.

Verified with `cargo build`, full `fabro-store` (271) and `fabro-server` (877) suites, `fmt`, and `clippy -D warnings` — all green.

## Skipped (noted for follow-up)

- **Candidate set grows forever.** `list_pull_request_creation_candidate_run_ids` returns every run that *ever* recorded a `pull_request.creation_requested` event — no "still pending" filter — so scan work is O(all runs that ever created a PR), not O(currently pending), and the in-memory queue exists partly to compensate. The deeper fix is persisting a `pull_request_creation_pending` flag (+ `requested_at_ms`) on the `runs` row the way `status` already is, making the scan one indexed query; the queue, its AppState accessors, and the notify-vs-scan split could then likely go away. That's a schema migration and a redesign of this PR's core mechanism, so it was left as-is; fix 1 reduces the steady-state per-candidate cost to a cache hash lookup in the meantime.
- **Queue's `run_ids: HashSet`** was flagged as derivable state, but removing it would weaken dedup from by-run-id to by-(timestamp, run-id) — a subtle behavior change not worth making silently. Same for making the queue capacity-evicting instead of refuse-when-full.
- **`list_all` computes a per-row `children_count` correlated subquery no caller uses**, but a separate SELECT/decode path would either duplicate SQL or produce rows with a wrong `children_count`, so it stays.
- Minor: `list_by_statuses` full-scans `runs` (the only status index leads with `archived_at_ms`), but it runs once at startup — only worth an index if it gains hot callers.

PR #831, opened 2026-08-31
Create CLI runs from immutable workflow intents

swerner:
## Summary

- Resolve and package the local workflow closure, register immutable workflow versions dependency-first, and submit one canonical run intent for the root version.
- Make full run validation server-authoritative so admission uses the selected environment, configured providers, MCP catalog, defaults, secrets, and stored workflow closure. The CLI still rejects unreadable or structurally unpackageable local sources before submission.
- Keep configured and effective sandbox providers distinct. Target compatibility uses the configured provider, while dry-run execution may use effective Local.
- Give dry-run Git and empty clone targets a canonical run-scoped scratch workspace without changing valid Local folder behavior.
- Require inferred clone targets to prove the exact local commit is available from the canonical GitHub origin. A failed proof stops before version registration or run creation and does not expose remote URLs or raw Git output. A non-repository caller still submits an empty target.
- Preserve typed scalar conversion failures in the error chain while adding the input key as CLI context.
- Keep create as submission-only; run still performs a separate start and attach flow.

This PR is now standalone on main. It changes no OpenAPI schema and regenerates no API clients.

## Verification

- cargo nextest run -p fabro-cli --test it create
- cargo nextest run -p fabro-cli commands::run::overrides
- cargo nextest run -p fabro-cli --test it dry_run_simple
- cargo nextest run -p fabro-cli --test it dry_run_with_goal_file_reads_contents_into_goal
- cargo nextest run -p fabro-cli --test it cmd::run
- cargo nextest run -p fabro-cli --test it workflow::dry_run_examples
- cargo nextest run -p fabro-cli --test it
- cargo nextest run -p fabro-server run_manifest
- cargo nextest run -p fabro-workflow dry_run
- cargo +nightly-2026-04-14 clippy -p fabro-cli -p fabro-server -p fabro-workflow --all-targets -- -D warnings
- cargo --locked dev docs check
- cargo insta pending-snapshots reports no pending snapshots

A full workspace Nextest build was also attempted locally, but the linker exhausted the host volume before tests started. CI remains the workspace-wide verification authority.

## Deliberate boundaries

- The downstream run-create tool producer remains separate. PR #832 must rebase onto this branch after it lands, remove its duplicate configured-provider and dry-run workspace changes, and align its own producer with server-authoritative admission.
- Remote workflow-source UX, a shared-filesystem connection guard, and an explicit empty-workspace CLI flag remain future work.
- Project configuration compatibility is not expanded; existing warnings remain.
- OpenAPI, generated clients, and the run-tool producer are unchanged here.

### DIFF
--- a/docs/public/changelog/2026-08-31.mdx
+++ b/docs/public/changelog/2026-08-31.mdx
@@ -15,7 +15,11 @@ Workflows can still be selected by project name, from user workflow storage,
 from another local checkout, or as loose local files. The workflow source is
 independent from the execution target: local environments use the directory
 where Fabro was invoked, while Docker and Daytona derive a GitHub target from
-that directory. Remote Git workflow acquisition is not included.
+that directory. Clone-based runs invoked from a Git checkout require a
+non-bare checkout with an attached branch and at least one commit; detached
+HEAD and unborn checkouts fail before workflow registration instead of being
+treated as empty workspaces. A directory outside Git still selects an empty
+workspace. Remote Git workflow acquisition is not included.
 
 CLI input overrides retain string, Boolean, integer, and finite-float types,
 and Boolean flags remain sparse. `--goal-file` is read locally and sent by
--- a/docs/public/changelog/2026-08-31.mdx
+++ b/docs/public/changelog/2026-08-31.mdx
@@ -1,0 +1,31 @@
+---
+title: "Immutable workflow versions for CLI runs"
+date: "2026-08-31"
+---
+
+`fabro run` and `fabro create` now validate local workflows, register their
+immutable workflow versions and dependencies, and create runs by intent.
+`fabro create` leaves the run submitted; `fabro run` starts it with a separate
+request.
+
+Workflows can still be selected by project name, from user workflow storage,
+from another local checkout, or as loose local files. The workflow source is
+independent from the execution target: local environments use the directory
+where Fabro was invoked, while Docker and Daytona derive a GitHub target from
+that directory. Remote Git workflow acquisition is not included.
+
+CLI input overrides retain string, Boolean, integer, and finite-float types,
+and Boolean flags remain sparse. `--goal-file` is read locally and sent by
+value as a per-run override; goal files referenced by `workflow.toml` remain
+part of the immutable workflow.
+
+Project and CLI-machine `[run]` and `[environments]` settings are no longer
+transmitted during intent creation. Fabro warns with only the affected file and
+key names. Move workflow behavior—including `[run.pull_request]`—to
+`workflow.toml`, and configure placement with server-managed environments.
+Malformed or unreadable active configuration still fails before creation.
+
+New repositories now place their default automatic pull-request policy in the
+starter workflow rather than `.fabro/project.toml`. Intent-created runs also
+derive their display slug from the immutable workflow entrypoint. These
+changes require no API schema update.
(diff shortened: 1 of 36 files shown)


swerner:
Ran a quality pass (reuse / simplification / efficiency / altitude) over this PR and pushed the cleanups as cf1a4141d. No intended behavior changes.

## Fixed

- **Dead code removed** — `run_manifest_args` lost its last production caller in this PR and was kept alive by `#[cfg(test)]` plus a `main.rs` test pinning behavior nothing executes. Deleted all three; `preflight_manifest_args` is the remaining real builder.
- **TOML→JSON scalar conversion moved to the chokepoint** — the hand-rolled conversion in `overrides.rs` is now `toml_scalar_to_json_value` in `fabro-types/src/input_scalar.rs`, next to its inverse `json_scalar_to_toml_value` (which the server uses on receipt), with typed errors and round-trip tests.
- **Goal resolution deduplicated** — `prepare_intent_overrides` re-implemented the `--goal`/`--goal-file` exclusivity check and cwd anchoring already in `goal_layer_from_args` in the same file; it now calls that helper and just reads the resolved file.
- **`"default"` literal** → `fabro_environment::DEFAULT_ENVIRONMENT_ID`, the same constant the server uses for the identical defaulting.
- **`run_target_for_environment` is now pure** — takes `(provider, cwd)` and returns `(RunTarget, dirty)` instead of threading `ctx`/`styles` in to print one warning; provider classification uses the established `is_clone_based()` predicate; the warning prints at the call site.
- **Parent resolution and environment retrieval run concurrently** (`tokio::try_join!`) — they were independent, serialized round trips.
- **Stale doc comment** on `create_run` (still described "allocate run dir, persist spec") rewritten; `.map_err(anyhow::Error::new)` replaced with a bare `?`.
- **`ResolvedCommandSettings` deleted** — a strict field-subset copy of `LoadedSettings` that cost six edit sites per new field; constructors use `LoadedSettings` directly.
- **Duplicated parse-error mapping** in `read_project_run_settings_key_presence` hoisted into one closure.
- **Test mocks consolidated** — `environment_json`, `mock_environment`, `mock_workflow_version_registrations` (plus a recording variant replacing two inline re-implementations), and `run_git` now live in `tests/it/cmd/support.rs`; `create.rs`, `run.rs`, and `mcp.rs` import them.

## Skipped (flagged, not applied)

- **Unifying the CLI intent pipeline with the server's `ServerRunCreateAdapter::prepare`/`resolve_target`**, and a shared `RunIntentArgs` → `RunLayer` mapping. Real altitude findings, but cross-crate redesigns beyond this PR. Worth a follow-up: the CLI's target derivation already silently drops the server-side "HEAD not fetchable" warning, and the CLI's preview `RunLayer` differs from the layer the server builds at admission (environment id, goal wiring).
- **Probe-before-upload for workflow version registration** and **`spawn_blocking` around the git observation** — the upload-every-time and validate-target-before-upload sequencing look deliberate, and the CLI has nothing else running concurrently at that point.
- **The double TOML parse** in `read_project_run_settings_key_presence` — the `SettingsLayer` parse carries version/unknown-key validation and the raw-document parse is required to detect empty `[run]`/`[environments]` tables; removing either needs a fabro-config API change (e.g. exposing the project-file loader or a Value-based parse).
- **`fabro workflow create` scaffolding** doesn't get the `[run.pull_request]` starter block that `repo init` now writes into `workflow.toml`, so the advertised default depends on which command scaffolded the workflow. Left alone as a product decision, not a mechanical cleanup.

## Verification

`cargo build --workspace`, `cargo nextest run -p fabro-cli` (995 passed) and `-p fabro-types` (496 passed), pinned-nightly `cargo fmt`, and `cargo clippy --all-targets -- -D warnings` on both changed crates all pass.

PR #832, opened 2026-08-31
Migrate fabro_run_create to the RunIntent API

swerner:
Run creation now submits a registered `workflow_version_id`, an independent `target`, and canonical RunIntent settings to the runs API. Workflow acquisition and registration are provided by `fabro_workflow_version_create` from merged #851. This PR is based on `main` and uses Pebble’s native application-tool dispatch.

## Behavior

- Remove workflow selectors, inline sources, `goal_file`, cwd/config handling, and the server run-create adapter. Callers register contents first, then reuse the returned ID.
- Require an explicit target for standalone MCP. Native agents remain parented to their current run and can inherit its fresh execution target, including the actual run branch without the parent's original SHA/tag.
- Pass nested canonical `args` and `environment_id` directly, preserving omitted overrides and explicit `false` values. Keep existing CLI acquisition and target-selection semantics.
- Keep create and optional start separate. Sequential batches stop at the first failure and report already-created IDs, including when start or summary retrieval fails.
- Enforce folder-target restrictions in server admission so Docker/Daytona workers cannot select server-host folders.
- Update MCP and child-run docs with the two-call flow and migration guidance. Old run-create source forms are intentionally rejected.

## Validation

- Live Daytona smoke passed using the PR binary and an isolated temporary server: a Pebble parent agent invoked native `fabro_workflow_version_create` and `fabro_run_create`, omitting target and parent ID. The child inherited the parent’s actual run branch, passed the expected user-approval gate, cloned that branch in Daytona, and completed its command stage. Both runs succeeded. Test sandboxes, the temporary Git branch, and temporary server credentials were cleaned up.

- 203 focused tests passed across tool, MCP, native Pebble dispatch, workflow registration, run admission, CLI creation, and Git observation. Includes the current-parent dispatch regression and proof that a child checkout contains its parent's pushed work.
- One CLI test emitted a process-cleanup warning in the full run; its isolated rerun passed without the warning.
- `cargo +nightly-2026-04-14 clippy -p fabro-tool -p fabro-mcp-server -p fabro-workflow -p fabro-server -p fabro-cli -p fabro-manifest --all-targets -- -D warnings` passed.
- `cargo +nightly-2026-04-14 fmt --check --all` and diff whitespace checks passed. No pending snapshots.

The full workspace test suite was not run locally; focused tests and the live Daytona smoke cover the changed contracts.


swerner:
## Cleanup review (`/simplify`)

Ran a four-angle cleanup review (reuse, simplification, efficiency, altitude) over this PR and applied the safe, behavior-preserving fixes in 59dc532. `cargo fmt --check`, clippy `-D warnings`, and the full test suites for the five affected crates pass; web typecheck and tests pass.

### Applied

- **`run_tool_create.rs`** — removed capability guards duplicated between `prepare()` and its helpers (the helpers keep them; they already gate before any read); collapsed the `Stored` early-return + `unreachable!()` arm into one match; deleted the `LocalWorkflowSource` enum and its phantom `TempDir` field (the collected closure owns every file, exposed via a new `ResolvedLocalWorkflowPackage::into_closure()`); moved blocking git observation and workflow-package collection onto `spawn_blocking` so `prepare` no longer pins a tokio worker through a `git push`/`ls-remote`.
- **`fabro-tool/create.rs`** — merged `ValidatedCreateRunWorkflowSource` (a byte-for-byte copy) into `CreateRunWorkflowSource`; dropped the `MAX_INLINE_WORKFLOW_*` alias constants and derived the "512 KiB"/"2 MiB" error strings from the `fabro_types` limits; added a schema-parity test so a field added to `RunTarget` without updating the hand-written `json_schema!` literal fails CI.
- **`fabro-store`** — deleted the retired `reconcile` write path (demoted to `#[cfg(test)]` with only self-referential tests) and reworked the slate test to assert the event-path summary directly; removed the `list_events_with_json_in_transaction` deref shim; shared the fail-closed activation-marker query with `delete_canonical`; dropped the redundant `contains()` pre-checks in `open_run`/`open_run_reader` (`RunDatabase::build` already surfaces `RunNotFound`).
- **Migrations** — moved the WAL `wal_checkpoint(TRUNCATE)` busy-handling into `sqlite_activation_backup` so the blob and run-history bridges share one mechanism, as that module's doc comment promises.
- **Misc** — dropped the always-`None` `sha` parameter from `github_run_target`; deleted the `worker_run_create_adapter` pass-through in the CLI runner; extracted `runBlockedReason()` into `lib/automation.ts` so the automation run-button rules and messages live once instead of in both the list row and detail header.

### Bonus correctness fix

The suite exposed a pre-existing flake on this branch (fails ~2 of 3 full `fabro-server` runs): inline workflow files staged with tokio `File::write_all` were sometimes read back **empty**, because dropping a tokio `File` doesn't wait for queued writes and the collector reads synchronously. Added a `flush().await`; four consecutive full-suite runs are now clean.

### Skipped (flagged by review, worth separate consideration)

- **Startup cost**: the run-history activation bridge scans the legacy source three times and replays/verifies every SQLite run twice on *every* boot — O(all events ever) ×4 until the bridge is removed. Restructuring the fail-closed import/verify design felt beyond a cleanup pass.
- **`active_runs` mutex held across SQLite commits** on the create/delete paths, serializing all run opens behind each other's transactions — a locking redesign, not a cleanup.
- **Producer-side validation resolves environments against the seeded catalog**, so it will falsely reject server-defined environments that admission would accept — behavior decision.
- **Selector grammar** diverges between `collect_workflow_versions` (any relative extensionless path) and `resolve_local_workflow_package` (bare names only) in the same crate.
- **Provider/target compatibility matrix** is restated at three layers (intent admission, automation environment resolution, worker start) with three phrasings — a shared predicate next to the types would help, but changes error strings across layers.
- **`user_workflows_root`** is derived as `settings_path.parent()/workflows` at two wiring sites instead of `Home::workflows_dir()`; these diverge when `FABRO_CONFIG` points outside the Fabro home.
- **`WorkflowClosureView` trait** in `run_intent.rs` was flagged as adapter weight, but the proposed single-type fix would force a deep clone of the whole closure (the caller still needs it for registration) — the borrowing view earns its keep.
- The per-retry `WorkflowVersion` clone in `fabro-client` is forced by the generated builder taking an owned body, and the id recompute is a deliberate fail-loudly verification — left alone.

PR #833, opened 2026-09-01
Load inactive run projections on demand

swerner:
## Summary

- stop replaying and retaining every run projection during server startup
- keep coordinated in-memory projection state for opened runs, while replaying inactive run details on demand from one validated SQLite snapshot
- preserve indexed pull-request recovery by querying candidate run IDs first and lazily loading only the candidate histories needed for confirmation
- preserve existing API and event wire contracts, missing-run behavior, and structured replay errors

## Testing

- `cargo nextest run -p fabro-store -E 'test(projection) | test(reader) | test(replay) | test(unreadable)'`
- `cargo nextest run -p fabro-server --features test-support -E 'test(startup) | test(run_state) | test(pull_request)'`
- `cargo build --workspace`
- `cargo nextest run --workspace --no-fail-fast` (8,377 passed with ambient provider credentials removed from the unit-test environment)

## Deferred work

This intentionally does not add durable projection snapshots, redesign workers, evict active run handles, revisit session or cross-run query design, or perform the final SlateDB compatibility cleanup. Those remain separate follow-up work informed by post-cutover measurements and migration timing.


swerner:
## Simplify pass (Claude Fable 5.1)

Pushed as 13f3ff867. Four cleanup reviews (reuse, simplification, efficiency, altitude) over the diff, then applied what held up. `cargo clippy --all-targets -D warnings` and `cargo fmt --check` are clean; fabro-store (277) and fabro-server with `test-support` (1123) pass.

### Fixed

- **Removed the unreachable `matches_run` guard** from `load_run_projection`, `open_run`, and `open_run_reader`, and deleted the method. `active_runs` is only ever inserted under the handle's own run ID, so the check could never fire. This PR had added a third copy of it.
- **Trimmed `install_in_memory_state`** to take only the committed projection. The `event` parameter duplicated `last_seq`, and the inner scope block existed only to release the lock before the now-deleted shared cache update.
- **Added `From<CachedRunProjection> for EventProjectionCache`** and used it in `RunDatabase::build` instead of a hand-built struct literal. Added a doc comment on the moved struct.
- **Renamed `projected_state_locked` to `projection_snapshot_locked`** so the lock-taking method and its lock-held sibling share a stem like the rest of the file.
- **Fixed the stale comment** on `EventProjectionCache::state` that still referred to the deleted shared projection cache.
- **Cheaper existence and archived checks.** `reject_if_archived` and `ensure_run_exists` replayed the full event history for inactive runs just to read one column. Both now read the run summary row, which is written in the same SQLite transaction as the event. `reject_if_archived` guards roughly twenty mutation handlers, most of which then `open_run` and replay again, so this was a double replay per request on inactive runs.
- **Deduplicated test fixtures.** `reopen_store` in the fabro-store test module replaces five identical reopen blocks; `test_app_state_over_shared_stores` in the server tests replaces two ~30-line blocks.

### Skipped (noted, not changed)

- **Renaming `CachedRunProjection`.** Nothing caches it anymore, so the name misleads, but it has ~25 uses across `run_summary_store.rs` and `legacy_run_history_import.rs`. Follow-up.
- **Moving `build_projection` onto `RunSummaryStore`.** It is a SQLite query that never touches `RunDatabase` state, but the placement predates this PR.
- **Collapsing `load_run_projection` into `open_run_reader`.** With the dead guard gone each is three lines, and merging would add a `contains` round trip plus a `RunDatabase` allocation on the inactive path.
- **Pull-request recovery replay cost.** `list_pull_request_creation_candidate_run_ids` returns every run that ever requested a PR, and each inactive candidate is now fully replayed every 30s scan. Under the old warm cache this was O(1) per candidate. Narrowing the candidate SQL to runs whose latest `pull_request.*` event is `creation_requested`, or prefiltering on the summary's `pull_request` link, is a design change worth its own PR.
- **Two test nits.** `inactive_projection_loads_on_demand_without_registering_run` creates three runs and asserts on one; `inactive_projection_read_concurrent_with_append_is_one_coherent_snapshot` cannot fail on the property it names since replay folds a `Vec` and can only yield one of two titles. Left for the author.

swerner:
## Follow-up: bounded pull request recovery scan (064074233)

Addresses the one remaining performance concern from the simplify pass.

**Problem.** The supervisor wakes every 30 seconds and asked SQLite for every run that had *ever* recorded a `pull_request.creation_requested` event, then replayed each inactive candidate's full history just to learn the request was no longer pending. Under the old warm cache that was a hash lookup per candidate; with on-demand loading it became a full replay per candidate per scan, over a set that only grows.

**Change.** `list_pull_request_creation_candidate_run_ids` now returns only runs whose latest creation request has no later resolving event, mirroring the projection reducer exactly:

- a later `creation_requested` supersedes the earlier one
- `created`, `linked`, and `unlinked` resolve any pending request
- `failed` resolves only when its `creation_id` matches the request (a publish-stage failure with no id does not)

The supervisor is unchanged and still replays each candidate to confirm, so the query only has to be a superset of the truly pending set. The replayed set is now bounded by in-flight requests instead of by history. The outer query still uses the existing partial index on `creation_requested` events; the `NOT EXISTS` walks the primary key for events after the request.

**Test.** Added `pull_request_creation_candidates_exclude_requests_resolved_by_later_events`, which seeds eight runs covering every resolving and non-resolving case above and asserts exactly the three still-pending ones come back. The existing indexed-scan recovery test in fabro-server passes unchanged.

PR #835, opened 2026-09-02
Move run projection replay into the summary store

swerner:
## Summary

- Rename the private `CachedRunProjection` value to `ProjectedRun`, representing a replayed or newly committed projection pinned to its last event sequence.
- Move canonical SQLite history replay to `RunSummaryStore::load_projection`, which owns the rows being read.
- Remove redundant run-existence queries from inactive writer and reader opens; replay already returns `RunNotFound` for missing runs.
- Keep replay frequency, the per-handle `EventProjectionCache`, public APIs, errors, and wire behavior unchanged.

## Testing

- `cargo nextest run -p fabro-store` (281 passed)
- `cargo nextest run -p fabro-server --features test-support -E` with four exact projection/restart/archive/unreadable regressions (4 passed)
- `cargo +nightly-2026-04-14 clippy -p fabro-store --all-targets -- -D warnings`

### DIFF
--- a/lib/components/fabro-store/src/legacy_run_history_import.rs
+++ b/lib/components/fabro-store/src/legacy_run_history_import.rs
@@ -16,7 +16,7 @@ use tokio::sync::Barrier;
 use tracing::debug;
 
 use crate::keys::SlateKey;
-use crate::slate::CachedRunProjection;
+use crate::slate::ProjectedRun;
 use crate::{Database, EventPayload, RunProjectionReducer, RunSummaryStore, keys};
 
 /// Count-only observations about the legacy catalog and session indexes.
@@ -458,7 +458,7 @@ struct ValidatedLegacyRunEvent {
 struct ValidatedLegacyRunHistory {
     run_id:  RunId,
     events:  Vec<ValidatedLegacyRunEvent>,
-    current: CachedRunProjection,
+    current: ProjectedRun,
 }
 
 struct LegacyRunHistorySource {
@@ -534,7 +534,7 @@ impl LegacyRunHistorySource {
             .expect("a validated history contains at least one event")
             .envelope
             .seq;
-        let current = CachedRunProjection::from_projection(run_id, projection, last_seq);
+        let current = ProjectedRun::new(run_id, projection, last_seq);
         Ok(Some(ValidatedLegacyRunHistory {
             run_id,
             events,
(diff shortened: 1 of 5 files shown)


swerner:
## `/simplify` pass (commit 0f1e5e1c9)

Four cleanup lenses (reuse, simplification, efficiency, altitude) run over this PR's diff. Clippy, rustfmt, and all 281 fabro-store tests pass; the workspace still checks.

### Fixed

- **`ProjectedRun` moved to `run_state.rs`** next to `EventProjectionCache`. It lived in the `RunDatabase` handle module while `RunSummaryStore` both produced and consumed it, forcing an upward import plus a re-export from `slate`. Doc comment now names the real producers and consumers.
- **Added `ProjectedRun::replay(run_id, events)`** and pointed five sites at it: `load_projection`, both replay paths in the legacy importer, and three test helpers. Each carried its own apply-events / take-last-seq / wrap sequence with three different empty-history error messages. `RunProjectionReducer::apply_events` already rejects an empty slice, so those guards were unreachable.
- **`load_projection` returns `Result<ProjectedRun>`** and lets `RunNotFound` flow. It previously erased the error to `None` and `RunDatabase::build` immediately rebuilt the same error. `load_run_projection` is now the single `Option` translation point. `list_unreadable_runs` lost its `Ok(None)` arm, whose "run has no events" label was misleading: that condition actually surfaces as `RunHeadMismatch`.
- **One cache mapping instead of two.** `install_in_memory_state` takes the value and assigns `projected.into()`, reusing the existing `From<ProjectedRun> for EventProjectionCache` rather than setting fields by hand.
- **`ProjectedRun::new` takes an `Arc<RunProjection>`**, so `commit_event_locked` no longer does `Arc::unwrap_or_clone` only for the constructor to `Arc::new` again on every event.
- **Trimmed `load_projection_reports_removed_events`.** The "without poisoning following reads" half asserted something a stateless store method cannot violate; `inactive_projection_replay_accepts_sequence_gaps_and_failures_do_not_poison_reads` already covers the real cache-poisoning scenario.

### Skipped

- Extracting a shared begin / `insert_first_event_on_connection` / commit test helper. The pattern predates this PR and appears in eight other tests in the module.
- Acquiring one connection for the whole `list_unreadable_runs` loop, and skipping `event_json` decoding during replay. Both pre-existing, diagnostics-path only.
- Adding `PartialEq` to `RunProjection` to replace the serde round-trip assertion. That change belongs in `fabro-types`.

PR #836, opened 2026-09-02
Select the sole managed run environment when default is absent

swerner:
## Summary

- preserve explicit environment selection and the existing `default` convention
- when `default` is absent, select the sole Docker or Daytona environment for both CLI and direct API run creation
- reject zero or multiple compatible environments with `environment_selection_required` instead of choosing arbitrarily
- document the fallback and refresh the generated TypeScript client comment

This matches the conservative selection rule used by the existing automation environment migration. The synthetic `local` environment does not count as a Docker/Daytona fallback candidate.

## Testing

- `cargo nextest run -p fabro-environment -p fabro-client -p fabro-cli -p fabro-server` (2,190 passed; one existing leaky-test warning)
- `cargo +nightly-2026-04-14 clippy -p fabro-environment -p fabro-client -p fabro-cli -p fabro-server --all-targets -- -D warnings`
- `bun run typecheck` in `lib/packages/fabro-api-client`

### DIFF
--- a/docs/public/api-reference/fabro-api.yaml
+++ b/docs/public/api-reference/fabro-api.yaml
@@ -9377,7 +9377,11 @@ components:
           $ref: "#/components/schemas/RunIntentArgs"
         environment_id:
           type: string
-          description: Server environment catalog ID. Omission selects `default`.
+          description: >-
+            Server environment catalog ID. When omitted, the server selects an
+            environment named `default`, or the sole Docker or Daytona
+            environment when `default` does not exist. Omission is rejected if
+            that fallback is absent or ambiguous.
         parent_id:
           type: string
           description: Optional orchestration parent run ID.
--- a/docs/public/api-reference/fabro-api.yaml
+++ b/docs/public/api-reference/fabro-api.yaml
@@ -1181,8 +1181,8 @@ paths:
         Failures return the standard error body. The intent lane responds
         `404` (`workflow_version_not_found`, `environment_not_found`), `422`
         (`run_intent_invalid`, `target_invalid`,
-        `target_environment_unsupported`, `workflow_version_unusable`,
-        `run_compile_invalid`), `503` (`integration_unavailable`), or `500`
+        `target_environment_unsupported`, `environment_selection_required`,
+        `workflow_version_unusable`, `run_compile_invalid`), `503` (`integration_unavailable`), or `500`
         (`workflow_version_store_error`, `credential_store_error`,
         `variable_store_error`, `run_persistence_failed`).
       requestBody:
--- a/docs/public/execution/environments.mdx
+++ b/docs/public/execution/environments.mdx
@@ -178,10 +178,10 @@ fabro server start --environment default
 ## Seeded Environments
 
 Install seeds a `default` environment into SQLite. It is a normal persisted
-environment, so you can edit or delete it. When it is absent, run creation that
-does not name an environment uses the sole Docker or Daytona environment in the
-catalog; zero or multiple compatible environments require an explicit
-selection. The standard Docker default is:
+environment, so you can edit or delete it. Runs that do not name an environment
+select it implicitly; see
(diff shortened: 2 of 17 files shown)


swerner:
## Simplify pass

Four review angles (reuse, simplification, efficiency, altitude) were run over this PR. Fixes landed in 72da10816; skipped findings are listed below for the record.

### Fixed

- **CLI now selects the environment the same way the server does.** When `--environment` is omitted, `resolve_run_environment` lists the catalog once and calls `select_implicit_run_environment`, instead of fetching `default`, sniffing the 404 via `is_not_found_error`, and then listing. The selection rule now lives in exactly one function, the implicit path is always one round trip, and the CLI no longer imports `DEFAULT_ENVIRONMENT_ID`. All four angles flagged this.
- **Test mocks updated to match.** Added `mock_environment_catalog` / `environment_catalog_json` to the CLI test support module and switched the tests that mocked `GET /environments/default` to the list endpoint. The new fallback test dropped its hand-written 404 stub.
- **Simpler selection helper.** The two `let ... else` steps and the `2 + count()` arithmetic became a collect plus a slice match on `[environment]`.
- **`?` instead of manual error mapping** in `select_intent_environment_id` and the CLI resolver; the `#[from]` conversion and anyhow already cover it.
- **`ImplicitRunEnvironmentError` moved to `error.rs`**, alongside the crate's other error types.
- **Docs.** Added `environment_selection_required` to the endpoint's enumerated 422 codes in the OpenAPI spec and the generated TS client comment. Fixed the stale "Defaults to `default`" row in run-configuration.mdx. environments.mdx now links to the run-configuration section instead of restating the rule a third time.

### Skipped

- **Server `list()` clones and sorts the whole catalog to pick one ID.** Catalogs are tiny and run creation already does heavier I/O; a store-level accessor is not worth the extra API surface.
- **Manual `refresh_manifest_run_settings_from_environment_catalog()` in the server test helper.** Mirrors what production handlers do by hand. Pre-existing leaky abstraction; belongs in its own change.
- **fabro-environment tests spin up a SQLite store to test a pure function.** Matches every other test in that file.
- **PR description overstates the match with the automation migration.** The migration only honors `default` when it is clone-based; the new helper honors any `default` (intentional per its doc comment). Worth softening the "matches the conservative selection rule" line.

swerner:
## Closing without merge

After looking at the history behind the `default` convention, we're taking a different direction. Notes for the record; the simplify pass above is moot.

### Background

- Before #831, the CLI built the manifest locally and took `run.environment.id` from the merged config layers, falling back to `default`. The catalog it resolved against always included the built-in seeded `default` layer, so `default` could never be missing.
- #780 (server intent handler) and #831 (CLI intent producer) replaced that with a server lookup of the literal `default`. The server catalog is a real SQLite table where `default` exists only if `fabro install` seeded it and nobody deleted it. That created a new failure mode this PR was trying to paper over.

### Why not this fallback

- A heuristic that picks "the sole Docker or Daytona environment" is an implicit choice a user cannot see, and it needs the same rule in two processes (CLI and server).
- Upcoming work will introduce a set of named official environments and require exactly one to carry a `default` boolean. Once that lands, "which environment is default" is an explicit user decision and this fallback would be removed anyway.
- `project.toml` is being migrated away, so there is no config-layer source of the id left to honor. The choice is binary: explicit `--environment`, or whatever the server marks as default.

### What we'll do instead

- Keep the pre-#836 behavior: an omitted id means `default`, and a miss is an error.
- Improve the CLI error on that miss: list the catalog and say something like "environment `default` not found; pass `--environment <id>` or create one named `default`. Available: `production`, `staging`." One call in the not-found branch, no shared selection rule.
- Leave the server's `environment_not_found` code as is; `environment_selection_required` is dropped.
- Drop the fallback docs. The stale "Defaults to `default`" table row in run-configuration.mdx stays correct.
- Salvage `Client::list_environments` from this branch for the error message.

PR #837, opened 2026-09-02
Reject automatic pull requests for Local runs

swerner:
## Summary

- reject RunIntent creation with a stable 422 error when resolved workflow settings enable automatic pull requests for a configured Local environment
- keep disabled pull requests valid for Local runs and keep configured Docker dry runs valid despite effective Local execution
- document that automatic pull requests require a clone-based Docker or Daytona environment

## Testing

- `cargo nextest run -p fabro-server run_intent`
- `cargo nextest run -p fabro-server folder_target`
- `cargo nextest run -p fabro-cli create_uses_workflow_owned_pull_request_settings_only`

### DIFF
--- a/docs/public/api-reference/fabro-api.yaml
+++ b/docs/public/api-reference/fabro-api.yaml
@@ -1181,7 +1181,8 @@ paths:
         Failures return the standard error body. The intent lane responds
         `404` (`workflow_version_not_found`, `environment_not_found`), `422`
         (`run_intent_invalid`, `target_invalid`,
-        `target_environment_unsupported`, `workflow_version_unusable`,
+        `target_environment_unsupported`,
+        `pull_request_environment_unsupported`, `workflow_version_unusable`,
         `run_compile_invalid`), `503` (`integration_unavailable`), or `500`
         (`workflow_version_store_error`, `credential_store_error`,
         `variable_store_error`, `run_persistence_failed`).
--- a/docs/public/execution/run-configuration.mdx
+++ b/docs/public/execution/run-configuration.mdx
@@ -555,9 +555,7 @@ The `sandbox` transport runs the MCP server inside the workflow's sandbox. This 
 
 ### `[run.pull_request]`
 
-Automatically open a GitHub pull request when the workflow run completes successfully. Requires a [GitHub App](/integrations/github) to be configured.
-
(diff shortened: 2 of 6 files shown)


swerner:
## Simplify pass (Fable 5.1)

Pushed c5010c0 with cleanups from a four-angle review (reuse, simplification, efficiency, altitude). Formatting, clippy, and the affected fabro-server / fabro-api tests pass.

### Fixed

- **Admission check uses the shared idioms.** `validate_intent_environment` now reads `!configured_provider.is_clone_based() && settings.run.pull_request.is_some()`. `is_clone_based()` is what every other pull-request gate uses, and the settings resolver already drops `run.pull_request` unless it is enabled, so re-checking `.enabled` was dead. A comment records that invariant.
- **OpenAPI spec drift.** The create-run description enumerates the intent lane's 422 codes and was missing `pull_request_environment_unsupported`. Added.
- **Trimmed the disabled-pull-request acceptance test** to the 201 outcome; the projection read was asserting resolver behavior and test-state defaults rather than this change.
- **Simplified the dry-run test's final assertion** to `pull_request.is_some()`.
- **Docs** folded into the existing requirements sentence instead of stating the rule twice.

### Considered and skipped

- **Moving the check into `fabro_config::resolve_run`** next to the existing `run.pull_request` / `run_branch` cross-validation. It would also cover the legacy manifest lane, which currently still admits Local + `pull_request.enabled`. Skipped because: the legacy lane has a test deliberately allowing that combination (`preflight_allows_pull_request_enabled_without_github_credentials`), `defaults.toml` seeds the default environment as `local` so config-level tests and `fabro repo init` output would start failing, and resolver failures reach API clients as a bare "failed to resolve manifest settings" with no reason. That is a product decision rather than a cleanup, but worth deciding explicitly.
- **Collapsing `AutomaticPullRequestUnsupported` into `TargetUnsupported { detail }`.** The distinct error code is a deliberate wire-contract choice, so it stays and is now documented.
- **Using `--environment default` in both halves of the CLI test.** The `local` half is load-bearing: it proves project-only pull-request settings do not cross admission, and the new check would reject them if they did.
- **Extracting "nothing persisted" / "read projection" test helpers.** Pre-existing duplication across dozens of tests, not introduced here.

PR #838, opened 2026-09-02
Persist pre-start worker failures

swerner:
## Summary

- Allow durable Runnable runs to transition directly to LaunchFailed, WorkflowError, Terminated, or BootstrapFailed when setup or a worker fails before the worker emits Starting.
- Preserve pending cancellation as the authoritative outcome when a worker launch fails.
- Keep the existing scheduler, process-local mutex guard, worker-owned Starting timing, and every other failure-transition rule unchanged.

Runnable startup reconciliation remains separate follow-up work and is not included here.

## Verification

- cargo nextest run -p fabro-types runnable_pre_start_failures
- cargo nextest run -p fabro-server subprocess_pre_start_failure
- cargo nextest run -p fabro-store rejected_transition
- cargo build --workspace
- cargo nextest run --workspace --status-level slow --profile ci (8,412 passed; 208 skipped)

### DIFF
--- a/lib/apps/fabro-server/src/server.rs
+++ b/lib/apps/fabro-server/src/server.rs
@@ -3585,18 +3585,45 @@ async fn fail_worker_launch(
     err: anyhow::Error,
 ) {
     tracing::error!(run_id = %run_id, error = %err, "Failed to spawn worker");
-    let message = format!("Failed to spawn worker: {err}");
+    let cancellation_pending = match run_store.state().await {
+        Ok(run_state) => run_state.pending_control == Some(RunControlAction::Cancel),
+        Err(state_err) => {
+            tracing::warn!(
+                run_id = %run_id,
+                error = %render_compact_with_causes(
+                    &state_err.to_string(),
+                    &collect_causes(&state_err),
+                ),
+                "Failed to load run state while recording worker launch failure"
+            );
+            false
+        }
+    };
+    let (error, reason, message) = if cancellation_pending {
+        (
+            WorkflowError::Cancelled,
+            FailureReason::Cancelled,
+            "Run cancelled before worker launch completed".to_string(),
+        )
+    } else {
+        let message = format!("Failed to spawn worker: {err}");
+        (
+            WorkflowError::engine_with_anyhow("Failed to spawn worker", err),
+            FailureReason::LaunchFailed,
+            message,
+        )
+    };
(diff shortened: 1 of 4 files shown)


swerner:
## `/simplify` pass

Applied in c437dc012 (clippy, fmt, and the fabro-types / fabro-store / fabro-server test suites are clean).

**Fixed**

- **Pre-start classification lives on the type.** `FailureReason::can_occur_before_start()` is an exhaustive match, and `can_transition_to` uses it for every `Runnable -> Failed` case. This replaces the four-reason allowlist and its "keep aligned with callers" comment, so adding a `FailureReason` variant forces the decision at compile time. The original `(Runnable, Starting)` arm is restored unchanged. The test now iterates all variants via strum's `VariantArray`, and the comment explains that the guard protects live appends only (replay synthesizes missing statuses).
- **Cancel precedence has one owner.** `failure_for_incomplete_run` is a thin wrapper over a new `failure_honoring_pending_cancel` that takes the fallback. `fail_worker_launch` calls the same helper instead of re-deriving the pending-cancel check inline, and its state-load warning logs the error the same way the worker-exit path does.
- **Test runtime simplified.** `PreStartWorkerRuntime` drops the `hold_launch_failure` flag for `release_start: Option<Notify>` with a `held(outcome)` constructor that works for either outcome.
- **Shared test helpers.** `wait_until` replaces two copies of the Notify/timeout poll loop; `write_test_server_record` replaces two copies of the fake `ServerDaemon` fixture; `assert_run_failed_before_start` and `run_failed_reasons` replace ~60 duplicated assertion lines across the two new tests, and the redundant "no LaunchFailed event" check is gone.
- **Slate test payloads deduplicated.** `failure_payload(label, reason, message)` backs both `workflow_failure_payload` and `sandbox_init_failure_payload`; the latter asserts its reason is post-start so the rejection test can't silently lose meaning again.

**Skipped**

- `fail_worker_launch` deep-clones the whole run projection to read `pending_control`. Avoiding that needs a new fabro-store accessor, and this is a cold failure path.
- Other pre-start failure paths (sandbox-provider rejection, secret-load failure, "worker wait failed") still ignore a pending cancel. Pre-existing and outside this PR's scope.

**Worth noting in the description:** this PR flips an older deliberate assertion so `Runnable -> Failed { Terminated }` is now allowed. Justified by the early-exit path, but it is a semantic change.

PR #845, opened 2026-09-08
Select CLI workflow sources and run targets independently

swerner:
`fabro run` and `fabro create` select workflow instructions independently from the repository or directory the run works on. Both commands accept remote workflow shorthand `OWNER/REPO[@REF]:WORKFLOW` and target shorthand `--target OWNER/REPO[@BRANCH]`, alongside `--workflow-repo`, `--workflow-ref`, `--target-from`, `--target-repo`, and `--target-branch`.

```sh
fabro run acme/workflows@v1.2:review --target acme/app@release/v2
fabro create ./review.toml --target-from ../app

# Equivalent explicit repository selection:
fabro run review --workflow-repo acme/workflows --workflow-ref v1.2 \
  --target-repo acme/app --target-branch release/v2
```

The workflow selector remains required; repository default workflows are deferred. Bare names and paths retain local lookup, and explicit local paths escape shorthand parsing. Workflow refs accept branches, tags, HEAD, or full commit SHAs; target suffixes accept working branches. Shorthand and explicit flags for the same resource cannot be combined. Repository slugs currently imply GitHub.com.

Explicit targets take precedence. For local workflows without target flags, the configured `run.scm` repository still participates in target inference; without that configuration, the target is derived from the caller directory. Clone-based inference retains repository-mismatch and exact-commit checks. Local targets use live directories, while clone targets exclude uncommitted changes and may publish committed local changes. Target selection does not change workflow or goal-file lookup.

Remote workflow acquisition uses local native Git credentials and uploads the canonical workflow-version closure from a temporary checkout. Sandbox clone credentials remain separate. Missing or moved refs never fall back to another revision. Existing containment, bounded diagnostics, and process/temporary-file cleanup apply to both syntax forms. Create leaves the run submitted; run starts it and attaches unless detached. Dry-run can still fetch and upload source.

Validation of the syntax update: all 1,033 CLI tests passed (49 skipped), pinned-nightly formatting and CLI Clippy with warnings denied, and generated documentation checks. Real Git fixtures exercise source/target combinations, shorthand equivalence, invalid inputs, explicit target precedence over `run.scm`, and run/create lifecycle behavior. A normal-binary smoke using an isolated local server and temporary directories verifies create-only behavior, exact-SHA shorthand execution, local-file execution, target isolation, missing-ref rejection, and cleanup. Live Docker/Daytona execution was not exercised.


swerner:
## `/simplify` pass (commit 449c23629)

Four cleanup reviewers (reuse, simplification, efficiency, altitude) ran over the diff. Findings were deduplicated and applied in one commit; the rest are listed as skipped with the reason.

### Applied

- **Dropped hand-written flag rules** in `selection::parse`. Clap already enforces `requires` / `conflicts_with` on the same flags, so the three duplicate checks were unreachable.
- **Removed downstream re-validation** in `remote_workflow.rs`. Selector, branch name, ref, and commit SHA were validated at parse time and then again in `collect`, `resolve_target`, and `resolve_revision`, with divergent error messages.
- **Delegated selector shape rules to `fabro_types::WorkflowPath`** instead of a parallel component walk. The CLI keeps only its own rule: bare name or an explicit `.toml` / `.fabro` file.
- **Replaced the hand-rolled `killpg` with `fabro_proc::sigkill_process_group`**, and the test liveness probe with `fabro_proc::process_exists`. Drops the `signal` and `process` features from the CLI's `nix` dependency.
- **Collapsed duplicated ref derivation** into one `RefCandidates` type that owns both the `ls-remote` patterns and branch-vs-tag resolution.
- **Made the failure label an explicit argument** to `NativeGit::command`. The old `operation()` lookup labeled only two of five call sites and had an unreachable `checkout` arm.
- **Hoisted the duplicated resolver call** in `create_run` into one closure.
- **Merged the two directory arms** in `resolution::target` and dropped `TargetSelection::CurrentDirectory`, which was just `Path(".")`.
- **Hoisted the `.git` path allocation** out of the symlink walk's per-entry filter.
- **Consolidated test scaffolding** into a `test_support` module shared by the args, selection, resolution, and remote workflow unit tests. Deleted the integration test that was one cell of the cross-product test, and made two vacuous test vectors assert the clap rejection they actually exercise.

### Kept after checking

- `Box<RunArgs>` is justified: unboxing it fires clippy's `large_enum_variant` on the outer `Commands` enum. Added a one-line comment saying why.

### Skipped (out of scope for this PR)

- Extracting a shared Git subprocess runner so the CLI and `fabro-server`'s `git_checkout.rs` stop maintaining parallel fetch/checkout pipelines. Cross-crate; the hardening added here would benefit the server path too.
- Moving selector validation (a `RepositoryWorkflowSelector` newtype) and the symlink containment check into `fabro-manifest` so local packages get the same protection. Needs manifest changes.
- Joining the target and workflow `ls-remote` lookups, and a single-round-trip fetch for `HEAD` / fully qualified refs. Both alter the PR's stated ordering guarantees (fetch waits until target validation passes).
- Flattening the five new flags into `#[derive(Args)]` groups. Nice-to-have restructure of `RunArgs`; the duplicate enforcement it would have removed is already gone.

### Verification

Formatting, clippy with warnings denied, and `cargo build --workspace` pass. All 1015 fabro-cli tests pass.

### Pre-existing flake (not from this pass)

`remote_workflow_timeout_and_cancel_reap_owned_children` fails reliably when run with only its sibling selection / remote workflow tests (`-E 'test(/selection|resolution|remote_workflow/)'`), but passes alone and in the full suite. The original branch fails identically. The timeout branch reads the fake git's `pid` file without waiting for it, and under that scheduling the file is missing. Left as-is since it's a correctness issue rather than a cleanup.

swerner:
Pushed one commit per review finding on top of the branch. Each commit stands alone and the suite is green after each (fmt, workspace build, workspace clippy with `-D warnings`, and the `fabro-cli`, `fabro-server`, `fabro-manifest`, and `fabro-proc` test suites).

**Correctness**

- **Ctrl-C was swallowed after native Git acquisition.** `owned()` polled tokio's `ctrl_c()` only for one acquisition; on Unix that listener permanently replaces the default SIGINT disposition, so `fabro create`, `fabro run --detach`, and `fabro run` before `attach` ignored Ctrl-C while waiting on the server. New `Interruption` handle: the command entry points create it from the run args, it installs a listener only when `--workflow-git`/`--target-git` is present, guards create (and start for `fabro run`) as one phase, and tracks owned Git tasks so interruption waits for their cleanup. Covered by an in-process SIGINT test.
- **`ls-remote` ran in the caller's cwd while fetch ran in the temp repo**, so repo-local `insteadOf`/credential/http config applied to one but not the other, and a broken `.git` in cwd failed the lookup. Every command now runs inside an owned scratch repository that is initialized first; `NativeGit` no longer takes a cwd at all.
- **Target observation could `git push` before the workflow ref was validated.** For `--workflow-git`, target resolution (which on Docker/Daytona may push the attached branch of a path target) ran before the remote ref existed. Remote acquisition now runs after parent/environment validation and before target observation. Regression test uses a caller checkout with an unpushed branch and asserts the bare origin never receives it.
- **Symlink pre-walk failed on unrelated links.** Dropped `check_source_symlinks`; `collect_workflow_versions` now canonicalizes and root-checks only the selected TOML (or a graph selector's sibling TOML), which were the only unchecked reads. Dangling submodule-style links and links into the host elsewhere in the repo no longer fail acquisition. `walkdir` moved to dev-dependencies.
- **Hardening list omitted `core.fsmonitor`.** Added `-c core.fsmonitor=false` to match the sandbox's `GIT_HARDENED`; the hooks/filters test now also plants an fsmonitor hook and fails without the fix.
- **Remote default branch grammar mismatch.** A default HEAD like `heads/main` passed the selector check and then failed in `GitRunTarget::validate` with a generic error. `resolve_target` validates it as a working branch name up front and tells the user to pass `--target-branch`.

**Efficiency / simplification / reuse**

- **Unconditional process-group SIGKILL** after every Git command destroyed `credential-cache--daemon` right after warming it. The group is killed only on timeout, cancellation, or failure; the redundant kill/wait on a reaped child is gone.
- **`RefCandidates` re-derived the branch/tag split** that `RemoteWorkflowRevision::parse` already enforced. `parse` now yields `Branch`/`Tag`/`Name` variants and resolution matches on them; the `Option/Option` encoding and its impossible `(None, None)` state are gone.
- **Shared supervised runner.** New `fabro_proc::SupervisedCommand` owns stdio, the process group, timeout, cooperative cancellation, and bounded capture. Both the server's `run_git_plan` and the CLI's `NativeGit::command` sit on it. Side effect for the server: a timed-out clone/fetch now takes its helper processes down with it. The CLI's `-c` overrides are one named `HARDENED_GIT_CONFIG` list.

**Tests**

- The run-driven iterations of `remote_workflow_run_starts_once_create_leaves_submitted_and_failures_do_not_refetch` moved to `cmd/run.rs`; `cmd/create.rs` keeps the single create-leaves-submitted case. The workflow and remote-repo fixture helpers moved to the shared `cmd/support.rs`.

**Not changed**

- The sync `ls-remote` parser in `fabro_workflow::git::remote_branch_sha_noninteractive` and the shell-string `GIT_HARDENED` in `sandbox_git.rs` are left as they are; they live in a synchronous, shell-context world and unifying them would widen this PR well past the CLI.
- The local-package resolver still reads the selected TOML before canonicalizing; the containment check landed only in the remote collector, where the checkout is untrusted.

PR #851, opened 2026-09-11
Add fabro_workflow_version_create to agent tools and MCP

swerner:
Adds `fabro_workflow_version_create` to native agent tools and standalone MCP. Callers supply an exact `entrypoint` and a map of text file contents, then receive `{workflow_version_id}` without creating or starting a run.

The supplied tree is staged in an owned temporary directory and packaged with the same collector, bundler, semantic validation, and content hashing used by CLI and automation workflows. The backend registers dependencies before parents through the existing client/API; content-addressed versions make partial-upload retries safe. Shell content is preserved and never executed during packaging.

Packaging results share `CollectedWorkflowClosure` from `fabro-workflow-version`, and input file limits reuse the existing workflow-version validation. The supplied-content adapter remains the dependency-injection boundary between tools and the manifest collector. Test fixtures are shared in a test-only module.

Validation enforces exact paths, containment, portable path collisions, file and byte limits, and a 64-level traversal/assembly limit. Supplied sibling `workflow.toml` files must be valid; valid configs selecting a different graph remain omitted. Errors identify paths and limits without exposing source contents, and staging is cleaned up on success and failure.

The HTTP route reuses `RequiredRunManagementActor` to accept users and workers with `agent:run_tools`. Ordinary workers and same-run backends remain denied. Native dispatch is integrated with the Pebble agent implementation on current main.

## Validation

- 937 selected tests passed across packaging, workflow types/config, native dispatch, MCP, CLI, server authorization, and automation materialization.
- Pinned nightly formatting and workspace Clippy (`--all-targets -- -D warnings`) passed.
- `cargo build --locked --workspace` passed; the shared fixture content is absent from the normal binary and manifest library artifact.
- A freshly built server and MCP bridge passed 12 checks in isolated temporary storage: persisted parent/child contents and SHA-256 IDs, idempotent retries, graph/config parity, changed-content IDs, authorization, safe malformed/deep-input failures, staging cleanup, no runs created, and no supplied source in logs.
- `git diff --check` passed.

Release-mode and live-provider tests were not rerun.


swerner:
## Simplify pass (Claude Fable 5.1)

Four cleanup reviewers (reuse, simplification, efficiency, altitude) went over the diff; the fixes below landed in 5f838b9a8. `cargo build --workspace`, fmt, clippy `--all-targets -D warnings`, and 3,267 tests across the touched crates plus the CLI MCP tests all pass.

### Fixed

- **Packaging moved to the right layer.** Staging and collection now live in `fabro-manifest` as `collect_supplied_workflow_versions`, beside the checkout collector. The `fabro-server` module shrank to a thin `ServerWorkflowVersionPackager` that runs it on a blocking thread and curates the error.
- **Seam narrowed.** The injected trait is now `WorkflowVersionPackager`, returning the dependency-ordered versions. `ClientBackend` performs the upload with the client it already owns, instead of the trait taking a client and uploading itself.
- **Validation runs once.** Added `ValidatedWorkflowVersionCreate` with `TryFrom`, matching the other tools. The MCP handler validates before auth; the backend and packager receive the validated type. The three repeated `validate()` calls are gone.
- **Reused existing helpers.** The hand-rolled duplicate-key visitor was replaced by the `fabro-types` unique-map deserializer (now exposed as `deserialize_unique_map`). The "not available" error goes through a shared helper keyed on the tool-name constant. Budget error messages derive from the `MAX_WORKFLOW_VERSION_*` constants instead of hardcoded literals.
- **Tool output matches siblings.** The text summary is now a one-line message and the agent dispatch goes through `render_fabro_tool_result` like the other eight tools.
- **Path collision check.** The bool flag became a key closure returning `Cow`, so the canonical exact-match path no longer allocates per key, and ASCII paths take a lowercase fast path before Unicode folding.
- **Smaller items.** `WorkflowLocation::resolve` and `from_exact_path` share one dispatch helper. The bundler's two branches compute the normalized path once. The packaging error chain is logged with `warn!` before the generic message crosses the tool boundary, per the error-handling strategy. The hand-rolled axum test server was replaced with httpmock. Long unformatted lines were hoisted so rustfmt can format them. `unicase` and `unicode-normalization` moved to workspace dependencies.

### Skipped

- **Collector-level portability check.** The collector still calls `validate_workflow_source_paths`, which also applies to on-disk automation checkouts. Removing or relocating it is a policy decision about whether all versions must be portable, so it was left as-is for a human call.
- **Per-method run-scope denial in `ClientBackend`.** It duplicates what the Ask Fabro composition site already enforces, but it is three explicit lines with a test. Kept.
- **Sequential uploads and per-attempt clones in `register_workflow_versions`.** Pre-existing `fabro-client` code outside this PR.
- **Duplicated MCP tool descriptions and `RequiredRunToolActor` vs `RequiredRunManagementActor`.** Pre-existing patterns across all tools and routes; belong in a separate change.
- **Docs prose overlap** between the MCP page and the API spec description. Minor, and arguably useful for readers of each page.

